PR #281, opened 2017-07-21
ACCUMULO-4655 Added a Response Time column to the monitor

ivakegg:
Added a Response Time column to the monitor which shows how long it took to get status from that tserver.
This is useful in determining whether long Last Contact times are dues to one or more tservers or whether
this is due to churn in the master.


ivakegg:
This is the merge of the ACCUMULO-4655 from 1.8.2 into master.  I updated the tablets.js and the TabletServerInformation to include this in the new monitor tservers page.

lstav:
Did you branched out of the work I did on #280? I noticed that your code has the version variable introduced on that PR.

ivakegg:
I branched off of master.  As you noted it appears that #280 was commited into master but was not closed._

lstav:
Thanks, I'll close #280 to avoid future confusions.

PR #284, opened 2017-07-26
ACCUMULO-4686 Fix upgrade process to set version in all volumes.

ivakegg:
The upgrade process was only setting the version in one of a multi-volume system.
This fixes the code to set the version on all volumes.


ivakegg:
I did not realize we were still supporting changes in 1.7.....I will check.

ivakegg:
yes, this applies to 1.7 as well.

joshelser:
Hurm, in hindsight, did you happen to take a look at the feasibility of adding a unit test for the change to `updateAccumuloVersion(VolumeManager, int)`? Doesn't look like it would be too bad at a glance. Trying to check this in an IT would probably be more pain than its worth.

ivakegg:
OK, working on a unit test now....

ivakegg:
Commits pushed to gitbox

PR #286, opened 2017-07-31
ACCUMULO-4689 Do not set empty SSL store types

joshelser:
Setting an empty type will override Jetty's implicit default value
of JKS (which is most likely what Accumulo users would use).

### DIFF
--- a/server/monitor/src/main/java/org/apache/accumulo/monitor/EmbeddedWebServer.java
+++ b/server/monitor/src/main/java/org/apache/accumulo/monitor/EmbeddedWebServer.java
@@ -52,27 +52,7 @@ public class EmbeddedWebServer {
       usingSsl = false;
     } else {
       SslContextFactory sslContextFactory = new SslContextFactory();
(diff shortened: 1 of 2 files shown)


milleruntime:
Just curious but why open a PR for only a few hours?  Even if the change is minor, seems like hardly enough time for review.

joshelser:
Read the JIRA issue, please. And, for the record, there is no obligation of review before commit.

milleruntime:
> there is no obligation of review before commit.

Right, that's why I was wondering why you bothered with the PR.  

I see your comments on JIRA, thanks.

joshelser:
> >    there is no obligation of review before commit.

> Right, that's why I was wondering why you bothered with the PR.

Personal preference. If someone wants to make time to review my code, I like to go out of my way to make it easier for them to do so. The level of confidence I have in a change determines on how long I'll wait for that. If it's truly trivial I won't (in this case specifically, there were two ways to implement it that were essentially the same thing. The irony is that the other implementation is actually what Billie had done previously)

PR #293, opened 2017-08-17
ACCUMULO-4669 Use windowed statistics in RFile

keith-turner:


core/src/main/java/org/apache/accumulo/core/file/rfile/RFile.java
```diff
@@ -416,8 +416,9 @@ public void flushIfNeeded() throws IOException {
     }
 
     private boolean isGiantKey(Key k) {
-      // consider a key thats more than 3 standard deviations from previously seen key sizes as giant
-      return k.getSize() > keyLenStats.getMean() + keyLenStats.getStandardDeviation() * 3;
+      double mean = keyLenStats.getMean();
+      double stddev = keyLenStats.getStandardDeviation();
+      return k.getSize() > mean + 3 * Math.max(mean, stddev);
```
mikewalch: Could add some comments here about why this logic is used to determine if key is giant...

core/src/main/java/org/apache/accumulo/core/file/rfile/RFile.java
```diff
@@ -403,7 +402,8 @@ public void flushIfNeeded() throws IOException {
 
     private SampleLocalityGroupWriter sample;
 
-    private SummaryStatistics keyLenStats = new SummaryStatistics();
+    // Use windowed stats to fix ACCUMULO-4669
+    private RollingStats keyLenStats = new RollingStats(2017);
```
mikewalch: Why 2017 for window size?

core/src/main/java/org/apache/accumulo/core/file/rfile/RollingStats.java
```diff
@@ -0,0 +1,114 @@
...
+  RollingStats(int windowSize) {
+    this.windowFull = false;
+    this.position = 0;
+    this.window = new double[windowSize];
+  }
+
+  /**
+   * @see <a href= "http://jonisalonen.com/2014/efficient-and-accurate-rolling-standard-deviation/">Efficient and accurate rolling standard deviation</a>
+   */
+  private void update(double n, double o, int w) {
```
mikewalch: I guess `n` & `o` is for new & old.  Could instead use `newValue` & `oldValue` to make things clear.  What is `w`?

asfgit:

keith-turner:
merged in 26e83f05d1448631f0b8a0da1b8671abe9beb922

PR #296, opened 2017-09-05
ACCUMULO-4700 Set log4j appender threshold to ALL

mikewalch:


assemble/conf/log4j-monitor.properties
```diff
@@ -21,7 +21,7 @@ log4j.appender.file=org.apache.log4j.RollingFileAppender
 log4j.appender.file.File=${accumulo.log.dir}/${accumulo.application}.log
 log4j.appender.file.MaxFileSize=100MB
 log4j.appender.file.MaxBackupIndex=10
-log4j.appender.file.Threshold=INFO
+log4j.appender.file.Threshold=ALL
```
milleruntime: Any difference by setting it to ALL vs just removing it completely?

mikewalch: I cannot find docs confirming this but I think the default is ALL so it could be removed.  I left it to make it easy for users to change but I am OK with it being removed.

milleruntime: Ok I was just curious.  Doesn't matter to me

PR #297, opened 2017-09-11
ACCUMULO-4702 Code fixes to work with Guava 23

milleruntime:
* Remove the dependency on Beta-annotated Guava classes and methods
* Pulled in HostAndPort from Guava 23
* Replaced usage of HostAndPort getHostText with getHost
* Replaced Iterators.emptyIterator with one from Collections
* Kept the same Guava 14 dependency

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
+++ b/core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
@@ -22,6 +22,13 @@ import java.io.Serializable;
 import com.google.common.base.Strings;
 
 /**
+ * This class was copied from Guava release 23.0 to replace the older Guava 14 version that had been used in Accumulo.
+ * It was annotated as Beta by Google, therefore unstable to use in a core Accumulo library.  We learned this the hard
+ * way when Guava version 20 deprecated the getHostText method and then removed the method all together in version 22.
+ *
+ * Unused methods and annotations were removed to reduce maintenance costs.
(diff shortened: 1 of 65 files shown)


core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
```diff
@@ -0,0 +1,262 @@
...
+package org.apache.accumulo.core.util;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+
+import java.io.Serializable;
+
+import com.google.common.base.Strings;
+
+/**
```
keith-turner: I think it would be nice to add a comment somewhere in the javadoc stating which version of guava this was copied from and why...

core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
```diff
@@ -22,6 +22,13 @@
 import com.google.common.base.Strings;
 
 /**
+ * This class was copied from Guava release 23.0 to replace the older Guava 14 version that had been used in Accumulo.
+ * It was annotated as Beta by Google, therefore unstable to use in a core Accumulo library.  We learned this the hard
+ * way when Guava version 20 deprecated the getHostText method and then removed the method all together in version 22.
```
keith-turner: could reference the issue number

PR #299, opened 2017-09-14
ACCUMULO-4702 Removed Beta Guava dependencies

milleruntime:
Created code to remove Guava dependencies which are still beta in release 23:
com.google.common.util.concurrent.Uninterruptibles
com.google.common.io.Files
com.google.common.hash.HashCode;
com.google.common.hash.HashFunction;
com.google.common.hash.Hashing;

### DIFF
--- a/fate/src/main/java/org/apache/accumulo/fate/util/UtilWaitThread.java
+++ b/fate/src/main/java/org/apache/accumulo/fate/util/UtilWaitThread.java
@@ -16,6 +16,10 @@
  */
 package org.apache.accumulo.fate.util;
 
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+import java.util.concurrent.TimeUnit;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


fate/src/main/java/org/apache/accumulo/fate/util/UtilWaitThread.java
```diff
@@ -29,4 +33,36 @@ public static void sleep(long millis) {
...
+   * Copied from Guava release 23. The Uniterruptibles class was annotated as Beta by Google, therefore unstable to use. The following javadoc was copied from
+   * com.google.common.util.concurrent.Uninterruptibles:
+   *
+   * Utilities for treating interruptible operations as uninterruptible. In all cases, if a thread is interrupted during such a call, the call continues to
+   * block until the result is available or the timeout elapses, and only then re-interrupts the thread.
+   *
+   * @author Anthony Zana
+   * @since 10.0
+   */
+  public static void sleepUninterruptibly(long sleepFor, TimeUnit unit) {
```
keith-turner: It may be nice to put all copied guava code into one package.  Not sure if this is possible.

milleruntime:
This work for 1.7 was pretty minor but 1.8 has a lot more reliance on Guava and uses a LOT of beta code... and I am not sure its worth it to copy in/replace all Beta code in 1.8.

milleruntime:
Concerns for 1.8 (I will put them here since this may change the solution for all versions):
AbstractHashSampler.java:[89,31] md5() in com.google.common.hash.Hashing has been deprecated
AbstractHashSampler.java:[92,31] sha1() in com.google.common.hash.Hashing has been deprecated
com.google.common.hash.Hasher, which is annotated with @Beta At DataoutputHasher.java
Use of class com.google.common.util.concurrent.RateLimiter, which is annotated with @Beta At GuavaRateLimiter.java

milleruntime:
If there are no objections for this solution to 1.7 I will merge this today.  There will need to be follow on work for 1.8 and master.

PR #304, opened 2017-10-05
ACCUMULO-4716 Don't cache blks over max array size

jmark99:
ACCUMULO-4716 Don't cache blocks over max array size

Prevents byte array from caching up to Integer.MAX_VALUE to prevent
possible OutofMemory error as described in StackOverflow post
https://stackoverflow.com/a/8381338

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/file/blockfile/impl/CachableBlockFile.java
+++ b/core/src/main/java/org/apache/accumulo/core/file/blockfile/impl/CachableBlockFile.java
@@ -156,6 +156,10 @@ public class CachableBlockFile {
     private boolean closed = false;
     private AccumuloConfiguration accumuloConfiguration = null;
 
+    // ACCUMULO-4716 - Define MAX_ARRAY_SIZE smaller than Integer.MAX_VALUE to prevent possible OutOfMemory
+    // errors when allocating arrays - described in stackoverflow post: https://stackoverflow.com/a/8381338
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
+
     private interface BlockLoader {
       BlockReader get() throws IOException;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/accumulo/core/file/blockfile/impl/CachableBlockFile.java
```diff
@@ -342,7 +342,11 @@ private BlockRead getBlock(String _lookup, BlockCache cache, BlockLoader loader)
 
     private BlockRead cacheBlock(String _lookup, BlockCache cache, BlockReader _currBlock, String block) throws IOException {
 
-      if ((cache == null) || (_currBlock.getRawSize() > cache.getMaxSize())) {
+      // ACCUMULO-4716 - Define MAX_ARRAY_SIZE smaller than Integer.MAX_VALUE to prevent possible OutOfMemory
+      // error as described in stackoverflow post: https://stackoverflow.com/a/8381338.
+      int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8;
```
keith-turner: Personally, I would make this a static constants because all caps usually implies that by convention.

jmark99:
ACCUMULO-4716 Don't cache blocks over max array size

Modified to move MAX_ARRAY_SIZE from within the method to a constant class value instead.

core/src/main/java/org/apache/accumulo/core/file/blockfile/impl/CachableBlockFile.java
```diff
@@ -156,6 +156,10 @@ public long getStartPos() throws IOException {
     private boolean closed = false;
     private AccumuloConfiguration accumuloConfiguration = null;
 
+    // ACCUMULO-4716 - Define MAX_ARRAY_SIZE smaller than Integer.MAX_VALUE to prevent possible OutOfMemory
+    // errors when allocating arrays - described in stackoverflow post: https://stackoverflow.com/a/8381338
```
keith-turner: I like this comment.

keith-turner:
@jmark99  thanks for the contribution.  I squashed these commits and cherry picked them back to 1.7 in commit 6ead4740dbc24485b51d73d0a006d44bb2f52833.  I then merged the changes forward in commits ae3aa10027cc0926cd5b319d31bc3fd005ffb50f and 47d13beff82e547019faf551e9bdc76c4429e576.

jmark99:
Thanks, Keith.

PR #305, opened 2017-10-10
[ACCUMULO-4591] Add replication latency metrics

adamjshook:
Supersedes #220 

Trying to capture the latency based on WAL modification time and the current time once the file disappears from the status section of the replication table.

### DIFF
--- a/server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
@@ -38,6 +38,8 @@ import org.apache.hadoop.metrics2.lib.Interns;
 import org.apache.hadoop.metrics2.lib.MetricsRegistry;
 import org.apache.hadoop.metrics2.lib.MutableQuantiles;
 import org.apache.hadoop.metrics2.lib.MutableStat;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  *
@@ -47,6 +49,8 @@ public class Metrics2ReplicationMetrics implements Metrics, MetricsSource {
       RECORD = "MasterReplication";
   public static final String PENDING_FILES = "filesPendingReplication", NUM_PEERS = "numPeers", MAX_REPLICATION_THREADS = "maxReplicationThreads",
       REPLICATION_QUEUE_TIME_QUANTILES = "replicationQueue10m", REPLICATION_QUEUE_TIME = "replicationQueue";
+
+  private final static Logger log = LoggerFactory.getLogger(Metrics2ReplicationMetrics.class);
 
   private final Master master;
   private final MetricsSystem system;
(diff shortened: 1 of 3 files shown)


server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
```diff
@@ -124,4 +142,55 @@ protected int getNumConfiguredPeers() {
...
+
+    // Iterate through all the pending paths and update the mod time if we don't know it yet
+    for (Path path : paths) {
+      if (!pathModTimes.containsKey(path)) {
+        try {
+          pathModTimes.put(path, master.getFileSystem().getFileStatus(path).getModificationTime());
+        } catch (IOException e) {
+          // Ignore all IOExceptions
+          // Either the system is unavailable or the file was deleted
+          // since the initial scan and this check
```
joshelser: Add a `log.trace` in case you/someone ever runs into troubles and wants to try to debug this.

server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
```diff
@@ -124,4 +142,55 @@ protected int getNumConfiguredPeers() {
...
+    // Exit early if we have no replicated files to report on
+    if (deletedPaths.isEmpty()) {
+      return;
+    }
+
+    replicationQueueTimeStat.resetMinMax();
+
+    for (Path path : deletedPaths) {
+      // Remove this path and add the latency
+      long modTime = pathModTimes.remove(path);
```
joshelser: This will throw an error is `pathModTimes` ever does not have a mapping for `path` because you're unboxing the `Long` into a `long`.

server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
```diff
@@ -124,4 +142,55 @@ protected int getNumConfiguredPeers() {
   protected int getMaxReplicationThreads() {
     return replicationUtil.getMaxReplicationThreads(master.getMasterMonitorInfo());
   }
+
+  protected void addReplicationQueueTimeMetrics() {
```
joshelser: Could you add a simple test case for this, please?

server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
```diff
@@ -26,38 +29,51 @@
...
 
   Metrics2ReplicationMetrics(Master master, MetricsSystem system) {
     this.master = master;
     this.system = system;
 
+    pathModTimes = new HashMap<>();
+
     registry = new MetricsRegistry(Interns.info(NAME, DESCRIPTION));
     replicationUtil = new ReplicationUtil(master);
+    replicationQueueTimeQuantiles = registry.newQuantiles(REPLICATION_QUEUE_TIME_QUANTILES, "replication queue time quantiles in milliseconds", "ops",
```
joshelser: nit: capitalize "Replication" please

server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
```diff
@@ -26,38 +29,51 @@
...
+        "latency", 600);
+    replicationQueueTimeStat = registry.newStat(REPLICATION_QUEUE_TIME, "replication queue time stat in milliseconds", "ops", "latency", true);
```
joshelser: nit: "statistics" instead of "stat" (and capitalize Replication too)

server/master/src/main/java/org/apache/accumulo/master/metrics/Metrics2ReplicationMetrics.java
adamjshook: I was actually thinking how best to handle this.  It'd be an implementation error if the path to be removed wasn't in `pathModTimes`, since `deletedPaths` is a subset of `pathModTimes`.  Maybe check for existence and log an error if it doesn't exist?

joshelser: It's just the unboxing you really have to worry about. Something like the following would be fine:

```java
Long modTime = pathModTimes.remove(path);
if (modTime != null) {
  ...
}
```

Java will handle the unboxing for you once you guaranteed that `modTime` is definitely non-null.

adamjshook:
@joshelser Back at you!

adamjshook:
@joshelser Thanks, if it wouldn't be too much of a hassle.

joshelser:
No problem, dude. Thanks for the contribution!

PR #306, opened 2017-10-12

jmark99:

Reworded a few of the sentences and moved the path order up to the class
level rather than the method level.

classv17

methodv17

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/ClientConfiguration.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/ClientConfiguration.java
@@ -199,22 +199,22 @@ public class ClientConfiguration extends CompositeConfiguration {
   }
 
   /**
-  * Attempts to load a configuration file from the system using the default search paths.
-  * Uses the <em>ACCUMULO_CLIENT_CONF_PATH</em> environment variable, split on
-  * <em>File.pathSeparator</em>, for a list of target files.
-  * <p>
-  * If <em>ACCUMULO_CLIENT_CONF_PATH</em> is not set, uses the following in this order:
+   * Attempts to load a configuration file from the system using the default search paths. Uses the <em>ACCUMULO_CLIENT_CONF_PATH</em> environment variable,
+   * split on <em>File.pathSeparator</em>, for a list of target files.
+   * <p>
+   * If <em>ACCUMULO_CLIENT_CONF_PATH</em> is not set, uses the following in this order:
    * <ul>
-   *   <li> ~/.accumulo/config
-   *   <li> either:
-   *   <ul>
-   *     <li><em>$ACCUMULO_CONF_DIR</em>/client.conf, if <em>$ACCUMULO_CONF_DIR</em> is defined.
-   *     <li><em>$ACCUMULO_HOME</em>/conf/client.conf, otherwise.
-   *   </ul>
-   *   <li>/etc/accumulo/client.conf
-   *   <li>/etc/accumulo/conf/client.conf
+   * <li>~/.accumulo/config
+   * <li>either:
+   * <ul>
+   * <li><em>$ACCUMULO_CONF_DIR</em>/client.conf, if <em>$ACCUMULO_CONF_DIR</em> is defined.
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/accumulo/core/client/ClientConfiguration.java
```diff
@@ -41,7 +41,22 @@
 
 /**
  * Contains a list of property keys recognized by the Accumulo client and convenience methods for setting them.
- *
+ * <p>
+ * When loading a configuration file from the system, Accumulo uses the <em>ACCUMULO_CLIENT_CONF_PATH</em> environment variable,
```
keith-turner: Does the behavior described here only happen when calling `loadDefault()`?  Or do all of the constuctors do this search?

jmark99: Not all the constructors do the search. The described behavior occurs when load configurations use a default load path. Previously the path order was described only in the method. I moved it up to class level based upon the suggestion given by @madrob in the ticket comment.  It can be moved back into the method call only if you think that would be more appropriate.

keith-turner: >  It can be moved back into the method call only if you think that would be more appropriate.

Personally I would put the docs on the loadDefault method and then any constuctors that call load default I would state in the constuctors javadoc that it follows the search behavior described in loadDefatult using s javadoc link.  

If the docs are at the class level, I think the class level docs would need to detail which constuctors and methods do and/or do not follow the search behavior.

jmark99: Sounds good. I'll update to make the adjustments.  So if I understand you correctly I will only need 2 PR's.  One for the 1.7 change and another for the 1.8 change and then the 1.8 change will be merged forward to 2.0? Thanks.

keith-turner: >  One for the 1.7 change and another for the 1.8 change and then the 1.8 change will be merged forward to 2.0?

So you want the changes in 1.8 to be slightly different?  If so it would be nice to end up with the following on 1.8 branch.

 * Your commit with 1.8 changes
 * Merge commit with your changes from 1.7 
 * Your 1.7 commit

Lets go for the easiest way to get to those changes.  Lets get the changes in 1.7 squared away and then we can get the 1.8 changes figured out.

core/src/main/java/org/apache/accumulo/core/client/ClientConfiguration.java
```diff
@@ -199,15 +199,23 @@ public ClientConfiguration(Configuration... configs) {
...
+   * <li>/etc/accumulo/conf/client.conf
+   * </ul>
+   * <p>
    *
-   * A client configuration will then be read from each location using PropertiesConfiguration to construct a configuration. That means the latest item will be
-   * the one in the configuration.
-   *
-   * @see PropertiesConfiguration
-   * @see File#pathSeparator
+   * @since 1.6.0
```
keith-turner: This since tag is not needed because its the same as the class level since tag.  Methods inherit the since tag from the class.   Only need to set it on method if it differs from class since tag.

jmark99: Thanks for the catch! I realized I had inadvertently removed a line of documentation when I was performing the previous revision.

jmark99:
Thanks @keith-turner. I've rendered the javadocs for all versions on my local machine and they render correctly.

keith-turner:
I can merge this.  @jmark99  does #307 currently have the changes you want for 1.8?  If so I will look at that when merging forward.

jmark99:
Yes, @keith-turner , #307 has the correct documentation for 1.8. This documentation is also valid for the current master line of development. These was a slight modification to the loadDefaults() method between 1.7 and 1.8 that necessitated a small change to the documentation.

keith-turner:
@jmark99 I squashed+merged the changes.   There was one diff, I ran mvn compile and it reformatted some stuff.

PR #307, opened 2017-10-12

jmark99:

Reworded a few of the sentences and moved the path order up to the class
level rather than the method level.

classv18

methodv18


keith-turner:
Is this the same change as #306?  We usually make the change in the oldest branch and then merge it forward.

jmark99:
The change for 1.8 is slightly different than for 1.7. Although the 1.8 and 2.0 changes are identical so the #307 could be merged forward.  I wasn't sure how that situation was handled so I ended up creating  a pull request for master as well.

PR #308, opened 2017-10-12

jmark99:

Additionally moved the search path list to the class level vice the
method level.

classv2

methodv2


keith-turner:
This is closed by 63f5713708900e50b2177405be699cd4d07e97a0

PR #310, opened 2017-10-19
Accumulo 4721 Document rfile-info in the user manual

jmark99:
Accumulo 4721 Document rfile-info in the user manual

Updated the 1.7 user manual to provide examples using rfile-info rather than the org.apache.accumulo.core.file.rfile.PrintInfo class.


jmark99:
Thanks, Mike.

PR #321, opened 2017-11-10
ACCUMULO-4641 Added loading to cache API

keith-turner:


core/src/main/java/org/apache/accumulo/core/file/blockfile/cache/tinylfu/TinyLfuBlockCache.java
```diff
@@ -47,7 +47,7 @@
  * <li>Cache design: http://highscalability.com/blog/2016/1/25/design-of-a-modern-cache.html</li>
  * </ul>
  */
-public final class TinyLfuBlockCache implements BlockCache {
+public final class TinyLfuBlockCache extends SynchronousLoadingBlockCache implements BlockCache {
```
ben-manes: The underlying cache already does memoization using a per-entry lock. Is this needed?

keith-turner: I am not sure.  I saw a comment somewhere that said load operations should be *fast*, which is why I avoided relying on the cache.   I can't remember where I saw that.  I am certainly open to refactoring the code to use Caffeine for load synchronization.   I tried looking at the internals of the code to see where and how it did its load sync, but got lost.  Any tips would be appreciated.

ben-manes: Yes, the internals are unfriendly to glance through. Sadly performant code with lots of features quickly becomes gnarly.

A `get(k, loader)` or `Map.computeIfAbsent` delegate to an internal [computeIfAbsent](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/BoundedLocalCache.java#L1991) method. This does a non-blocking get, validates the entry if found, and does a blocking `computeIfAbsent` if absent or expired. This way hot entries are a cheap read.

ConcurrentHashMap does say computations should be fast, which we propagate into our docs too. I take it as a good reminder than a red flag warning, since it will be a blocking call when computed. More along the lines of keep your lock hold times short, or don't abuse db transactions for batch imports, rather than use locks or transactions sparingly.

keith-turner: So multiple loads can run concurrently as along as they are for different cache keys?  Is this what you meant by a per-entry lock?

I will refactor the code to use computeIfAbsent, but it may be a few days.

ben-manes: Sure, no rush. I mostly wanted to point out that the feature is built-in.

Like yours, there is a chance of collisions so both are effectively per-entry but not absolutely. In `ConcurrentHashMap` it locks on the bin which contains a small number of entries. The likelihood of collision on a write is stated to be...
> Lock contention probability for two threads accessing distinct elements is roughly 1 / (8 * #elements) under random hashes.

keith-turner: Thanks for sharing.  That is exactly the type of information I was looking for when spelunking in the Caffeine code.

ben-manes:
fyi, soft references are notorious for causing bad performance in heavily loaded systems. They increase the number of full GCs due to heap pollution and will be evicted when low on memory, which can cause thrashing. Not sure I understand well enough to advise except to always question whether a usage is justified.

keith-turner:
> fyi, soft references are notorious for causing bad performance in heavily loaded systems.

I may be able to redesign that.  How does Caffeine handle the weight of an entry changing?

ben-manes:
We require that the user update the entry using the cache so that it can
reevaluate the weight. That could be a put or computation. The weight is
stored with the entry and not re-evaluated on eviction.

asfgit:

PR #324, opened 2017-11-20
ACCUMULO-4744 Fixed RFile API scanner bug

keith-turner:


asfgit:

milleruntime:
@keith-turner I can take a look at this.  I want to checkout your branch though so I can see the bug.

core/src/main/java/org/apache/accumulo/core/file/blockfile/impl/CachableBlockFile.java
```diff
@@ -147,7 +147,7 @@ public long getStartPos() throws IOException {
   public static class Reader implements BlockFileReader {
     private final RateLimiter readLimiter;
     private BCFile.Reader _bc;
-    private String fileName = "not_available";
```
milleruntime: OK I see why you made the changes but this old fileName does make me curious as to why "not_available" was chosen... any idea?

keith-turner:
Merged in ed313f7

PR #327, opened 2017-11-28
ACCUMULO-4546 Create default log message for IllegalTableTransitionException

jmark99:
ACCUMULO-4546 Create default log message for IllegalTableTransitionException

Updated the IllegalTableTransitionException class within TableManager.java to print
a default log message when exception is thrown. User also has option of supplying a
custom message. Created unit test.

Moved import in AccumuloTest.java to make mvn happy.

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
@@ -115,10 +115,21 @@ public class TableManager {
 
     final TableState oldState;
     final TableState newState;
+    final String message;
 
     public IllegalTableTransitionException(TableState oldState, TableState newState) {
+      this(oldState, newState, "");
+    }
+
+    public IllegalTableTransitionException(TableState oldState, TableState newState, String message) {
       this.oldState = oldState;
       this.newState = newState;
+
+      String defaultMessage = "Error transitioning from " + oldState + " state to " + newState + " state";
+      if (message != null && !message.isEmpty())
+        this.message = message;
+      else
+        this.message = defaultMessage;
     }
 
     public TableState getOldState() {
@@ -127,6 +138,11 @@ public class TableManager {
 
     public TableState getNewState() {
       return newState;
+    }
+
+    @Override
+    public String getMessage() {
+      return message;
     }
 
   }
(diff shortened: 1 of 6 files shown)


asfgit:

server/base/src/test/java/org/apache/accumulo/server/tables/IllegalTableTransitionExceptionTest.java
```diff
@@ -0,0 +1,59 @@
...
+      throw new TableManager.IllegalTableTransitionException(oldState, newState);
+    } catch (IllegalTableTransitionException e) {
+      assertEquals(defaultMsg, e.getMessage());
+    }
+  }
+
+  @Test
+  public void testIllegalTableTransitionExceptionEmptyMessage() {
+    try {
+      throw new TableManager.IllegalTableTransitionException(oldState, newState, "");
```
keith-turner: In addition to `""`, could also try null.

server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
```diff
@@ -115,10 +115,21 @@ public TableState getTableState(String tableId) {
...
 
     public IllegalTableTransitionException(TableState oldState, TableState newState) {
+      this(oldState, newState, "");
+    }
+
+    public IllegalTableTransitionException(TableState oldState, TableState newState, String message) {
       this.oldState = oldState;
       this.newState = newState;
+
+      String defaultMessage = "Error transitioning from " + oldState + " state to " + newState + " state";
```
keith-turner: Could this be moved inside the else block?

server/base/src/test/java/org/apache/accumulo/server/tables/IllegalTableTransitionExceptionTest.java
jmark99: I went back and forth between null and the empty string.  Both will work. I can go with either if you have a preference.

server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
jmark99: Yes...done.

server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
```diff
@@ -115,10 +115,22 @@ public TableState getTableState(String tableId) {
...
 
     public IllegalTableTransitionException(TableState oldState, TableState newState) {
+      this(oldState, newState, "");
+    }
+
+    public IllegalTableTransitionException(TableState oldState, TableState newState, String message) {
       this.oldState = oldState;
       this.newState = newState;
+
+      if (message != null && !message.isEmpty())
```
milleruntime: Could use StringUtils.isNotEmpty() from apache commons.  Or even static import it so it just reads if(isNotEmpty(message))

server/base/src/test/java/org/apache/accumulo/server/tables/IllegalTableTransitionExceptionTest.java
keith-turner: I saw both were supported in the code.  I was thinking could have test for both.

jmark99: Added additional unit test.

server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
jmark99: Thanks, @milleruntime . I like that better.

PR #328, opened 2017-11-29
ACCUMULO-4743 Replaced general custom with tserver prefix for cache config

jkrdev:
Initial PR to allow testing done by Keith. Since I can't test locally at the moment. I realize more will most likely need to be done.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/file/blockfile/cache/BlockCacheManager.java
+++ b/core/src/main/java/org/apache/accumulo/core/file/blockfile/cache/BlockCacheManager.java
@@ -54,8 +54,8 @@ public abstract class BlockCacheManager {
      * This method provides a way for a cache implementation to access arbitrary configuration set by a user.
      *
      * <p>
-     * Returns all Accumulo properties that have a prefix of {@code tserver.cache.<prefix>.<type>.} or {@code tserver.cache.<prefix>.default.}
-     * with values for specific cache types overriding defaults.
+     * Returns all Accumulo properties that have a prefix of {@code tserver.cache.<prefix>.<type>.} or {@code tserver.cache.<prefix>.default.} with values for
+     * specific cache types overriding defaults.
      *
      * <p>
      * For example assume the following data is in Accumulo's system config.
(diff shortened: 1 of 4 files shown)


asfgit:

jkrdev:
I don't know why this fails Travis build but my local build passes when ran through uno, I guess it skips certain test. Does anyone know how I can stop that from happening? Would rather make sure it passes locally than push each time.

milleruntime:
Travis-ci runs all the tests but skipsITs.  Try running this locally:
```mvn clean verify -DskipITs```

**Comment by jkrdev** [comment]
Okay will do that, after this next push I did fails since I didn't know that beforehand. Thanks.

**Comment by jkrdev** [comment]
@keith-turner If you could test this, now that it builds, would be grateful thanks.

File: core/src/main/java/org/apache/accumulo/core/file/blockfile/cache/BlockCacheManager.java [hunk 1]
```diff
@@ -30,7 +30,7 @@
 
   private final Map<CacheType,BlockCache> caches = new HashMap<>();
 
-  public static final String CACHE_PROPERTY_BASE = Property.GENERAL_ARBITRARY_PROP_PREFIX + "cache.";
+  public static final String CACHE_PROPERTY_BASE = Property.TSERV_PREFIX + "cache.";
```
keith-turner: Inorder to avoid system cache properties from conflicting with custom cache config properties, I am thinking it would be better to make this `Property.TSERV_PREFIX + "cache.config."`.

keith-turner:
@jkrdev  I tested this with [keith-turner/accumulo-ohc](https://github.com/keith-turner/accumulo-ohc) and it worked fine.

core/src/main/java/org/apache/accumulo/core/file/blockfile/cache/BlockCacheManager.java
jkrdev: I actually had that before, I will change it and rebuild and then hopefully it works with keith-turner/accumulo-ohc .

core/src/main/java/org/apache/accumulo/core/file/blockfile/cache/BlockCacheManager.java
```diff
@@ -54,17 +54,19 @@
...
      *
      * <p>
      * For example assume the following data is in Accumulo's system config.
      *
      * <pre>
-     * general.custom.cache.lru.default.evictAfter=3600
-     * general.custom.cache.lru.default.loadFactor=.75
-     * general.custom.cache.lru.index.loadFactor=.55
-     * general.custom.cache.lru.data.loadFactor=.65
+     * tserver.cache.lru.default.evictAfter=3600
```
keith-turner: these need to be updated with `config`

jkrdev:
Great glad to hear that. alrighty.

PR #329, opened 2017-11-29
ACCUMULO-3185 Removed references to walogs directory to stop it's creation.

jkrdev:


asfgit:

PR #334, opened 2017-12-08
ACCUMULO-4758 throw correct exception when meta block absent

keith-turner:


asfgit:

milleruntime:
My approval was merely a thank you for the information so thanks!

PR #339, opened 2017-12-14
[ACCUMULO-4751] Set createdTime for status records

adamjshook:


asfgit:

adamjshook:
@joshelser Would you have a moment to review this?  Thank you!

test/src/main/java/org/apache/accumulo/test/replication/StatusMakerIT.java
```diff
@@ -246,4 +253,97 @@ public void closedMessagesCreateOrderRecords() throws Exception {
...
+    BatchWriter bw = conn.createBatchWriter(sourceTable, new BatchWriterConfig());
+    String walPrefix = "hdfs://localhost:8020/accumulo/wals/tserver+port/";
+    List<String> files = Arrays.asList(walPrefix + UUID.randomUUID(), walPrefix + UUID.randomUUID(), walPrefix + UUID.randomUUID(),
+        walPrefix + UUID.randomUUID());
+    Map<String,Integer> fileToTableId = new HashMap<>();
+
+    Status.Builder statBuilder = Status.newBuilder().setBegin(0).setEnd(0).setInfiniteEnd(true).setClosed(true);
+
+    Map<String,Long> statuses = new HashMap<>();
+    int index = 1;
```
joshelser: Just make this a `long` and change the `Integer.toString()` to `Long.toString()`?

server/master/src/main/java/org/apache/accumulo/master/replication/StatusMaker.java
```diff
@@ -197,8 +202,18 @@ protected boolean addStatusRecord(Text file, String tableId, Value v) {
   protected boolean addOrderRecord(Text file, String tableId, Status stat, Value value) {
     try {
       if (!stat.hasCreatedTime()) {
-        log.error("Status record ({}) for {} in table {} was written to metadata table which lacked createdTime", ProtobufUtil.toString(stat), file, tableId);
-        return false;
+        try {
+          long createdTime = setAndGetCreatedTime(new Path(file.toString()), tableId);
```
joshelser: Add a comment here alluding to the problem and this workaround, please?

adamjshook:
@joshelser Added a comment and changed to `long`. Ran the integration tests; a couple time outs, one of which was the `CyclicReplicationIT.dataIsNotOverReplicated` test.  May be the hardware I am running on isn't fast enough.

joshelser:
> Ran the integration tests; a couple time outs, one of which was the CyclicReplicationIT.dataIsNotOverReplicated test. May be the hardware I am running on isn't fast enough.

Hurm. I think some of these ITs magically became broken on 1.8 at some point and no one has made time to investigate them. I'd not expect your changes to be the cause of them not passing.

PR #345, opened 2017-12-20
ACCUMULO-4769 Sanity check for valid CryptoModule and KeyEncryptionStrategy in config

PircDef:
Added a check to the ConfigSanityCheck to validate the cryptomodule and the keyencryptionstrategy
Added unit tests to make sure the checks work
Removed silent failure in the CryptoModuleFactory since we have already checked for these failures when these values were set.


asfgit:

core/src/main/java/org/apache/accumulo/core/conf/ConfigSanityCheck.java
```diff
@@ -88,6 +90,31 @@ else if (!prop.getType().isValidFormat(value))
       if (key.equals(Property.CRYPTO_CIPHER_KEY_ALGORITHM_NAME.getKey())) {
         keyAlgorithm = Objects.requireNonNull(value);
       }
+      if (key.equals(Property.CRYPTO_SECRET_KEY_ENCRYPTION_STRATEGY_CLASS.getKey()) && !(value == null || value.equals("NullSecretKeyEncryptionStrategy"))) {
+        Class<?> strat = null;
+        try {
+          strat = Class.forName(value);
```
ctubbsii: We have a "ConfigurationTypeHelper" class with utility methods for loading classes from config properties. Should use that, if possible.

core/src/main/java/org/apache/accumulo/core/conf/ConfigSanityCheck.java
```diff
@@ -88,6 +91,12 @@ else if (!prop.getType().isValidFormat(value))
       if (key.equals(Property.CRYPTO_CIPHER_KEY_ALGORITHM_NAME.getKey())) {
         keyAlgorithm = Objects.requireNonNull(value);
       }
+      if (key.equals(Property.CRYPTO_SECRET_KEY_ENCRYPTION_STRATEGY_CLASS.getKey()) && !(value == null || value.equals("NullSecretKeyEncryptionStrategy"))) {
```
ctubbsii: Should this be `value.equals(NullSecretKeyEncryptionStrategy.class.getName()) || value.equals(NullSecretKeyEncryptionStrategy.class.getSimpleName())` instead of checking value by String literal?

(Similar comment below on String literal comparison for "NullCryptoModule")

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
```diff
@@ -80,8 +80,7 @@ private static CryptoModule instantiateCryptoModule(String cryptoModuleClassname
     try {
       cryptoModuleClazz = AccumuloVFSClassLoader.loadClass(cryptoModuleClassname);
     } catch (ClassNotFoundException e1) {
-      log.warn("Could not find configured crypto module \"{}\".  No encryption will be used.", cryptoModuleClassname);
-      return new NullCryptoModule();
+      throw new RuntimeException("Could not find configured crypto module " + cryptoModuleClassname);
```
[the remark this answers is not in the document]
ctubbsii: I still think this one could be a more specific RuntimeException (IllegalArgumentException?)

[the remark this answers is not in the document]
ctubbsii: This was preexisting before this change, but I also noticed that the rawtypes should not be suppressed. Generic `<?>` wildcard params should be used on `Class` instead.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
```diff
@@ -156,8 +147,7 @@ private static SecretKeyEncryptionStrategy instantiateSecreteKeyEncryptionStrate
     try {
       keyEncryptionStrategyClazz = AccumuloVFSClassLoader.loadClass(className);
     } catch (ClassNotFoundException e1) {
-      log.warn("Could not find configured secret key encryption strategy \"{}\".  No encryption will be used.", className);
-      return new NullSecretKeyEncryptionStrategy();
+      throw new RuntimeException("Could not find configured secret key encryption strategy: " + className);
```
ctubbsii: I think this could be a more specific RuntimeException also.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
```diff
@@ -96,23 +95,15 @@ private static CryptoModule instantiateCryptoModule(String cryptoModuleClassname
     }
 
     if (!implementsCryptoModule) {
-      log.warn("Configured Accumulo crypto module \"{}\" does not implement the CryptoModule interface. No encryption will be used.", cryptoModuleClassname);
-      return new NullCryptoModule();
+      throw new IllegalArgumentException("Configured Accumulo crypto module " + cryptoModuleClassname + " does not implement the CryptoModule interface.");
```
ctubbsii: I'm not sure any of these checks in this section are necessary if the class was properly loaded with the ConfigurationTypeHelper from the configuration. It may just be redundant.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
```diff
@@ -172,23 +162,16 @@ private static SecretKeyEncryptionStrategy instantiateSecreteKeyEncryptionStrate
     }
 
     if (!implementsSecretKeyStrategy) {
-      log.warn("Configured Accumulo secret key encryption strategy \"%s\" does not implement the SecretKeyEncryptionStrategy interface. No encryption will be used.");
-      return new NullSecretKeyEncryptionStrategy();
+      throw new IllegalArgumentException(
```
ctubbsii: Again, I'm not sure any of this is necessary if the class is loaded with the ConfigurationTypeHelper from the config. It may be redundant. I'm not 100% sure on this, though.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
PircDef: For this and the SecretKeyEncryptionStrategy, they might be redundant. I am not sure if anything can or would change these values both after the configuration is read and before the class is loaded here. Leaving them in for now is safer.

core/src/main/java/org/apache/accumulo/core/conf/ConfigSanityCheck.java
PircDef: NullSecretKeyEncryptionStrategy and NullCryptoModule are both package private and nested within CryptoModuleFactory. I am not sure why, so I am hesitant to change them here. It is probably worth considering making both public and moving both out of CryptoModuleFactory in a future PR.

ctubbsii: Ah, I see. It probably doesn't matter much anyway.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
ctubbsii: Agreed. This can be refactored/deferred to later. It's safer to leave it as is for now.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
```diff
@@ -146,49 +136,40 @@ public static SecretKeyEncryptionStrategy getSecretKeyEncryptionStrategy(String
...
-  @SuppressWarnings("rawtypes")
   private static SecretKeyEncryptionStrategy instantiateSecreteKeyEncryptionStrategy(String className) {
 
     log.debug("About to instantiate secret key encryption strategy {}", className);
 
     SecretKeyEncryptionStrategy strategy = null;
-    Class keyEncryptionStrategyClazz = null;
+    Class<?> keyEncryptionStrategyClazz = null;
     try {
       keyEncryptionStrategyClazz = AccumuloVFSClassLoader.loadClass(className);
```
keith-turner: Could do 

```java
Class<? extends SecretKeyEncryptionStrategy> keyEncryptionStrategyClazz;
try{
  keyEncryptionStrategyClazz = AccumuloVFSClassLoader.loadClass(className).asSubclass(SecretKeyEncryptionStrategy.class);

//would need to catch more exceptions here, but could drop code that uses reflection to look at interfaces
```

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoModuleFactory.java
```diff
@@ -146,49 +136,40 @@ public static SecretKeyEncryptionStrategy getSecretKeyEncryptionStrategy(String
...
     }
 
     if (!implementsSecretKeyStrategy) {
-      log.warn("Configured Accumulo secret key encryption strategy \"%s\" does not implement the SecretKeyEncryptionStrategy interface. No encryption will be used.");
-      return new NullSecretKeyEncryptionStrategy();
+      throw new IllegalArgumentException(
+          "Configured Accumulo secret key encryption strategy \"%s\" does not implement the SecretKeyEncryptionStrategy interface.");
     } else {
       try {
         strategy = (SecretKeyEncryptionStrategy) keyEncryptionStrategyClazz.newInstance();
```
keith-turner: Should be able to drop this cast when class type is `Class<? extends SecretKeyEncryptionStrategy>`

keith-turner:
@PircDef sorry this sat so long.  I can make the `asSubclass()` changes if needed.  After those changes are made I can merge.

keith-turner:
Merged in 416b743.  Used asSubclass in dce8c359056905405abaea9b99efc21dadbc5920

PR #347, opened 2017-12-20
ACCUMULO-4746 Fluent API for Mutation

bfach10:
This PR is incomplete. Requesting feedback on design to make sure this is a good direction for the Mutation API.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/data/Mutation.java
+++ b/core/src/main/java/org/apache/accumulo/core/data/Mutation.java
@@ -797,9 +797,9 @@ public class Mutation implements Writable {
   public interface FamilyStep extends QualifierStep {
     QualifierStep family(byte[] colFam);
 
-    QualifierStep family(CharSequence colFam);
-
     QualifierStep family(ByteBuffer colFam);
+
+    QualifierStep family(CharSequence colFam);
 
     QualifierStep family(Text colFam);
   }
@@ -815,9 +815,9 @@ public class Mutation implements Writable {
   public interface QualifierStep extends VisibilityStep {
     VisibilityStep qualifier(byte[] colQual);
 
-    VisibilityStep qualifier(CharSequence colQual);
-
     VisibilityStep qualifier(ByteBuffer colQual);
+
+    VisibilityStep qualifier(CharSequence colQual);
 
     VisibilityStep qualifier(Text colQual);
   }
@@ -833,7 +833,13 @@ public class Mutation implements Writable {
   public interface VisibilityStep extends TimestampStep {
     TimestampStep visibility(byte[] colVis);
 
+    TimestampStep visibility(ByteBuffer colVis);
+
+    TimestampStep visibility(CharSequence colVis);
+
     TimestampStep visibility(ColumnVisibility colVis);
+
+    TimestampStep visibility(Text colVis);
   }
 
   /**
@@ -856,13 +862,17 @@ public class Mutation implements Writable {
    * @since 2.0
    */
   public interface MutationStep {
-    void set(byte[] val, boolean delete);
+    Mutation set(byte[] val);
 
-    void set(ByteBuffer val);
+    Mutation set(ByteBuffer val);
 
-    void set(Value val);
+    Mutation set(CharSequence val);
 
-    void delete();
+    Mutation set(Text val);
+
+    Mutation set(Value val);
+
+    Mutation delete();
   }
 
   /**
@@ -901,7 +911,7 @@ public class Mutation implements Writable {
      *          column family length
      * @return a QualifierStep object, advancing the method chain
      */
-    public QualifierStep family(byte[] colFam, int colFamLength) {
+    private QualifierStep family(byte[] colFam, int colFamLength) {
       columnFamily = colFam;
       columnFamilyLength = colFamLength;
       return this;
@@ -914,6 +924,7 @@ public class Mutation implements Writable {
      *          column family
      * @return a QualifierStep object, advancing the method chain
      */
+    @Override
     public QualifierStep family(byte[] colFam) {
       return family(colFam, colFam.length);
     }
@@ -925,8 +936,9 @@ public class Mutation implements Writable {
      *          column family
      * @return a QualifierStep object, advancing the method chain
      */
-    public QualifierStep family(Text colFam) {
-      return family(colFam.getBytes(), colFam.getLength());
+    @Override
+    public QualifierStep family(ByteBuffer colFam) {
+      return family(ByteBufferUtil.toBytes(colFam));
     }
 
     /**
@@ -936,6 +948,7 @@ public class Mutation implements Writable {
      *          column family
      * @return a QualifierStep object, advancing the method chain
      */
+    @Override
     public QualifierStep family(CharSequence colFam) {
       return family(new Text(colFam.toString()));
     }
@@ -947,8 +960,9 @@ public class Mutation implements Writable {
      *          column family
      * @return a QualifierStep object, advancing the method chain
      */
-    public QualifierStep family(ByteBuffer colFam) {
-      return family(ByteBufferUtil.toBytes(colFam));
+    @Override
+    public QualifierStep family(Text colFam) {
+      return family(colFam.getBytes(), colFam.getLength());
     }
 
     // methods for changing the column qualifier of a Mutation
@@ -961,7 +975,7 @@ public class Mutation implements Writable {
      *          column qualifier
      * @return a VisibilityStep object, advancing the method chain
      */
-    public VisibilityStep qualifier(byte[] colQual, int colQualLength) {
+    private VisibilityStep qualifier(byte[] colQual, int colQualLength) {
       columnQualifier = colQual;
       columnQualifierLength = colQualLength;
       return this;
@@ -974,6 +988,7 @@ public class Mutation implements Writable {
      *          column qualifier
      * @return a VisibilityStep object, advancing the method chain
      */
+    @Override
     public VisibilityStep qualifier(byte[] colQual) {
       return qualifier(colQual, colQual.length);
     }
@@ -985,8 +1000,9 @@ public class Mutation implements Writable {
      *          column qualifier
      * @return a VisibilityStep object, advancing the method chain
      */
-    public VisibilityStep qualifier(Text colQual) {
-      return qualifier(colQual.getBytes(), colQual.getLength());
+    @Override
+    public VisibilityStep qualifier(ByteBuffer colQual) {
+      return qualifier(ByteBufferUtil.toBytes(colQual));
     }
 
     /**
@@ -996,6 +1012,7 @@ public class Mutation implements Writable {
      *          column qualifier
      * @return a VisibilityStep object, advancing the method chain
      */
+    @Override
     public VisibilityStep qualifier(CharSequence colQual) {
       return qualifier(new Text(colQual.toString()));
     }
@@ -1007,8 +1024,9 @@ public class Mutation implements Writable {
      *          column qualifier
      * @return a VisibilityStep object, advancing the method chain
      */
-    public VisibilityStep qualifier(ByteBuffer colQual) {
-      return qualifier(ByteBufferUtil.toBytes(colQual));
+    @Override
+    public VisibilityStep qualifier(Text colQual) {
+      return qualifier(colQual.getBytes(), colQual.getLength());
     }
 
     // methods for changing the column visibility of a Mutation
@@ -1017,10 +1035,13 @@ public class Mutation implements Writable {
      *
      * @param colVis
      *          column visibility
+     * @param colVisLen
+     *          column visibility length
      * @return a TimestampStep object, advancing the method chain
      */
-    public TimestampStep visibility(byte[] colVis) {
+    public TimestampStep visibility(byte[] colVis, int colVisLen) {
       columnVisibility = colVis;
+      columnVisibilityLength = colVisLen;
       return this;
     }
 
@@ -1031,8 +1052,57 @@ public class Mutation implements Writable {
      *          column visibility
      * @return a TimestampStep object, advancing the method chain
      */
+    @Override
+    public TimestampStep visibility(byte[] colVis) {
+      return visibility(colVis, colVis.length);
+    }
+
+    /**
+     * Modifies the column visibility of a mutation.
+     *
+     * @param colVis
+     *          column visibility
+     * @return a TimestampStep object, advancing the method chain
+     */
+    @Override
+    public TimestampStep visibility(ByteBuffer colVis) {
+      return visibility(ByteBufferUtil.toBytes(colVis));
+    }
+
+    /**
+     * Modifies the column visibility of a mutation.
+     *
+     * @param colVis
+     *          column visibility
+     * @return a TimestampStep object, advancing the method chain
+     */
+    @Override
+    public TimestampStep visibility(CharSequence colVis) {
+      return visibility(new Text(colVis.toString()));
+    }
+
+    /**
+     * Modifies the column visibility of a mutation.
+     *
+     * @param colVis
+     *          column visibility
+     * @return a TimestampStep object, advancing the method chain
+     */
+    @Override
     public TimestampStep visibility(ColumnVisibility colVis) {
       return visibility(colVis.getExpression());
+    }
+
+    /**
+     * Modifies the column visibility of a mutation.
+     *
+     * @param colVis
+     *          column visibility
+     * @return a TimestampStep object, advancing the method chain
+     */
+    @Override
+    public TimestampStep visibility(Text colVis) {
+      return visibility(colVis.toString().getBytes());
     }
 
     // methods for changing the timestamp of a Mutation
@@ -1043,6 +1113,7 @@ public class Mutation implements Writable {
      *          timestamp
      * @return a MutationStep object, advancing the method chain
      */
+    @Override
     public MutationStep timestamp(long ts) {
       hasTs = true;
       timestamp = ts;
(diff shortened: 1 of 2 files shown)


asfgit:

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,236 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+   * <p>
+   * Each step of the Mutation is represented by an interface. Inheritance order ensures that chained methods follow prescribed order of:
+   *
+   * <p>
+   * row, family, qualifier, visibility, timestamp
+   *
+   * <p>
+   * set and delete methods end the chain and finalize the mutation
+   */
+  public interface RowStep extends FamilyStep {
```
mikewalch: What happens if someone creates a mutation with a specific row and then uses `row()` with a different one?

```java
Mutation m = new Mutation("row1");
m.at().row("row2").family("fam").set("val");
```

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,236 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
     put(columnFamily, columnQualifier, columnVisibility.getExpression(), true, timestamp, true, EMPTY_BYTES);
   }
 
+  /**
+   * Fluent API for Mutation
+   *
+   * <p>
+   * Each step of the Mutation is represented by an interface. Inheritance order ensures that chained methods follow prescribed order of:
```
mikewalch: Could mention that methods need to be in specified order but are optional

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,236 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+   * set and delete methods end the chain and finalize the mutation
+   */
+  public interface RowStep extends FamilyStep {
+    FamilyStep row(byte[] r);
+
+    FamilyStep row(CharSequence r);
+
+    FamilyStep row(ByteBuffer r);
+
+    FamilyStep row(Text t);
```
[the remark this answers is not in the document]
keith-turner: Personally I think consistency is important.  As long as Text methods elsewhere are not deprecated I think it should be supported here.  For better or worse Text is widely used in the Accumulo API.  Some Accumulo methods only accept Text, which means in some cases users may already have Text objects that could also be used here.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
keith-turner: There should not be a row step.  All changes to a mutation are for a single row.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,236 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+  }
+
+  public interface FamilyStep extends QualifierStep {
```
keith-turner: All of these interfaces will need a javadoc `@since 2.0` tag.  Do not need it on the method as they will inherit it from the class level javadoc.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,236 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+    void set(byte[] v, boolean delete);
+
+    void set(ByteBuffer v);
+
+    void set(Value v);
+
+    void delete();
+  }
+
+  public RowStep at() {
```
keith-turner: This method will need a `@since 2.0` javadoc tag.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
bfach10: I wasn't sure about this one so I didn't include any implementation for it. Thanks for the clarification! I've removed the row methods and interface in my most recent commit.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
bfach10: Added to comment an explanation that methods are optional.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
bfach10: Updated the interface classes with the `@since 2.0` tag, thanks!

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
bfach10: Updated the method  with the `@since 2.0` tag, thanks!

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,380 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+  /**
+   * Provides methods for setting the column visibility of a Mutation. The user can provide the visibility as a byte array or
+   * {@link org.apache.accumulo.core.security.ColumnVisibility} object instance and the backend will do the necessary transformation.
+   *
+   * All QualifierStep methods return an instance derived from the VisibilityStep interface, allowing the methods to be semantically chained.
+   *
+   * @since 2.0
+   */
+  public interface VisibilityStep extends TimestampStep {
+    TimestampStep visibility(byte[] colVis);
```
keith-turner: There should be visibility methods that take CharSeq, Text, and ByteBuffer

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,380 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+
+  /**
+   * Provides methods for setting the value of a Mutation. The user can provide the value as a byte array, Value, or ByteBuffer object instance and the backend
+   * will do the necessary transformation.
+   *
+   * All MutationStep methods complete a fluent Mutation API method chain.
+   *
+   * @since 2.0
+   */
+  public interface MutationStep {
```
keith-turner: There should be set methods that take CharSeq, Text, and byte[]

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,380 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+    }
+
+    /**
+     * Modifies the column family of a mutation.
+     *
+     * @param colFam
+     *          column family
+     * @return a QualifierStep object, advancing the method chain
+     */
+    public QualifierStep family(ByteBuffer colFam) {
```
keith-turner: ByteBuffers are tricky because they have a mutable position.  It would be good to document that this method does not advance the position.  Would also be good to test this.

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
```diff
@@ -187,7 +187,8 @@ private Value nv(String s) {
   public void testPuts() {
     Mutation m = new Mutation(new Text("r1"));
 
-    m.put(nt("cf1"), nt("cq1"), nv("v1"));
+    m.at().family(nt("cf1")).qualifier(nt("cq1")).set(nv("v1"));
```
keith-turner: More test are needed.  Need to test all of the types and test adding multiple columns.  I would recommend doing this in separate test.  Could do something like the following.

```java
 @Test
  public void testAtTypes() {
    String fam = "f145";
    String qual = "q452";
    String val = "v52093209284";

    Mutation expected = new Mutation("row5");
    expected.put(fam, qual, val);

    //Test all family methods
    Mutation actual = new Mutation("row5");
    actual.at().family(fam).qualifier(qual).set(val);
    assertEquals(expected, actual);

    actual = new Mutation("row5");
    actual.at().family(fam.getBytes(UTF_8)).qualifier(qual).set(val);
    assertEquals(expected, actual);

    actual = new Mutation("row5");
    actual.at().family(new Text(fam)).qualifier(qual).set(val);
    assertEquals(expected, actual);

    ByteBuffer bbFam = ByteBuffer.wrap(fam.getBytes(UTF_8));
    actual = new Mutation("row5");
    actual.at().family(bbFam).qualifier(qual).set(val);
    assertEquals(expected, actual);
    //TODO ensure that bbFam position was not changed

    //Test all qualifier methods
    actual = new Mutation("row5");
    actual.at().family(fam).qualifier(qual.getBytes(UTF_8)).set(val);
    assertEquals(expected, actual);

    actual = new Mutation("row5");
    actual.at().family(fam).qualifier(new Text(qual)).set(val);
    assertEquals(expected, actual);

    ByteBuffer bbQual = ByteBuffer.wrap(qual.getBytes(UTF_8));
    actual = new Mutation("row5");
    actual.at().family(fam).qualifier(bbQual).set(val);
    assertEquals(expected, actual);
    //TODO ensure bbQual position was not changed

    //Test all visibility methods
    //TODO

    //Test all set method
    //TODO

    // The following will not compile because set returns void.  It would make these test a lot shorter if set() returned the
    // mutation.  However, not sure this is a good idea.  Would not work with subclasses like ConditionalMutation.
    assertEquals(expected, new Mutation("row5").at().family(fam).qualifier(qual).set(val));
  }

  @Test
  public void testAtMultiple() {
    ColumnVisibility cv = new ColumnVisibility("A&B");

    Mutation expected = new Mutation("row5");
    expected.put("fam123", "q77", cv, "v4224");
    expected.put("fam123", "q47", 56, "v345");
    expected.put("fam999", "q77", "v1");
    expected.putDelete("fam89", "q69");

    Mutation actual = new Mutation("row5");
    actual.at().family("fam123").qualifier("q77").visibility(cv).set("v4224");
    actual.at().family("fam123").qualifier("q47").timestamp(56).set("v345");
    actual.at().family("fam999").qualifier("q77").set("v1");
    actual.at().family("fam89").qualifier("q69").delete();

    assertEquals(expected, actual);
  }
```

Writing these test made me wonder if the `set()` methods should have a return type of `Mutation`.  I am not sure about this, need to think through the implications for ConditionalMutation.

bfach10: For sure needs more testing. I'm going to try out the design of having the `set()` methods return `Mutation` in this next commit, in addition to more tests. Thank you for the great start you've given me here.

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
```diff
@@ -183,6 +185,190 @@ private Value nv(String s) {
...
+    final String qual = "q1pm2";
+    final String val = "v8672194923750";
+
+    Mutation expected = new Mutation("row5");
+    expected.put(fam, qual, val);
+
+    // Test all family methods, keeping qual and val constant as Strings
+    // fam: byte[]
+    Mutation actual = new Mutation("row5");
+    actual.at().family(fam.getBytes(UTF_8)).qualifier(qual).set(val);
```
joshelser: This makes me uncomfortable seeing Mutation be mutable. Can the fluent API be applied as a MutationBuilder instead of to the Object we're creating? I think that would also remove the need for the (to me) odd `at()` method.

e.g.

`Mutation actual = Mutation.newBuilder("row5").family(fam.getBytes(UTF_8)).qualifier(qual).set(val);`

bfach10: To make `Mutation` immutable we'd have to remove the `put()` methods since those allow us to change the `Mutation`'s `UnsynchronizedBuffer`. We'd also have to remove `ConditionalMutation`'s ability to add conditions. 

The consensus in the JIRA ticket and PR comments seems to be that `row` should be immutable but other aspects are fair game to change after object creation.

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
bfach10: Idea: add a new group of `set` methods called `conditionalSet` that return a `ConditionalMutation` object. This will enable users to continue the chain and add conditions as well.

`public ConditionalMutation conditionalSet(byte[] val) { }`

`ConditionalMutation actual = new ConditionalMutation("row5");
actual.at().family("fam123").qualifier("q77").visibility(cv)
.conditionalSet("v4224").addCondition(new Condition())`

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
joshelser: Oh durr. Sorry, I was thinking about ServerMutation (I think is the name of it). Are you aware of that sub-class?

Is there a reason to not make the fluent-methods you want to add on a new class, e.g. MutationBuilder?

keith-turner: > Is there a reason to not make the fluent-methods you want to add on a new class, e.g. MutationBuilder?

I was puzzling over this question and shared what I came up with on the issue.

https://issues.apache.org/jira/browse/ACCUMULO-4746?focusedCommentId=16269423&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16269423

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
keith-turner: Where were you thinking of adding the `conditionalSet()` method?

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
bfach10: Since the goal (in my view at least) is to have the method chain be easy to use semantically for the caller, using a new class could be a way to go. As long as the API call stays streamlined like below and the user doesn't have to necessarily be aware of this new class in order to make modifications on their `Mutation`.

`m.at().family().qualifier().visibility().set()`

So in the case of a new class the `at` method could return a new class instead of the nested interface class it's returning now.

joshelser: > I was puzzling over this question and shared what I came up with on the issue.

Thanks for the context, Keith.

> As long as the API call stays streamlined like below and the user doesn't have to necessarily be aware of this new class in order to make modifications on their Mutation.

Ya, that's what I was thinking. Less junk in one Java file. That has the indirect benefit of preventing unintended changes to the existing public API of Mutation.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,474 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+    Mutation delete();
+  }
+
+  /**
+   * Returns a new FamilyStep object allowing the user to define changes to the Mutation.
+   *
+   * @return a new FamilyStep object, advancing the method chain
+   * @since 2.0
+   */
+  public FamilyStep at() {
```
joshelser: I'm also curious about this choice of name (after seeing some of the other suggestions from the JIRA comments).

`add()` was the first thing I thought of. But, Mutation is essentially a map of "column updates" for a row. `put()` may make more sense (aligning with the underlying implementation -- for multiple updates to the same column in a Mutation, I think the last update would "win").

core/src/test/java/org/apache/accumulo/core/data/MutationTest.java
keith-turner: > Less junk in one Java file.

The new interfaces and implementation may not have to be in the the Mutation file.  The only thing that has to be there is the `at()` method.

keith-turner: > That has the indirect benefit of preventing unintended changes to the existing public API of Mutation.

In general we should be using tools to check for API changes between releases.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
keith-turner: I was wondering if `set()` should be changed to `put()` for consistency with existing API, not sure.  Would look like the following.

```java
  mutation.at().family(fam).qualifier(qual).put(val)
```

Personally I don't think changing `at()` to `put()` reads nicely.   Would look like the following.

```java
mutation.put().family(fam).qualifier(qual).set(val)
```

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,474 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
     put(columnFamily, columnQualifier, columnVisibility.getExpression(), true, timestamp, true, EMPTY_BYTES);
   }
 
+  /**
```
keith-turner: The two asterisk at the beginning of the comment indicate that this is javadoc.  However the javadoc block is not associated with a method or class. So I am not sure where this would render when the javadoc html is created.

Could convert it to a regular comment if your intention was that it was just a code comment.  If you want it in the javadoc, maybe move the content to the javadoc for the `at()` method.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,474 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+     * Finalizes the method chain by filling the buffer with the gathered Mutation configuration
+     *
+     * @param val
+     *          value
+     * @param delete
+     *          deletion flag
+     */
+    private Mutation set(byte[] val, boolean delete) {
+      if (buffer == null) {
+        throw new IllegalStateException("Can not add to mutation after serializing it");
```
keith-turner: I ran the test with code coverage analysis and only two spots were missed in this new inner class. One was here.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -772,6 +772,474 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier, ColumnVisibil
...
+      }
+
+      // indicate if this is a deletion
+      put(delete);
+
+      // fill buffer with value
+      if (val.length < VALUE_SIZE_COPY_CUTOFF) {
+        put(val, val.length);
+      } else {
+        if (values == null) {
```
keith-turner: This case was not covered by the unit test.  Would probably be worthwhile to try adding a large value.

mikewalch:
@bfach10, this is pretty nice addition to Accumulo. It would be great to get it committed. Let me know if you need any help.

mikewalch:
Thanks @bfach10 for the work so far!  I created a new pull request (#590) that addresses comments from the code review on this PR.  I am going to close this PR. Please comment on #590.

PR #349, opened 2017-12-23
[ACCUMULO-4775] Skip negate column entirely in ColumnAgeOffFilter

adamjshook:


core/src/main/java/org/apache/accumulo/core/iterators/user/ColumnAgeOffFilter.java
```diff
@@ -44,8 +44,10 @@ public TTLSet(Map<String,String> objectStrings) {
         String column = entry.getKey();
         String ttl = entry.getValue();
```
keith-turner: I was thinking about the case of ignoring only whitespace also. Could trim here like the following to do that.

```java
String ttl  = entry.getValue().trim();
```

adamjshook: Good idea!  Pushed the `trim`.

shell/src/test/java/org/apache/accumulo/shell/commands/SetIterCommandTest.java
```diff
@@ -0,0 +1,114 @@
...
+    Field field = reader.getClass().getSuperclass().getDeclaredField("out");
+    field.setAccessible(true);
+    field.set(reader, out);
+
+    // Parsing iterator options
+    reader.flush();
+    EasyMock.expectLastCall().times(3);
+
+    EasyMock.expect(shellState.getReader()).andReturn(reader);
+    EasyMock.expect(reader.readLine(EasyMock.anyObject(String.class))).andReturn(""); // Default of 'false'
```
keith-turner: Is this when the shell ask for the negate option?  If so would be nice to comment for each of these line what the shell is asking for.

adamjshook: Yup -- pushed some comments.

keith-turner:
I am going merge this

PR #350, opened 2017-12-27
ACCUMULO-4528 Add import/export table info to docs

jmark99:
ACCUMULO-4328 importtable and exporttable deserve descriptions in the user manual

Updated 1.8 and 2.0 user manual documentation to include the import/export example within the
documentation directly rather than requiring a user to click away from the manual pages to a
different project (i.e., the accumulo-examples project) for that information.


jmark99:
Yep. I built 1.8 and then loaded the updated user manual into my local browser and things looked as expected.

PR #351, opened 2017-12-28
ACCUMULO-1975 Refactor Property.createInstanceFromPropertyName into AccumuloConfiguration.instantiateClassProperty

jkrdev:


jkrdev:
Oh ha okay I saw that Property.createInstanceFromPropertyName was only in 3 places so I just assumed. I will go through and review those.

jkrdev:
I believe I have replaced all the code in the appropriate places as well as removed AccumuloConfiguration.instantiateClassProperty method from the class.

PR #364, opened 2018-01-24
ACCUMULO-4778 Initial feedback for table name to id mapping cache

milleruntime:
Looking for some initial feedback.  Running Sunny tests now but haven't attempted to measure performance yet.

Class generates table id to name mapping every second.  Previously, a mapping (of all table ids) was generated every time a single table id or name look up was called.


core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -37,66 +35,25 @@
...
   public static final String VALID_NAME_REGEX = "^(\\w+\\.)?(\\w+)$";
+  public static final Long TABLE_MAP_CACHE_EXPIRATION = TimeUnit.SECONDS.toNanos(1L);
+  private static final AtomicLong tableMapTimestamp = new AtomicLong(System.nanoTime());
 
   private static final SecurityPermission TABLES_PERMISSION = new SecurityPermission("tablesPermission");
   private static final AtomicLong cacheResetCount = new AtomicLong(0);
+  private static volatile TableMap tableMapCache;
 
-  private static ZooCache getZooCache(Instance instance) {
+  public static ZooCache getZooCache(Instance instance) {
```
keith-turner: Could make this package private

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -147,6 +118,7 @@ public static void clearCache(Instance instance) {
     cacheResetCount.incrementAndGet();
     getZooCache(instance).clear(ZooUtil.getRoot(instance) + Constants.ZTABLES);
     getZooCache(instance).clear(ZooUtil.getRoot(instance) + Constants.ZNAMESPACES);
+    tableMapCache = null;
```
keith-turner: There is also a clearCacheByPath method where this should be set to null

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -129,12 +86,26 @@ public static String getTableName(Instance instance, String tableId) throws Tabl
...
+    }
+    return map.getNameToIdMap();
   }
 
-  public static SortedMap<String,String> getIdToNameMap(Instance instance) {
-    return getMap(instance, false);
+  public static Map<String,String> getIdToNameMap(Instance instance) {
+    TableMap map = tableMapCache;
+    // check if map needs to be updated
+    if (map == null || tableMapTimestamp.longValue() + TABLE_MAP_CACHE_EXPIRATION <= System.nanoTime()) {
```
keith-turner: Could make the TableMap class have a final long creationTime field instead of having the mutable tableMapTimestamp.  Then the CPU may be able to avoid two trips to main memory.  Also I think keeping it all together and immutable makes it easier to reason about.

Could also do a double check to avoid multiple threads regenerating and put all of this in a single function.

```java
private static TableMap getTableMap() {
  TableMap map = tableMapCache;
  if (map == null || map.getCreationTime() + TABLE_MAP_CACHE_EXPIRATION <= System.nanoTime()) {
    synchronized(Tables.class) {
       map = tableMapCache;
       if (map == null || map.getCreationTime() + TABLE_MAP_CACHE_EXPIRATION <= System.nanoTime()) {
         map = tableMapCache = new TableMap(instance);  //this could set creation time, should do that last in constructor
       }     
    }
  }
  return map;
}
```

core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
```diff
@@ -0,0 +1,97 @@
...
+ * Used for thread safe caching of table ID maps
+ */
+public class TableMap {
+  private static final Logger log = LoggerFactory.getLogger(TableMap.class);
+
+  private Map<String,String> tableNameToIdMap;
+  private Map<String,String> tableIdtoNameMap;
+
+  public TableMap(Instance instance) {
+    generateMaps(instance);
```
keith-turner: If this code is inlined in the constructor then the two maps could be made final.

core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  private void generateMaps(Instance instance) {
+    ZooCache zc = getZooCache(instance);
```
keith-turner: An alternative to calling getZooCache here, is to pass the ZooCache into the constructor.

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
keith-turner: If the ZooCache is passed to the TableMap constructor, then this could go back to private.

core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
milleruntime: Good call.

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
milleruntime: I had thought about this as well but wasn't sure where to put it in TableMap.  You have this replacing the constructor?

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
milleruntime: Good catch, thank you!

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
keith-turner: I was thinking of a new static method in Tables class

core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
milleruntime: I like that, thanks.

milleruntime:
Changes so far are causing errors in TCompactProxyIT because it reuses table names.  I think we need a unit test to pin point where exactly the cache needs to cleared where it did not have to previously.

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -129,12 +85,29 @@ public static String getTableName(Instance instance, String tableId) throws Tabl
...
+   * Create a new TableMap if needed and cache it locally. The creation of the TableMap is synchronized so only one thread will generate the map. Added for
+   * ACCUMULO-4778.
+   */
+  private static TableMap getTableMap(Instance instance) {
+    TableMap map = tableMapCache;
+    if (map == null || map.getCreationTime() + TABLE_MAP_CACHE_EXPIRATION <= System.nanoTime()) {
+      synchronized (Tables.class) {
+        map = tableMapCache;
+        if (map == null || map.getCreationTime() + TABLE_MAP_CACHE_EXPIRATION <= System.nanoTime()) {
+          map = tableMapCache = new TableMap(instance, getZooCache(instance));
```
keith-turner: probably need a TableMap per instance

core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Used for thread safe caching of immutable table ID maps. See ACCUMULO-4778.
+ */
+public class TableMap {
+  private static final Logger log = LoggerFactory.getLogger(TableMap.class);
+
+  private final Map<String,String> tableNameToIdMap;
+  private final Map<String,String> tableIdToNameMap;
+  private final Long creationTime;
```
keith-turner: this can be dropped now.

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -37,64 +36,43 @@
...
-              namespaceName = Namespaces.getNamespaceName(instance, namespaceId);
-              namespaceIdToNameMap.put(namespaceId, namespaceName);
+    try {
+      return instanceToZooCache.get(instance.getInstanceID(), new Callable<ZooCache>() {
+        @Override
+        public ZooCache call() {
+          return new ZooCacheFactory().getZooCache(zks, timeOut, new Watcher() {
+            @Override
+            public void process(WatchedEvent watchedEvent) {
+              instanceToMapCache.invalidateAll();
```
keith-turner: Could only clear for the instance id

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -144,9 +141,9 @@ public static boolean exists(Instance instance, String tableId) {
   }
 
   public static void clearCache(Instance instance) {
-    cacheResetCount.incrementAndGet();
     getZooCache(instance).clear(ZooUtil.getRoot(instance) + Constants.ZTABLES);
     getZooCache(instance).clear(ZooUtil.getRoot(instance) + Constants.ZNAMESPACES);
+    instanceToMapCache.invalidateAll();
```
keith-turner: could only clear for the instance id

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -158,17 +155,9 @@ public static void clearCache(Instance instance) {
...
-    if (zooPath.startsWith("/")) {
-      thePath = zooPath;
-    } else {
-      thePath = "/" + zooPath;
-    }
-
+    String thePath = zooPath.startsWith("/") ? zooPath : "/" + zooPath;
     getZooCache(instance).clear(ZooUtil.getRoot(instance) + thePath);
-
+    instanceToMapCache.invalidateAll();
```
keith-turner: could clear just the instance id

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
milleruntime: Good call.

milleruntime:
Latest cache implementation is looking promising.  Sunny day ITs were passing on Friday.  I am now looking if I can use my JMH benchmark repo to measure any performance gains.

milleruntime:
Measuring this branch against 1.7.3 using MiniAccumuloCluster with 100 tables.  I created 3 tests that call Tables.getTableId(), tableOperations.tableIdMap() and tableOperations.exists().  Each of these methods are commonly used and shouldn't hinder performance with 2 being from the API and one a common internal method. 

1.7.3 Results:
Benchmark                     Mode  Cnt     Score     Error  Units
TablesBenchmark.getExists    thrpt   50  5293.911 ± 127.073  ops/s
TablesBenchmark.getTableId   thrpt   50  5356.971 ± 127.876  ops/s
TablesBenchmark.getTableMap  thrpt   50  5239.343 ± 137.264  ops/s

ACCUMULO-4778 branch:
Benchmark                     Mode  Cnt        Score       Error  Units
TablesBenchmark.getExists    thrpt   50   917886.356 ± 16067.502  ops/s
TablesBenchmark.getTableId   thrpt   50  1696546.994 ± 26953.792  ops/s
TablesBenchmark.getTableMap  thrpt   50  1718777.835 ± 21612.689  ops/s

Test: TablesBenchmark.getExists showed 17238.5% increase in ops/s
Test: TablesBenchmark.getTableId showed 31569.9% increase in ops/s
Test: TablesBenchmark.getTableMap showed 32705.2% increase in ops/s
Total average 27171.21% increase in ops/s

milleruntime:
Running the benchmark tests with multiple threads only shows a greater improvement.  I think this is ready to merge.

milleruntime:
For future reference, I pushed the tests I created to my accumulo-benchmark repo: https://github.com/milleruntime/accumulo-benchmark/blob/master/jmh-test/src/main/java/org/apache/accumulo/TablesBenchmark.java

PR #365, opened 2018-01-25
ACCUMULO-4779 Avoid locks in ZooCache when data in cache

keith-turner:
ZooCache was using read and write locks.  For the case where lots
of threads were accessing data present in the cache the read locks
were really slowing things down.  This commit switches to immutable
copies of all the data present in the cache which require no locks
to access.  When the cache changes the immutable copies are
regenerated and then made available.


keith-turner:
merged in cf9e754b045e3fac452df282bcf4ec97974038a0

PR #368, opened 2018-01-26
ACCUMULO-4777: fix log messages

ivakegg:
Reverted to only log the warning message after attempt 1.  Fixed log messages to display 1 based attempt number.


server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -338,9 +338,10 @@ private void write(final Collection<CommitSession> sessions, boolean mincFinish,
           success = (currentLogSet == logSetId.get());
         }
       } catch (DfsLogger.LogClosedException ex) {
-        log.debug("Logs closed while writing, retrying attempt " + writeRetry.retriesCompleted());
+        log.debug("Logs closed while writing, retrying attempt " + (writeRetry.retriesCompleted() + 1));
       } catch (Exception t) {
-        log.warn("Failed to write to WAL, retrying attempt " + writeRetry.retriesCompleted(), t);
+        if (writeRetry.retriesCompleted() != 0)
```
keith-turner: Could do something like `if(writeRetry.retriesCompleted() % 10 == 0)`  or `if(writeRetry.retriesCompleted() % 100 == 0)`.  

With the current change, I think it would be nice if the log message indicated that subsequent retires will not be logged.  If logging periodically it would be nice to indicate will log again after X failures.

ivakegg: ok, I will see what I can do in this regard.

fate/src/test/java/org/apache/accumulo/fate/zookeeper/RetryTest.java
```diff
@@ -147,4 +154,344 @@ public void testUnlimitedRetry() {
...
+      // now observe what log messages we got which should be around 5 +- 1
+      Assert.assertTrue(i > 10);
+      Assert.assertTrue(testLogger.getMessages().size() >= 4 && testLogger.getMessages().size() <= 6);
+    } finally {
+      Retry.setLogger(LoggerFactory.getLogger(Retry.class));
+    }
+
+  }
+
+  private static class TestLogger implements Logger {
```
keith-turner: Could possibly use EasyMock to create a mock instance that expects warn and/or debug to be called a certain number of times.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -338,9 +338,10 @@ private void write(final Collection<CommitSession> sessions, boolean mincFinish,
           success = (currentLogSet == logSetId.get());
         }
       } catch (DfsLogger.LogClosedException ex) {
-        log.debug("Logs closed while writing, retrying attempt " + writeRetry.retriesCompleted());
+        writeRetry.logRetry("Logs closed while writing");
```
keith-turner: When I saw this code is made me think about also logging when the message is different.

fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
```diff
@@ -125,12 +149,51 @@ public long retriesCompleted() {
...
+
+  public void logRetry(String message, Throwable t) {
+    // log the first time, and then after every logInterval
+    if (lastRetryLog < 0 || (System.currentTimeMillis() - lastRetryLog) > logInterval) {
+      log.warn(getMessage(message), t);
+      lastRetryLog = System.currentTimeMillis();
+    }
+  }
+
+  public void logRetry(String message) {
```
keith-turner: Random thought, not advocating for this.  In 2.0.0-SNAP which uses java 8, could pass in a lambda that does the actual logging.  This code just decides whether or not to call it.  Not sure, but this may remove the need to inspect the stack trace to get info about caller.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
ivakegg: yea, I thought about that but also thought that perhaps that is going to far.

fate/src/test/java/org/apache/accumulo/fate/zookeeper/RetryTest.java
ivakegg: true....I was initially thinking about actually examining the contents of the message which is why I went this route....I can rework it however

fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
ivakegg: interesting thought

fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
```diff
@@ -100,10 +110,24 @@ void setMaxWait(long maxWait) {
     this.maxWait = maxWait;
   }
 
+  // Visible for testing
```
keith-turner: Guava has a visible for testing annotation that could be used here.

fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
keith-turner: Another possibility is to have the caller pass in a logger like `logRetry(Logger log, String message)`.  If this is done, would the `getCaller()` method be needed?  Also the `setLogger()` method would not be needed for testing.

fate/src/test/java/org/apache/accumulo/fate/zookeeper/RetryTest.java
keith-turner: You may be able to check expected parameters with easymock.

fate/src/main/java/org/apache/accumulo/fate/zookeeper/RetryFactory.java
```diff
@@ -21,9 +21,11 @@
  */
 public class RetryFactory {
   public static final long DEFAULT_MAX_RETRIES = 10l, DEFAULT_START_WAIT = 250l, DEFAULT_WAIT_INCREMENT = 250l, DEFAULT_MAX_WAIT = 5000l;
-  public static final RetryFactory DEFAULT_INSTANCE = new RetryFactory(DEFAULT_MAX_RETRIES, DEFAULT_START_WAIT, DEFAULT_WAIT_INCREMENT, DEFAULT_MAX_WAIT);
+  public static final long DEFAULT_LOG_INTERVAL = 10 * 60 * 1000;
```
keith-turner: This default seems a bit long to me.

fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
ivakegg: I like that

fate/src/main/java/org/apache/accumulo/fate/zookeeper/RetryFactory.java
ivakegg: What were you thinking?  1m, 5m, 42s ?

keith-turner: 1 to 3 minutes

ivakegg:
Commits squashed and commited into 1.7, 1.8, and master

PR #370, opened 2018-01-30
ACCUMULO-4772 Update shell to use NewTableConfiguration methods

jmark99:
ACCUMULO-4772 Update Accumulo shell to utilize new NewTableConfiguration methods

Added capability to pre-configure iterators and locality groups when creating
new tables via the Accumulo shell. Updated CreateTableCommand.java to accept locality group
and iterator data. Updated ShellServerIT.java to test the new capabilities.

### DIFF
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
@@ -242,8 +242,7 @@ public class CreateTableCommand extends Command {
   }
 
   /**
-   * Validate that the provided scope arguments are valid iterator scope settings.
-   * Checking for duplicate entries and invalid scope values.
+   * Validate that the provided scope arguments are valid iterator scope settings. Checking for duplicate entries and invalid scope values.
    *
    * Returns an EnumSet of scopes to be set.
    */
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
@@ -199,8 +199,8 @@ public class CreateTableCommand extends Command {
    * Used in conjunction with createtable shell command to allow an iterator to be configured upon table creation.
    */
   private NewTableConfiguration attachIteratorToNewTable(final CommandLine cl, final Shell shellState, NewTableConfiguration ntc) {
-    EnumSet<IteratorScope> scopeEnumSet = null;
-    IteratorSetting iteratorSetting = null;
+    EnumSet<IteratorScope> scopeEnumSet;
+    IteratorSetting iteratorSetting;
     if (shellState.iteratorProfiles.size() == 0)
       throw new IllegalArgumentException("No shell iterator profiles have been created.");
     String[] options = cl.getOptionValues(createTableOptIteratorProps.getOpt());
@@ -233,8 +233,7 @@ public class CreateTableCommand extends Command {
           scopeEnumSet = EnumSet.allOf(IteratorScope.class);
         } else {
           // 'all' is not involved, examine the scope arguments and populate iterator scope EnumSet
-          validateScopes(scopeArgs);
-          scopeEnumSet = addScopeArgsToIteratorEnumSet(scopeArgs);
+          scopeEnumSet = validateScopes(scopeArgs);
         }
       }
       ntc.attachIterator(iteratorSetting, scopeEnumSet);
@@ -243,35 +242,22 @@ public class CreateTableCommand extends Command {
   }
 
   /**
-   * Validate that the provided scope arguments are valid iterator scope settings. Checking for duplicate entries and invalid scope values.
+   * Validate that the provided scope arguments are valid iterator scope settings.
+   * Checking for duplicate entries and invalid scope values.
+   *
+   * Returns an EnumSet of scopes to be set.
    */
-  private void validateScopes(final List<String> scopeList) {
-    Set<String> dupes = new HashSet<>();
-    for (String scope : scopeList) {
-      if (dupes.contains(scope))
-        throw new IllegalArgumentException("duplicate scope argument provided.");
-      dupes.add(scope);
+  private EnumSet<IteratorScope> validateScopes(final List<String> scopeList) {
+    EnumSet<IteratorScope> scopes = EnumSet.noneOf(IteratorScope.class);
+    for (String scopeStr : scopeList) {
       try {
-        IteratorScope.valueOf(scope);
+        IteratorScope scope = IteratorScope.valueOf(scopeStr);
+        if (!scopes.add(scope))
+          throw new IllegalArgumentException("duplicate scope arguments found");
       } catch (IllegalArgumentException ex) {
-        throw new IllegalArgumentException("iterator scope value is invalid/missing or contains spaces.", ex);
+        throw new IllegalArgumentException("illegal scope arguments: " + ex.getMessage(), ex);
       }
     }
-  }
-
-  /**
-   * Assign iterator scope arguments to an IteratorUntil.IteratorScope EnumSet for use with NewTableConfiguration object.
-   */
-  private EnumSet<IteratorScope> addScopeArgsToIteratorEnumSet(final List<String> scopeList) {
-    EnumSet<IteratorScope> scopes = EnumSet.noneOf(IteratorScope.class);
-    if (scopeList.contains("scan"))
-      scopes.add(IteratorScope.scan);
-    if (scopeList.contains("minc"))
-      scopes.add(IteratorScope.minc);
-    if (scopeList.contains("majc"))
-      scopes.add(IteratorScope.majc);
-    if (scopes.isEmpty())
-      throw new IllegalArgumentException("supplied scope values are invalid.");
     return scopes;
   }
 
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
@@ -39,6 +39,7 @@ import org.apache.accumulo.core.client.impl.Tables;
 import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.core.constraints.VisibilityConstraint;
 import org.apache.accumulo.core.iterators.IteratorUtil;
+import org.apache.accumulo.core.iterators.IteratorUtil.IteratorScope;
 import org.apache.accumulo.shell.Shell;
 import org.apache.accumulo.shell.Shell.Command;
 import org.apache.accumulo.shell.ShellUtil;
@@ -198,7 +199,7 @@ public class CreateTableCommand extends Command {
    * Used in conjunction with createtable shell command to allow an iterator to be configured upon table creation.
    */
   private NewTableConfiguration attachIteratorToNewTable(final CommandLine cl, final Shell shellState, NewTableConfiguration ntc) {
-    EnumSet<IteratorUtil.IteratorScope> scopeEnumSet = null;
+    EnumSet<IteratorScope> scopeEnumSet = null;
     IteratorSetting iteratorSetting = null;
     if (shellState.iteratorProfiles.size() == 0)
       throw new IllegalArgumentException("No shell iterator profiles have been created.");
@@ -217,7 +218,7 @@ public class CreateTableCommand extends Command {
       // handle case where only the profile is supplied. Use all scopes by default if no scope args are provided.
       if (parts.length == 1) {
         // add all scopes to enum set
-        scopeEnumSet = EnumSet.allOf(IteratorUtil.IteratorScope.class);
+        scopeEnumSet = EnumSet.allOf(IteratorScope.class);
       } else {
         // user provided scope arguments exist, parse them
         List<String> scopeArgs = Arrays.asList(parts[1].split(","));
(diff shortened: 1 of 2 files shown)


milleruntime:
Looks good to me.  Does the command line help print the new options?  I've worked with the shell some but can't remember how it prints the usage.

jmark99:
Yep, I updated the createtable help usage to display the new options.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -2020,4 +2022,143 @@ public void testSummarySelection() throws Exception {
...
+  }
+
+  /**
+   * Due to the existing complexity of the createtable command, the createtable help only displays an example of setting one locality group. It is possible to
+   * set multiple groups if needed. This test verifies that capability.
+   */
+  @Test
+  public void testCreateTableWithMultipleLocalityGroups() throws Exception {
+    final String table = name.getMethodName();
+    ts.exec("createtable " + table + " -l locg1=fam1,fam2 locg2=colfam1", true);
```
keith-turner: I am surprised the space between `fam2` and `logg2=` does not throw things off.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -174,13 +250,20 @@ public Options getOptions() {
...
-
     createTableOptCopyConfig.setArgName("table");
     createTableOptCopySplits.setArgName("table");
     createTableOptSplit.setArgName("filename");
     createTableOptFormatter.setArgName("className");
     createTableOptInitProp.setArgName("properties");
 
+    createTableOptLocalityProps = new Option("l", "locality", true, "create locality groups at table creation");
+    createTableOptLocalityProps.setArgName("group=col_fam[,col_fam]");
+    createTableOptLocalityProps.setArgs(Option.UNLIMITED_VALUES);
```
keith-turner: Does this enables parsing of the space separated values?

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -2020,4 +2022,143 @@ public void testSummarySelection() throws Exception {
     // check that there are two files, with none having extra summary info
     assertMatches(output, "(?sm).*^.*total[:]2[,]\\s+missing[:]0[,]\\s+extra[:]0.*$.*");
   }
+
```
keith-turner: Very nice test!!

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -2020,4 +2022,143 @@ public void testSummarySelection() throws Exception {
...
+   * Due to the existing complexity of the createtable command, the createtable help only displays an example of setting one iterator upon table creation. It is
+   * possible to set multiple if needed. This test verifies that capability.
+   */
+  @Test
+  public void testCreateTableWithMultipleIterators() throws Exception {
+    final String tmpTable = "tmpTable";
+    final String table = name.getMethodName();
+
+    // create iterator profile
+    // Will use tmpTable for creating profile since setshelliter is requiring a table
```
keith-turner: It would be nice to look into fixing this.

jmark99: I agree. I created a ticket for it this morning (https://issues.apache.org/jira/browse/ACCUMULO-4791)

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
jmark99: The space enables splitting of the parameters so that multiple groups can be parsed easily. It mimics the format of the setgroups command: 

setgroups <group>=<col fam>{,<col fam>}{ <group>=<col fam>{,<col fam>}} [-?] [-t <table>].

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -150,9 +168,67 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
...
+      String profileName = parts[0];
+      IteratorSetting iteratorSetting = shellState.iteratorProfiles.get(profileName).get(0);
+      if (iteratorSetting == null)
+        throw new IllegalArgumentException("Provided iterator profile, " + profileName + ", does not exist");
+      // parse scope info
+      List<String> scopeList = Arrays.asList(parts[1].split(","));
+      if (scopeList.size() > 3) // max of three scope settings allowed
+        throw new IllegalArgumentException("Too many scopes supplied");
+      EnumSet<IteratorUtil.IteratorScope> scopes = EnumSet.noneOf(IteratorUtil.IteratorScope.class);
+      if (scopeList.contains("all") || scopeList.contains("scan"))
```
keith-turner: Will this code would ignore an unknown scope?  For example if I gave the scopes `scan,nimc`, I think it would not complain about `nimc`.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Yes.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -150,9 +168,67 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
...
+    String[] options = cl.getOptionValues(createTableOptLocalityProps.getOpt());
+    for (String localityInfo : options) {
+      final String parts[] = localityInfo.split("=", 2);
+      if (parts.length < 2)
+        throw new IllegalArgumentException("Missing '=' or there are spaces between entries");
+      final String groupName = parts[0];
+      final HashSet<Text> colFams = new HashSet<>();
+      for (String family : parts[1].split(","))
+        colFams.add(new Text(family.getBytes(Shell.CHARSET)));
+      localityGroupMap.put(groupName, colFams);
```
keith-turner: A sanity check to see if the if the group name already exist in the map would be nice.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Yep. I missed that. I'll add some error checking to verify the scopes are one of the acceptable values.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Will add sanity check.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
jmark99: Wish i could take credit, but this was already in the test :)

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
keith-turner: Neat, I didn't know about this capability of the option parser.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
keith-turner: Oh I meant tests, all of them together are a good verification of the new options.

jmark99: Thanks!

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Added check to verify group names cannot be duplicated.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Added check to verify only valid scope arguments are accepted.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -150,9 +169,83 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
...
+      String[] parts = profileInfo.split(":", 2);
+      if (parts.length < 2)
+        throw new IllegalArgumentException("Missing scope or there are spaces between parameters");
+      // get profile name
+      String profileName = parts[0];
+      IteratorSetting iteratorSetting = shellState.iteratorProfiles.get(profileName).get(0);
+      if (iteratorSetting == null)
+        throw new IllegalArgumentException("Provided iterator profile, " + profileName + ", does not exist");
+      // parse scope info
+      List<String> scopeList = Arrays.asList(parts[1].split(","));
```
keith-turner: I would validate by calling `IteratorUtil.IteratorScope.valueOf()` for each item in scopeList.  Could lower case the string before passing it to valueOf.  I think valueOf will throw an exception if its not a valid enum.  Can also add what valueOf returns to the set.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -150,9 +169,83 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
...
+    String[] options = cl.getOptionValues(createTableOptLocalityProps.getOpt());
+    for (String localityInfo : options) {
+      final String parts[] = localityInfo.split("=", 2);
+      if (parts.length < 2)
+        throw new IllegalArgumentException("Missing '=' or there are spaces between entries");
+      final String groupName = parts[0];
+      final HashSet<Text> colFams = new HashSet<>();
+      for (String family : parts[1].split(","))
+        colFams.add(new Text(family.getBytes(Shell.CHARSET)));
+      if (localityGroupMap.containsKey(groupName))
```
keith-turner: Instead of calling containsKey, could check that put returns null.  However the current code is probably more readable and it does not need to be efficient.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: @keith-turner, I refactored the attachIteratorToNewTable method . Take a look and let me know your thoughts. I made use of IteratorUtil.IteratorScope.valueOf() for the scopeList check and rearranged the order of several checks to handle more error conditions. I didn't make use lowercase as all arguments for other shell methods use lowercase and It seemed more consistent to require users to provide lowercase parameters. If you feel strongly about handling mixed case args you can let me know and I can update.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -150,9 +168,115 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
...
+          throw new IllegalArgumentException("Too many scope arguments supplied");
+        // handle the 'all' argument separately since it is not an allowable enum value for IteratorScope
+        // if 'all' is used, it should be the only scope provided
+        if (scopeArgs.contains("all")) {
+          if (scopeArgs.size() > 1)
+            throw new IllegalArgumentException("Cannot use 'all' in conjunction with other scopes");
+          scopeEnumSet = EnumSet.allOf(IteratorUtil.IteratorScope.class);
+        } else {
+          // 'all' is not involved, examine the scope arguments and populate iterator scope EnumSet
+          validateScopes(scopeArgs);
```
keith-turner: I wasn't sure if EnumSets were mutable, so I did an experiment and they are.  Therefore this code could be simplified to use what `valueOf` returns to insert into the enum set.

```java
    EnumSet<IteratorScope> scopes = EnumSet.noneOf(IteratorScope.class);

    List<String> scopeStrs = new ArrayList<>();
    scopeStrs.add("scan");
    scopeStrs.add("minc");
    scopeStrs.add("scan");

    for (String scopeStr : scopeStrs) {
      IteratorScope scope = IteratorScope.valueOf(scopeStr);
      if (!scopes.add(scope)) {
        System.out.println("dupe " + scope);
      }
    }
    System.out.println(scopes);
  }
```

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
```diff
@@ -26,6 +30,7 @@
 
 import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.AccumuloSecurityException;
+import org.apache.accumulo.core.client.IteratorSetting;
```
keith-turner: You can also import the IteratorScope inner enum to make code shorter.  

```java
import org.apache.accumulo.core.iterators.IteratorUtil.IteratorScope;
```

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
keith-turner: I agree with you, accepting mixed case args should be avoided.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Added import and shortened up the affected lines.

shell/src/main/java/org/apache/accumulo/shell/commands/CreateTableCommand.java
jmark99: Implemented the valueOf change and as a result was able to validate and return the EnumSet within the one validateScopes method. Removed the AddScopeArgsToEnumSet method

jmark99:
hi @keith-turner, have you had a chance to review the latest updates? Are there any other comments/suggestions on the ticket or is it ready to merge?

PR #374, opened 2018-02-02
ACCUMULO-4790 Avoid creating FileRef

keith-turner:


milleruntime:
Could you give a description of the solution?  Seems different than the one purposed in the JIRA ticket.

keith-turner:
When I got into the code I noticed it only needed to create these file refs in the case of bulk loaded files with logical time set.  So it seemed quick to check for that case only create file refs if those types of files exists.

milleruntime:
Gotcha!  Looks good to me.  I like the simple solution.

PR #378, opened 2018-02-07
ACCUMULO-4800 Cache parsing of iterator config

keith-turner:


server/base/src/main/java/org/apache/accumulo/server/conf/TableConfiguration.java
```diff
@@ -144,4 +163,50 @@ public String toString() {
...
+      return tableOpts;
+    }
+
+    public String getContext() {
+      return context;
+    }
+  }
+
+  public ParsedIteratorConfig getParsedIteratorConfig(IteratorScope scope) {
+    long count = getUpdateCount();
```
mikewalch: I see that the iterator config will change if this count changes.  When does this count change?

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/ScanDataSource.java
```diff
@@ -186,12 +188,28 @@ public boolean isCurrent() {
...
     if (!loadIters) {
       return visFilter;
-    } else if (null == options.getClassLoaderContext()) {
-      return iterEnv.getTopLevelIterator(IteratorUtil.loadIterators(IteratorScope.scan, visFilter, tablet.getExtent(), tablet.getTableConfiguration(),
-          options.getSsiList(), options.getSsio(), iterEnv));
     } else {
-      return iterEnv.getTopLevelIterator(IteratorUtil.loadIterators(IteratorScope.scan, visFilter, tablet.getExtent(), tablet.getTableConfiguration(),
-          options.getSsiList(), options.getSsio(), iterEnv, true, options.getClassLoaderContext()));
+      List<IterInfo> iterInfos;
+      Map<String,Map<String,String>> iterOpts;
```
mikewalch: Can you describe what is going on in this block?

server/base/src/main/java/org/apache/accumulo/server/conf/TableConfiguration.java
keith-turner: I added some comments to the AccumuloConfiguration super class.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/ScanDataSource.java
keith-turner: I added some comments

PR #382, opened 2018-02-09
ACCUMULO-4782 switch session manager to a concurrent map

keith-turner:


server/tserver/src/main/java/org/apache/accumulo/tserver/session/SessionManager.java
```diff
@@ -70,20 +84,20 @@ public void run() {
     };
 
     SimpleTimer.getInstance(conf).schedule(r, 0, Math.max(maxIdle / 2, 1000));
+
   }
 
-  public synchronized long createSession(Session session, boolean reserve) {
+  public long createSession(Session session, boolean reserve) {
```
mikewalch: Might be nice to have an initial state for Sessions and check that it is in that state here.

keith-turner: I added a NEW state and checked it in this method

PR #383, opened 2018-02-13
ACCUMULO-4809 Avoid blocking during session cleanup

keith-turner:


server/tserver/src/main/java/org/apache/accumulo/tserver/session/SessionManager.java
```diff
@@ -169,36 +169,36 @@ public Session removeSession(long sessionId, boolean unreserve) {
...
-        long configuredIdle = maxIdle;
-        if (session instanceof UpdateSession) {
-          configuredIdle = maxUpdateIdle;
-        }
-        long idleTime = System.currentTimeMillis() - session.lastAccessTime;
-        if (idleTime > configuredIdle && !session.reserved) {
-          log.info("Closing idle session from user=" + session.getUser() + ", client=" + session.client + ", idle=" + idleTime + "ms");
-          iter.remove();
-          sessionsToCleanup.add(session);
+    synchronized (idleSessions) {
```
mikewalch: why do you have a synchronized block on `idleSessions` with the other blocks removed?  Is this to prevent multiple interleaving calls to sweep?

keith-turner: > Is this to prevent multiple interleaving calls to sweep?

Yeah, that is the purpose.  I think in the current code that only one thread will ever call sweep at a time, but I added the sync just to be safe.  I will add a comment about this.

PR #384, opened 2018-02-13
ACCUMULO-4789 Avoid creating expensive thread name

keith-turner:


milleruntime:
I know there was some uncertainty with changing this debug information in 1.8... was there something that brought you to a decision?  I am just wondering what your final thoughts were...

keith-turner:
> I know there was some uncertainty with changing this debug information in 1.8... was there something that brought you to a decision? I am just wondering what your final thoughts were...

It was the magnitude of the performance impact.  For lots of really small lookups I was seeing the code spend 10% to 15% of its time creating this thread name.  With this change, it hardly shows up.  So my thinking was the thread names still has some useful info, but its much faster to create now.

PR #392, opened 2018-02-26
Fixes ACCUMULO-4832 WAL warnings

ivakegg:
… subsequent messages are warnings.

### DIFF
--- a/fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
+++ b/fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
@@ -32,7 +32,8 @@ public class Retry {
   private long maxRetries, maxWait, waitIncrement;
   private long retriesDone, currentWait;
 
-  private long logInterval;
+  private static long NANO_SEC_PER_MILLI_SEC = 1000000L;
+  private long logIntervalNanoSec;
   private long lastRetryLog;
 
   /**
@@ -54,7 +55,7 @@ public class Retry {
     this.waitIncrement = waitIncrement;
     this.retriesDone = 0l;
     this.currentWait = startWait;
-    this.logInterval = logInterval;
+    this.logIntervalNanoSec = (logInterval * NANO_SEC_PER_MILLI_SEC);
     this.lastRetryLog = -1;
   }
 
@@ -126,11 +127,11 @@ public class Retry {
 
   // Visible for testing
   void setLogInterval(long logInterval) {
-    this.logInterval = logInterval;
+    this.logIntervalNanoSec = logInterval * NANO_SEC_PER_MILLI_SEC;
   }
 
   public long getLogInterval() {
-    return logInterval;
+    return logIntervalNanoSec / NANO_SEC_PER_MILLI_SEC;
   }
 
   public boolean canRetry() {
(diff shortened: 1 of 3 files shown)


fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
```diff
@@ -166,18 +166,26 @@ protected void sleep(long wait) throws InterruptedException {
...
-      lastRetryLog = System.currentTimeMillis();
+      lastRetryLog = now;
     }
   }
 
   public void logRetry(Logger log, String message) {
-    // log the first time, and then after every logInterval
-    if (lastRetryLog < 0 || (System.currentTimeMillis() - lastRetryLog) > logInterval) {
+    // log the first time as debug, and then after every logInterval as a warning
+    long now = System.currentTimeMillis();
```
ctubbsii: I strongly recommend using `System.nanoTime()` for elapsed time. `System.currentTimeMillis()` is for clock time, which can go backwards and result in a surprising negative difference between the two values.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -338,10 +338,9 @@ private void write(final Collection<CommitSession> sessions, boolean mincFinish,
           success = (currentLogSet == logSetId.get());
         }
       } catch (DfsLogger.LogClosedException ex) {
-        writeRetry.logRetry(log, "Logs closed while writing");
+        writeRetry.logRetry(log, "Logs closed while writing", ex);
```
keith-turner: Seems like the warning will still happen during normal operations?

fate/src/main/java/org/apache/accumulo/fate/zookeeper/Retry.java
ivakegg: ok, will do

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
keith-turner: I don't have a good feel for how many times the code retries in this situation.  I wonder if the warning will show up when there is not a problem.

ivakegg: I bet your continuous ingest will tell you.

ivakegg: Now that the first message is debug, I can guarantee this is no worse that what we had before in terms of logging.

keith-turner: > I bet your continuous ingest will tell you.

yeah, it will.  I think the fix looks good.  Testing will tell for sure.

keith-turner: In the current test I am running, seeing the following on the monitor.  Noticed there was one time where the retry attempt was 3.

accumulo-4832

ivakegg:
Changes applied to 1.7, and merged into 1.8 and master

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
keith-turner: These were the final counts from the test I was running against 1.7.4 RC0.  

accumulo-4832-2

PR #394, opened 2018-03-01
ACCUMULO-4836 make online table always wait

keith-turner:


EdColeman:
Does this revert the behavior that ACCUMULO-4574 was trying to address - if the table is online, return and do not wait for additional actions? The original behavior would block if there was a fate operation on the table and online was called - the original fix was to not block and return because the table was online and no additional actions were required to be in online.

If the table is online - why do we need to wait?

keith-turner:
With this change, if the table is already online then the FATE operation is not called.  However if the method parameter `wait == true` then it will scan the metadata table waiting for all tablets to be assigned. The table below shows the new behavior. 

| Table state | Wait parameter | Fate called | Scan metadata table unil all tablets assigned
| ---------------|----------------------|-----------------|----------------------------------------------------------
| ONLINE | true | No | Yes
| ONLINE | false | No | No
| OFFLINE | true | Yes | Yes
| OFFLINE | false | Yes | No

Scanning the metadata table until all tablet assigned is a client side operation that does not block on FATE locks.

keith-turner:
Before the change the behavior was :

| Table state | Wait parameter | Fate called | Scan metadata table unil all tablets assigned
| ---------------|----------------------|-----------------|----------------------------------------------------------
| ONLINE | true | No | **No**
| ONLINE | false | No | No
| OFFLINE | true | Yes | Yes
| OFFLINE | false | Yes | No

keith-turner:
@EdColeman I would like to make a decision on this issue for 1.7.4.  Did I answer your question and do you have any concerns?

keith-turner:
merged in c4f3685

PR #395, opened 2018-03-05
ACCUMULO-4835 Make TableOps throw TableNotFound

milleruntime:


core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
```diff
@@ -1072,6 +1072,8 @@ private void waitForTableStateTransition(String tableId, TableState expectedStat
         if (Tables.getTableState(context.getInstance(), tableId) != expectedState) {
           if (!Tables.exists(context.getInstance(), tableId))
             throw new TableDeletedException(tableId);
+          if (Tables.getTableState(context.getInstance(), tableId) == TableState.DELETING)
```
keith-turner: After clearing the cache could store the return value of getTableState in a local var (could call it `currentState`).  Then could avoid two calls to getTableState after clearing the cache.

milleruntime: I thought about that but then was thinking that the state could change in that time.

keith-turner: To properly account for changes between the times the two calls are made a loop would be needed.  However the wider code is in a loop, so its not needed here.

milleruntime: OK I gotcha.  I didn't fully grasp what this loop and method was doing in the first place.

core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
```diff
@@ -1069,14 +1069,17 @@ private void waitForTableStateTransition(String tableId, TableState expectedStat
...
           if (!Tables.exists(context.getInstance(), tableId))
             throw new TableDeletedException(tableId);
+          if (currentState == TableState.DELETING)
+            throw new TableNotFoundException(tableId, "", "Table is being deleted.");
           throw new AccumuloException("Unexpected table state " + tableId + " " + Tables.getTableState(context.getInstance(), tableId) + " != " + expectedState);
         }
       }
 
-      Range range = new KeyExtent(new Text(tableId), null, null).toMetadataRange();
+      Range range;
```
milleruntime: I also noticed this redundant assignment and removed it.

keith-turner: Good catch

PR #402, opened 2018-03-16
ACCUMULO-4615: Updated get status for thread safety and with a per-task timeout

jschmidt10:
Changes to the status threadpool used by Master#StatusThread. Big changes are:

1) Use a timeout per status gathering task (instead of a timeout for the entire pool)
2) Changed the status gather results to a threadsafe data structure (ConcurrentSkipListMap)
3) Added separate property for the status timeout (per tserver)


keith-turner:
https://issues.apache.org/jira/browse/ACCUMULO-4615

server/master/src/test/java/org/apache/accumulo/master/TimeoutTaskExecutorTest.java
```diff
@@ -0,0 +1,134 @@
...
+    private final long timeout;
+
+    public DummyTask(String result, long timeout) {
+      this.result = result;
+      this.timeout = timeout;
+    }
+
+    @Override
+    public String call() throws Exception {
+      Thread.sleep(timeout);
```
keith-turner: try changing this code to 

```java
    public String call() throws Exception {
      try {
        Thread.sleep(timeout);
        Thread.sleep(timeout);
      }
      return result;
    }
```

This simulates badly behaved code (some hadoop code used to swallow thread interrupts and keep going, not sure if it still does).  This will cause a test to fail because a non-slow task is canceled.

jschmidt10: Thanks Keith. Made updates to only complete tasks after they've been started, ensuring that the misbehaving tasks won't fail others before they have a chance to run.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -121,25 +123,43 @@ public void onTimeout(TimeoutCallback<C> timeoutCallback) {
...
    * @throws InterruptedException
-   *           If interrupted while awaiting callable results.
+   *           If interrupted while awaiting trackingCallable results.
    */
   public void complete() throws InterruptedException {
     Preconditions.checkState(successCallback != null, "Must set a success callback before completing " + this);
     Preconditions.checkState(exceptionCallback != null, "Must set an exception callback before completing " + this);
     Preconditions.checkState(timeoutCallback != null, "Must set a timeout callback before completing " + this);
 
+    while (hasUnfinishedTasks()) {
```
keith-turner: It seems like this loop can burn CPU, could sleep a fraction of the timeout.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -152,15 +172,24 @@ public void close() {
...
+    public Future<T> future;
+
+    public WrappedTask(C callable) {
       this.callable = callable;
-      this.future = future;
+    }
+
+    @Override
+    public T call() throws Exception {
+      hasStarted = true;
```
keith-turner: What do you think about recording the time here instead of having a boolean?  Then that could be used to determine how long to wait for the task.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: Good call. I opted to use a BlockingQueue so I didn't have to implement my own waiting.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: Sounds good!

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+    public Future<T> future;
+
+    WrappedTask(C callable) {
+      this.callable = callable;
+    }
+
+    @Override
+    public T call() throws Exception {
+      startedTasks.put(this);
+      endTime = timeout + System.currentTimeMillis();
```
keith-turner: Need to set `endTime` before adding self to startedTask.  Its possible that the task could be taken off the blocking queue before endTime is set.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+      if (!wt.hasCompleted) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  private void completeTask(WrappedTask wt) throws InterruptedException {
+    try {
+      long waitTime = wt.endTime - System.currentTimeMillis();
```
keith-turner: For sanity, I think it would be nice to have a getEndTime() method that ensures that endTime was set. Something like

```java
long getEndTime(){
  long et = endTime; //only read volatile once
  Preconditions.checkState(et > 0);
  return et;
}
```

Its possible that System.currentTimeMillis() can go backwards which would result in a negative `waitTime`.  I would normally recommend using System.nanotime() to avoid this.  However I was reading that nanoTime may only be good for use within the same thread/cpu, but I think this was only true on older OS.  This code would call nanotime in diff threads.  I think using nanoTime and/or handling negative waitTime would be good.  If its negative could just assume a waitTime of zero.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+   * @throws IllegalStateException If all of the callbacks were not registered before calling this method.
+   * @throws InterruptedException  If interrupted while awaiting callable results.
+   */
+  public void complete() throws InterruptedException {
+    Preconditions.checkState(successCallback != null, "Must set a success callback before completing " + this);
+    Preconditions.checkState(exceptionCallback != null, "Must set an exception callback before completing " + this);
+    Preconditions.checkState(timeoutCallback != null, "Must set a timeout callback before completing " + this);
+
+    while (hasUnfinishedTasks()) {
+      completeTask(startedTasks.take());
```
keith-turner: When I looked at this code, I asked myself "Can the thread ever get stuck in take() forever?"  I can not see how this could happen, however I have not been able to prove to myself that it can't happen.  Could switch this loop to poll for a limited time (like a second), but I am not sure if this needs to be done. I have always wanted a way to close a blocking queue.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    wrappedTasks.clear();
+  }
+
+  private boolean hasUnfinishedTasks() {
+    for (WrappedTask wt : wrappedTasks) {
```
keith-turner: `wrappedTasks` is not a sync list so if this method runs in a different thread than calls to `submit()` then changes may not be seen.  Could `wrappedTask` be replaced with an AtomicLong?  Increment the long before adding to executor and make `call()` decrement in try with finally block.  This method could check if the AtomicLong is greater than zero.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -0,0 +1,235 @@
...
+ * @param <C> The type of Callable submitted to this executor.
+ */
+@NotThreadSafe
+public class TimeoutTaskExecutor<T, C extends Callable<T>> implements AutoCloseable {
+
+  private final static Logger log = LoggerFactory.getLogger(TimeoutTaskExecutor.class);
+
+  private final long timeout;
+  private final ExecutorService executorService;
+  private final BlockingQueue<WrappedTask> startedTasks;
```
keith-turner: I like how you solved the busy wait problem by adding this queue.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
ctubbsii: This hasn't been an issue on Linux since very old Linux 2.6 kernels prior to 2.6.18, which itself hasn't shipped since RHEL5, and even then, the problem occurred only on some hardware. The java source code uses `CLOCK_MONOTONIC` if available, and the same source as `System.currentTimeMillis()` if not available (to millisecond resolution in the latter case).

Absolute times can be negative, but elapsed time / durations / time spans, should always be positive. If not, it's an OS bug. I seriously doubt this is something we need to worry about on modern Linux or Mac/BSD systems.

I seem to recall debating this when similar reservations were raised a few years ago when I suggested using nanoTime more back then. It wasn't a current know bug then on any systems we support, and it isn't now. I think it's fine to use it.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: Thanks.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
keith-turner: > I like keeping it simpler as it solves the immediate use case for the Master status collections.

I agree with that.   I didn't realize only a single thread would access wrappedTask.  I am ok with leaving as is.  I was thinking the counter may be more efficient than repeatedly iterating over the list.  However the list is less bug prone.  I am ok with leaving as is.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: I know what you mean. I can make it a poll(timeout) as a safety/sanity catch.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: Will update to nanoTime(). Thanks for the info guys. Learned something new :)

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1074,60 +1077,92 @@ private long balanceTablets() {
...
+          log.error("Exception occurred while getting status from " + task.server, e);
+          handleBadServer(task.server);
+        }
+      });
+
+      executor.onTimeout(new TimeoutCallback<GetTServerStatus>() {
+        @Override
+        public void accept(GetTServerStatus task) {
+          log.warn("Timed out while fetching status from " + task.server);
+          handleBadServer(task.server);
```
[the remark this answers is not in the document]
jschmidt10: Yes, that's true. I wasn't sure what to do here. Should we just continue on without the status instead?

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: Nice catch. Thanks.

server/master/src/main/java/org/apache/accumulo/master/Master.java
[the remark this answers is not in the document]
keith-turner: > Should we just continue on without the status instead?

That depends on what the downstream behavior is.  For example would doing that negatively impact balancing?  I will look into this to get a sense of what the ripple effects might be.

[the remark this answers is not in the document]
keith-turner: I looked into this and what `gatherTableInformation()` returns is eventually passed to the balancer.  So omitting a live, temporarily unresponsive tserver would not be good.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: So I left it not thread safe but migrated to the counter. It'll reduce the complexity of the unfinished check by an order of magnitude. Definitely seems worth it.

ctubbsii:
I'm getting ready to put together a release candidate for 1.9.0; reading through the history, I'm not sure this one is going to make it... if it doesn't make it in 1.9.0, I think it should be bumped to 2.0, because it adds a new feature.

keith-turner:
I have a half baked idea about the balancing issue, I will try to write it up tomorrow for discussion.

ctubbsii:
1.9.1 is fine for a target. It may depend on how the final implementation changes the public API or other user-facing components, though. It may also depend on how risky/disruptive the final implementation is.

keith-turner:
I prototyped my half baked idea. The prototype can be found in commit keith-turner@5044241ca445a0bbc48e38931f09f77d8acb44b9  which is in the branch [keith-turner/ACCUMULO-4615](https://github.com/keith-turner/accumulo/tree/ACCUMULO-4615).  I am not proposing we do this, just posting it for consideration.  The concept needs to be tested on a busy cluster.

I say this idea is half baked because I have never used the thrift send and recv method separately.  I suspect it may work well, but I am not sure.  I think in the case where a tablet server is using THsHaServer it may work really well because there is a dedicated thread to accept connections (even if all threads in the servers thread pool for processing connections are busy).  However in the case where the tserver uses the TThreadPoolServer (this is used then Kerberos is used and it can also be configured) there is no dedicated thread to accept connections, so the send call from the master may block if all server threads are busy (I am not sure).  A dedicated thrift server and port for getting status in tablet servers would solve this issue.

A completely different way of solving this issue may be to use the thrift async client which uses non-blocking sockets for connections.  However I Am not sure how well this would work with SSL and Kerberos and I have never used the thrift async client.  I am interested in learning more about it though, but have not yet found the time to do so.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
```diff
@@ -0,0 +1,249 @@
...
+    this.timeoutInNanos = TimeUnit.MILLISECONDS.toNanos(timeoutInMillis);
+    this.wrappedTasks = new ArrayList<>(expectedNumCallables);
+  }
+
+  /**
+   * Submits a new task to the executor.
+   *
+   * @param callable Task to run
+   */
+  public void submit(C callable) {
```
keith-turner: It would be nice to add a sanity check that makes this fail if complete() was called or is being called()

ctubbsii:
Resolved conflicts in PR with updated 1.8 branch.

server/master/src/main/java/org/apache/accumulo/master/TimeoutTaskExecutor.java
jschmidt10: Agreed. Will fix.

jschmidt10:
@keith-turner RE: Your new thrift server. I'm not really too savvy with thrift. I'll have to do some poking around to understand it more. But the idea sounds reasonable to me. We still have the open question about what is appropriate when we cannot reach a tserver for its status in a timely manner (the onTimeout() case). 

What do you think about tracking "notReached" tservers in this scenario and delaying balancing rather than incrementing the badServers count (so we don't eventually kill it)?

jschmidt10:
@keith-turner Touching base on this. Do you have any opinions on the suggestion in my last comment? Thanks!

jschmidt10:
OBE'd by cached thread pool

PR #435, opened 2018-04-19
Replace anonymous types with lambdas

mikewalch:
I am doing this on a type by type basis as not every type/usage may make sense to convert

### DIFF
--- a/server/master/src/main/java/org/apache/accumulo/master/tableOps/LoadFiles.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/tableOps/LoadFiles.java
@@ -28,7 +28,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.Set;
-import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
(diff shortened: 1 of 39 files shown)


server/master/src/main/java/org/apache/accumulo/master/tableOps/LoadFiles.java
```diff
@@ -157,39 +156,36 @@ private static synchronized ExecutorService getThreadPool(Master master) {
...
+            HostAndPort server = null;
+            try {
+              // get a connection to a random tablet server, do not prefer cached connections
+              // because
+              // this is running on the master and there are lots of connections to tablet servers
+              // serving the metadata tablets
+              long timeInMillis = master.getConfiguration()
+                  .getTimeInMillis(Property.MASTER_BULK_TIMEOUT);
+              // Pair<String,Client> pair = ServerClient.getConnection(master, false,
+              // timeInMillis);
```
milleruntime: Could remove commented code and format comments above nicely.

mikewalch: Fixed in 1902c40d880111d6

mikewalch:
@ctubbsii, yes it is an automated IntelliJ refactor feature

PR #442, opened 2018-04-26
When filtering in config, include values

alerman:
It is not uncommon that the value of a config setting is what I am looking for. I often end up searching through the code for the property name when the filter could just check the value as well. 

I thought about adding a seperate parameter, but the description for filter made it feel as though this was the expected behavior anyway.

### DIFF
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ConfigCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ConfigCommand.java
@@ -189,10 +189,13 @@ public class ConfigCommand extends Command {
 
       for (Entry<String,String> propEntry : acuconf) {
         final String key = propEntry.getKey();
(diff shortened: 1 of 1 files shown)


shell/src/main/java/org/apache/accumulo/shell/commands/ConfigCommand.java
```diff
@@ -70,7 +70,7 @@ public void registerCompletion(final Token root,
   @Override
   public int execute(final String fullCommand, final CommandLine cl, final Shell shellState)
       throws AccumuloException, AccumuloSecurityException, TableNotFoundException, IOException,
-      ClassNotFoundException, NamespaceNotFoundException {
```
alerman: Meant to put this in its own commit. We arent throwing this anywhere so i removed it to clean up a a bit.

PR #455, opened 2018-05-01
Fixes #414 - Create Connector for new user from existing

### LINKED ISSUE #414: Can not create Connector from existing Connector

mikewalch:
* Also added method to retrieve ConnectionInfo from Connector

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/Connector.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/Connector.java
@@ -314,6 +314,23 @@ public abstract class Connector {
   public abstract ReplicationOperations replicationOperations();
 
   /**
+   * @return {@link ConnectionInfo} which contains information about Connection to Accumulo
+   * @since 2.0.0
+   */
+  public abstract ConnectionInfo info();
+
+  /**
+   * Creates new Connector for new user with principal and token
+   *
+   * @param principal User name/principal
+   * @param token Authentication token
+   *
+   * @return Connector for new user
+   * @since 2.0.0
+   */
+  public abstract Connector createConnector(String principal, AuthenticationToken token) throws AccumuloSecurityException, AccumuloException;
+
+  /**
    * Builds ConnectionInfo after all options have been specified
    *
    * @since 2.0.0
--- a/core/src/main/java/org/apache/accumulo/core/client/Connector.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/Connector.java
@@ -320,17 +320,6 @@ public abstract class Connector {
   public abstract ConnectionInfo info();
 
   /**
-   * Creates new Connector for new user with principal and token
-   *
-   * @param principal User name/principal
-   * @param token Authentication token
-   *
-   * @return Connector for new user
-   * @since 2.0.0
-   */
-  public abstract Connector createConnector(String principal, AuthenticationToken token) throws AccumuloSecurityException, AccumuloException;
-
-  /**
    * Builds ConnectionInfo after all options have been specified
    *
    * @since 2.0.0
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/accumulo/core/client/Connector.java
```diff
@@ -313,6 +313,23 @@ public abstract ConditionalWriter createConditionalWriter(String tableName,
...
+   */
+  public abstract ConnectionInfo info();
+
+  /**
+   * Creates new Connector for new user with principal and token
+   *
+   * @param principal User name/principal
+   * @param token Authentication token
+   *
+   * @return Connector for new user
```
keith-turner: I would write `another user` instead of `new user`

core/src/main/java/org/apache/accumulo/core/client/Connector.java
```diff
@@ -313,6 +313,23 @@ public abstract ConditionalWriter createConditionalWriter(String tableName,
...
   public abstract ReplicationOperations replicationOperations();
 
+  /**
+   * @return {@link ConnectionInfo} which contains information about Connection to Accumulo
+   * @since 2.0.0
+   */
+  public abstract ConnectionInfo info();
+
+  /**
+   * Creates new Connector for new user with principal and token
```
keith-turner: Could also replace `new user` with `another user`

core/src/main/java/org/apache/accumulo/core/client/Connector.java
```diff
@@ -313,6 +313,23 @@ public abstract ConditionalWriter createConditionalWriter(String tableName,
    */
   public abstract ReplicationOperations replicationOperations();
 
+  /**
+   * @return {@link ConnectionInfo} which contains information about Connection to Accumulo
+   * @since 2.0.0
+   */
+  public abstract ConnectionInfo info();
```
keith-turner: this seems nice, however it exposes authentication info.

core/src/main/java/org/apache/accumulo/core/client/Connector.java
```diff
@@ -313,6 +313,23 @@ public abstract ConditionalWriter createConditionalWriter(String tableName,
...
+  /**
+   * Creates new Connector for new user with principal and token
+   *
+   * @param principal User name/principal
+   * @param token Authentication token
+   *
+   * @return Connector for new user
+   * @since 2.0.0
+   */
+  public abstract Connector createConnector(String principal, AuthenticationToken token) throws AccumuloSecurityException, AccumuloException;
```
keith-turner: Another possible way to accomplish this is via the builder.

```java
  Connector existing = ...
  Connector newC = Connector.builder().usingConnectionInfo(existing.info()).usingPassword(user,pass).build();
```

core/src/main/java/org/apache/accumulo/core/client/Connector.java
mikewalch: Not sure how to get around that

core/src/main/java/org/apache/accumulo/core/client/Connector.java
mikewalch: Update PR to do this 368c222c543eba07

core/src/main/java/org/apache/accumulo/core/client/Connector.java
mikewalch: Removed this method 368c222c543eba07

core/src/main/java/org/apache/accumulo/core/client/Connector.java
mikewalch: Method was removed 368c222c543eba07

PR #488, opened 2018-05-16
fixes #469 made bulk import scan split tolerant

### LINKED ISSUE #469: Make bulk import metadata scan split tolerant
This is follow on work for #436.  The metadata scan that is done for loading files should be split tolerant.  The scan should check that the metadata table form a linked list and back up when it does not.  The GC scans the metadata table like this.

keith-turner:
* Made MetadataScanner check structure and retry when linked list broken
 * Removed TabletIterator
 * Made code that used TabletIterator use MetadataScanner


core/src/main/java/org/apache/accumulo/core/metadata/schema/MetadataConsistencyCheckIterator.java
```diff
@@ -0,0 +1,153 @@
...
+    if (!curr.getTableId().equals(prev.getTableId())) {
+      if (prev.getEndRow() != null) {
+        log.debug("Non-null end row for last tablet in table: " + prev.getExtent() + " "
+            + curr.getExtent());
+        return false;
+      }
+
+      if (curr.getPrevEndRow() != null) {
+        log.debug("First tablet for table had prev end row {} {} ", prev.getExtent(),
+            curr.getExtent());
```
milleruntime: How do you know you are testing the first tablet here?

core/src/test/java/org/apache/accumulo/core/metadata/schema/MetadataConsistencyCheckIteratorTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+  @Test
+  public void testHole() {
+
+    List<TabletMetadata> tablets1 = Arrays.asList(create("4", null, "f"), create("4", "f", "m"),
+        create("4", "r", "x"), create("4", "x", null));
+    List<TabletMetadata> tablets2 = Arrays.asList(create("4", null, "f"), create("4", "f", "m"),
+        create("4", "m", "r"), create("4", "r", "x"), create("4", "x", null));
+
+    check(tablets2, new IterFactory(tablets1, tablets2));
```
milleruntime: Exception doesn't get thrown here, only debug message?  I guess I'm wondering how/when you determine to throw exception or just fail quietly?

keith-turner: When a problem is detected with the last tablet that was returned, then I throw an exception.  When a problem is detected with the current tablet that has not yet been returned I would retry.

core/src/main/java/org/apache/accumulo/core/metadata/schema/MetadataConsistencyCheckIterator.java
keith-turner: The table ids changed.

keith-turner:
I am seeing WALSunnyDayIT fail with this PR

keith-turner:
Merged in commit 9feb5e1

PR #509, opened 2018-06-01
ACCUMULO-3510 Updated scheduler to use comparators for sessions

bbux-atg:
Acting as a proxy for @ marco polo to port changes from custom fork.  Currently fails to pass verify due to conflict between format and checkstyle plugin.  Format plugin will unwrap multi line declaration which makes checkstyle complain about line being too long.


keith-turner:
@bbux-atg  thanks for the contribution.  I have one reservation about this change and that is exposing internal types like Session to user pluggable code.  Developers currently change internal types as needed and my not be aware that they are breaking plugins.  I think it would be best to introduce a new type that the pluggable code compares.  If you would like I can make a pass at introducing this type, just let me know.

bbux-atg:
@keith-turner I am trying to port this over on behalf of other users.  I will defer to you for the best approach.  Please proceed at a first pass with the alternate type. Thanks.

keith-turner:
Functionality added in #549 #571

PR #510, opened 2018-06-01
ACCUMULO-4074

bbux-atg:
Depends on https://github.com/apache/accumulo/pull/509. Acting as a proxy for @ marco polo to port changes from custom fork. Currently fails to pass verify due to conflict between format and checkstyle plugin. Format plugin will unwrap multi line declaration which makes checkstyle complain about line being too long.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/session/DefaultSessionComparator.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/session/DefaultSessionComparator.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.accumulo.tserver.session;
+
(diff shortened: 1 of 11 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/session/DefaultSessionComparator.java
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.accumulo.tserver.session;
+
+public class DefaultSessionComparator extends SessionComparator {
+
+  @Override
+  public int compareSession(Session sessionA, Session sessionB) {
```
keith-turner: I did not carry this forward in #549 for the following reasons :
 * Its goal is not documented and its hard to understand the goal looking at the code.
 * Its seems to prioritize newer scans with an exception for scans with "large" idle times.  However, I think for two scans that have never run it should prioritize the oldest.  But this goes back to the goal not being documented.
  * I think there may be overflow bugs with casting of long to int, but maybe this was intentional?? Not sure.

server/tserver/src/main/java/org/apache/accumulo/tserver/session/DefaultSessionComparator.java
```diff
@@ -0,0 +1,67 @@
...
+    if (!(sessionA.lastExecTime == -1 && sessionB.lastExecTime == -1)) {
+      if (comparison >= 0) {
+        long idleTimeA = currentTime - sessionA.lastExecTime;
+
+        /*
+         * If session B is newer, let's make sure that we haven't reached the max idle time, where
+         * we have to begin aging A
+         */
+        if (idleTimeA > sessionA.maxIdleAccessTime) {
+          comparison = -1 * Long.valueOf(idleTimeA - maxIdle).intValue();
```
keith-turner: It seems that if we are here, then we know that
 * `idleTimeA > sessionA.maxIdleAccessTime`
 * `sessionA.maxIdleAccessTime >= maxIdle`

Therefore it seems that `idleTimeA - maxIdle` will always be positive, unless there is some expectation with casting long to int??

bbux-atg:
closing as the changes are updated in https://github.com/apache/accumulo/pull/549

PR #526, opened 2018-06-09
 ACCUMULO-4074 WIP

keith-turner:
Posting this for review, made following changes on top of #510.  

 * Created API for prioritizing scans that avoids plugins accessing internal types.
 * Made it possible to configure a comparator for each scan executor.  Was previously a single comparator type for all executors.
 * Made it possible to pass options for comparator creation
 * Made it possible to configure a per table dispatcher.  This dispatcher chooses which scan executor to use.
 * Fixed comparator type casting.  I don't think the comparators in #510 worked because the Runnables were never the type expected.  Also, abstracted all of this away from a user who may write a comparator.

This is still a WIP.  Still need to document, test, and do some more cleanup.  However it is working and I am happy with the overall structure.


keith-turner:
Replaced by #549

PR #539, opened 2018-06-22
Fix WAL race condition between zookeeper and metadata table

keith-turner:
I noticed this race condition while looking in to #535.  This seems like a very low probability event because multiple conditions would need to be met in a small time window for data loss to occur as a result of this race.

 * Only tablet A references WAL_X in its otherLogs set
 * Thread T1 one clears otherLogs for tablet A 
 * Thread T2 calls markUnusedWALs() removing WAL_X from zookeeper
 * Tablet server dies (tablet data that was in WAL_X is not in metadata table or recovery data)
 * Thread T1 would have updated the metadata table with a new file if tserver had not died

This change acquires a lock so that the call to markUnusedWALs() will block until the file is added to metadata table.  This lock prevents zookeeper updates from happening before metadata table updates.

This bug only affects Accumulo 1.8.0 and later.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
@@ -2458,11 +2458,22 @@ public class Tablet implements TabletCommitter {
 
   private Set<DfsLogger> currentLogs = new HashSet<>();
 
-  public synchronized void removeInUseLogs(Set<DfsLogger> candidates) {
-    // remove logs related to minor compacting data
-    candidates.removeAll(otherLogs);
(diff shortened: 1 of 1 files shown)


keith-turner:
I am seeing some WAL and Replication ITs timeout, so the locking introduced in this PR may be causing a problem.

keith-turner:
WriteAheadLogIT timed out once on Jenkins, it was automatically run again and it passed.    I ran WriteAheadLogIT 6 or 7 times locally without issue.

keith-turner:
This PR caused deadlock, which #545 attempted to fix.  However #545 also caused deadlock, which #559 ultimately fixed.  So #559 is really the fix for the problem described in this PR.

PR #548, opened 2018-07-03
Added additional client properties for Scanner and BatchScanner

mikewalch:


core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -61,6 +61,14 @@
       "Change the" + " durability for the BatchWriter session. To use the table's durability"
           + " setting. use \"default\" which is the table's durability setting."),
 
+  // Scanner
+  SCANNER_BATCH_SIZE("scanner.batch.size", "1000",
```
keith-turner: This should also be used in the factory methods for creating a scanner and batch scanner, setting the default for those.

mikewalch: It is used in the factory method for scanner.

keith-turner: ok I see.  When I looked at the diffs I thought it was setting it only for the batch deleter

PR #556, opened 2018-07-11
Updated scan executor prop docs

keith-turner:
Also improved the document generation to use new custom Liquid tags and
fixed some HTML escaping issues.


core/src/main/java/org/apache/accumulo/core/conf/ConfigurationDocGen.java
```diff
@@ -117,7 +117,13 @@ void property(Property prop) {
         // deal with multi-line values, skip strikethrough of value
         defaultValue = strike("**default value:** ", depr) + "\n```\n" + defaultValue + "\n```\n";
       } else {
-        defaultValue = strike("**default value:** " + "`" + defaultValue + "`", depr);
+        if (prop.getType() == PropertyType.CLASSNAME
+            && defaultValue.startsWith("org.apache.accumulo.core")) {
```
mikewalch: I think this should be `startsWith("org.apache.accumulo")`.  We can update the jlink tag to handle different package prefeixes.

PR #562, opened 2018-07-16
Fix array out of bounds in FastFormat

keith-turner:


keith-turner:
> Is this still faster than String.format for padding?

I suspect it is, unless something has optimized away parsing and executing the format string each time.

Issue #555, opened 2018-07-10
Add ability for Scanner to request an executor


keith-turner:
Building on the work done in #549 it would be nice if Scanners could request an executor.  Scanner could have a method like `requestScanExecutor(String name)`.  The requested executor would be made available (via the ScanInfo interface) to the per table ScanDispatcher on the server side, which may or may not honor this request.  The built in SimpleScanDispatcher could have an option to enable honoring these request, but ignore them by default.

keith-turner:
This could be a bit more general, with a scanner having a method like the following :

```
/**
 * Provides information to the server side scan dispatcher and prioritizer that may be used when making decisions.  However those server side classes may ignore these hints.
 */
void setHints(Map<String, String> hints){}
```

PR #571, opened 2018-07-27
Add scanner execution hints #555

keith-turner:


keith-turner:
Updated performance test to use hints apache/accumulo-testing#31

server/monitor/src/main/java/org/apache/accumulo/monitor/rest/tservers/TabletServerResource.java
```diff
@@ -233,7 +233,7 @@ public TabletServerSummary getTserverDetails(
   }
 
   private static final int concurrentScans = Monitor.getContext().getConfiguration()
-      .getCount(Property.TSERV_READ_AHEAD_MAXCONCURRENT);
+      .getCount(Property.TSERV_SCAN_EXECUTORS_DEFAULT_THREADS);
```
milleruntime: I see you are updating this for your changes but it makes me wonder if this value on the Monitor is deceiving?

core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
```diff
@@ -286,4 +292,18 @@
    * @since 1.8.0
    */
   String getClassLoaderContext();
+
+  /**
+   * Set hints for the configured {@link ScanPrioritizer} and {@link ScanDispatcher}. These hints
+   * are available on the server side via {@link ScanInfo#getExecutionHints()} Depending on the
+   * configuration, these hints may be ignored. These hints should never impacts what data is
```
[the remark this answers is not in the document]
keith-turner: Should be able to say that as long as iterators are never given access to the hints on the server side.

server/monitor/src/main/java/org/apache/accumulo/monitor/rest/tservers/TabletServerResource.java
keith-turner: I fixed that.

PR #575, opened 2018-07-30
Add splits to table at time of table creation #573

### LINKED ISSUE #573: Add splits to table at table creation.
Add capability to add table splits at table creation. Recent changes now allow iterator and locality groups to be created at table creation. Do the same with splits. See comment from [ACCUMULO-4806 ](https://issues.apache.org/jira/browse/ACCUMULO-4808?filter=-1 )for an explanation of the motivation for this request.

Linked to JIRA ticket [ACCUMULO-4808](https://issues.apache.org/jira/browse/ACCUMULO-4808?filter=-1)

jmark99:
This issue updates Accumulo to allow for tables to create splits at the  time of table creation. A new step is added to the createtable FATE operations that handle this task if it is activated.

The most significant modification is that the splits are written to the metadata table after the basic table information is created but before the table is brought online. This is handled by the CreateInitialSplits class within the FATE operation.

Closes #573


core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
```diff
@@ -234,8 +234,19 @@ public void create(String tableName, NewTableConfiguration ntc)
...
+    args.add(ByteBuffer.wrap(tableName.getBytes(UTF_8)));
+    args.add(ByteBuffer.wrap(ntc.getTimeType().name().getBytes(UTF_8)));
+    // Check for possible initial splits to be added at table creation
+    // Always send number of initial splits to be created, even if zero. If greater than zero,
+    // add the splits to the argument List which will be used by the FATE operations.
+    int numSplits = ntc.getSplits().size();
+    args.add(ByteBuffer.wrap(String.valueOf(numSplits).getBytes(UTF_8)));
+    if (numSplits > 0) {
+      for (Text t : ntc.getSplits()) {
+        args.add(ByteBuffer.wrap(t.toString().getBytes(UTF_8)));
```
keith-turner: Splits may be binary data.  Converting Text to  a String and then to a ByteBuffer could scramble the binary data (because it encodes and then decodes utf-8).  I would recommend doing `args.add(TextUtil. getByteBuffer(t))`.  Using TextUtil properly handles the edge case where Text has a byte array thats longer than its length.

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
```diff
@@ -199,6 +225,21 @@ public NewTableConfiguration setLocalityGroups(Map<String,Set<Text>> groups) {
     return this;
   }
 
+  /**
+   * Create a new table wkth pre-configured splits from the provided imput collection.
```
keith-turner: Spelling : `wkth` and `imput`

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
```diff
@@ -199,6 +225,21 @@ public NewTableConfiguration setLocalityGroups(Map<String,Set<Text>> groups) {
...
+   *
+   * @param splits
+   *          A SortedSet of String values to be used as split points in a newly created table.
+   * @return this
```
keith-turner: Should add `@since 2.0.0` javadoc tag.

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
```diff
@@ -103,6 +108,16 @@ public NewTableConfiguration withoutDefaultIterators() {
     return this;
   }
 
+  /**
+   * Create the new table in an offline state.
+   *
+   * @return this
+   */
```
keith-turner: Should add `@since 2.0.0` javadoc tag.

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
```diff
@@ -199,6 +225,21 @@ public NewTableConfiguration setLocalityGroups(Map<String,Set<Text>> groups) {
...
+   *
+   * @param splits
+   *          A SortedSet of String values to be used as split points in a newly created table.
+   * @return this
+   */
+  public NewTableConfiguration withSplits(final SortedSet<Text> splits) {
+    checkArgument(splits != null, "splits set is null");
+    checkArgument(splits.isEmpty() != true, "splits set is empty");
+    this.createInitialSplits = true;
+    this.splitProps = splits;
```
keith-turner: Personally I would do `splitProps = ImmuatbleSortedSet.copyOf(splits)` here.  This makes this API's behavior more predictable.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
```diff
@@ -664,4 +670,41 @@ private String validateNamespaceArgument(ByteBuffer namespaceArg, TableOperation
...
+   */
+  private String createSplitFile(final long opid, final List<ByteBuffer> arguments,
+      final int splitCount, final int splitOffset) {
+    String opidStr = String.format("%016x", opid);
+    String splitPath = getSplitPath("/tmp/splits-" + opidStr);
+    try (FSDataOutputStream stream = master.getOutputStream(splitPath)) {
+      writeSplitsToFileSystem(stream, arguments, splitCount, splitOffset);
+    } catch (IOException e) {
+      splitPath = null;
+      e.printStackTrace();
```
keith-turner: Seems like this should log a message (with filename, error, and opid) and maybe rethrow the exception which should cause the create table op to fail.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateTable.java
```diff
@@ -58,6 +68,10 @@ public long isReady(long tid, Master environment) throws Exception {
     Utils.idLock.lock();
     try {
       tableInfo.tableId = Utils.getNextId(tableInfo.tableName, master.getInstance(), Table.ID::of);
+      if (tableInfo.props.containsKey(Property.TABLE_OFFLINE_OPTS + "create.initial.splits")
+          && this.splitFile != null) {
+        storeSplitFileNameInZooKeeper();
```
keith-turner: I don't think this needs to be stored in ZK explicitly.  The way FATE works, each step of a FATE op is serialized to ZK.  Therefore if you store any info you want to keep around in the TableInfo object, then the TableInfo object will be serialized in subsequent steps.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
```diff
@@ -0,0 +1,179 @@
...
+    return 0;
+  }
+
+  @Override
+  public Repo<Master> call(long tid, Master environment) throws Exception {
+    String splitFile = getSplitFileName();
+
+    // Read table entry and store values for later use
+    SplitEntry splitEntry = new SplitEntry();
+    populateSplitEntry(environment, splitEntry);
```
keith-turner: I would recommend dropping this class and rolling the functionality it contains into `PopulateMetadata`.  It seems like the function `populateSplitEntry` is reading what `PopulateMetadata` wrote.   `PopulateMetadata` used to write a single tablet to the metadata table.   It could be changed to write one or more tablets.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    return new FinishCreateTable(tableInfo);
+  }
+
+  private void writeLastRowToMetadataTable(SplitEntry splitEntry, BatchWriter bw)
+      throws MutationsRejectedException {
+    Text lastRow = new Text(tableInfo.tableId.toString() + "<");
+    Mutation mut = new Mutation(lastRow);
+    mut.put(new Text("srv"), new Text("dir"), splitEntry.dirValue);
```
keith-turner: Really need to use constants here.   In PopulateMetadata it calls `MetadataTableUtil.addTablet()`.  If you dig into the source of `addTablet()` you can see the constants its uses.   Some of the constants are quirky, let me know if you have any questions.  I would be happy to provide examples if needed.

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
```diff
@@ -134,6 +149,13 @@ public NewTableConfiguration setProperties(Map<String,String> props) {
     if (limitVersion)
       propertyMap.putAll(IteratorUtil.generateInitialTableProperties(limitVersion));
 
+    if (createOffline)
+      propertyMap.put(Property.TABLE_OFFLINE_OPTS + "create.offline", "true");
+
+    if (createInitialSplits) {
+      propertyMap.put(Property.TABLE_OFFLINE_OPTS + "create.initial.splits", "true");
+    }
```
milleruntime: Along with what Keith said about the Properties, these map keys should probably just be internal constants. since it looks like these actions are just performed through the API and not something you would set in the config.

core/src/test/java/org/apache/accumulo/core/client/admin/NewTableConfigurationTest.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.TreeSet;
+
+import org.apache.accumulo.core.conf.Property;
+import org.apache.hadoop.io.Text;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class NewTableConfigurationTest {
```
milleruntime: Nice test.  There is also NewTableConfigurationIT you could add tests to. 

Since this is new, you could create a follow on ticket to have unit tests for the other parameters in NewTableConfiguration tested here as well.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
```diff
@@ -664,4 +670,41 @@ private String validateNamespaceArgument(ByteBuffer namespaceArg, TableOperation
           TableOperationExceptionType.INVALID_NAME, why);
     }
   }
+
+  /**
+   * Create a file on the file system to hold the splits to be created at table creation.
+   */
+  private String createSplitFile(final long opid, final List<ByteBuffer> arguments,
```
milleruntime: This is interesting and seems like a first for fate transactions.  I am just curious of your thoughts for storing the splits in a file in tmp?  Wouldn't this be a problem with recovery if the server is restarted in the middle of the fate transaction and the file is gone?

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
```diff
@@ -0,0 +1,179 @@
...
+              mut.put(new Text("~tab"), new Text("~pr"), splitEntry.firstRow);
+              splitEntry.first = false;
+            } else {
+              mut.put(new Text("~tab"), new Text("~pr"),
+                  KeyExtent.encodePrevEndRow(new Text(splitEntry.lastSplit)));
+            }
+            bw.addMutation(mut);
+            splitEntry.lastSplit = split;
+          } catch (Exception e) {
+            e.printStackTrace();
```
milleruntime: Should be a logger to log here.

jmark99:
@keith-turner, @milleruntime, Thanks for the initial pass over the PR. It is an involved change and also my first run at modifying a FATE operation. I will begin addressing and modifying the code based on your comments throughout the next several days. I'm sure I will have some questions along the way as well. Please let me know of any additional suggestions or concerns as they come to you.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -864,10 +864,11 @@
       "Prefix for configuring summarizers for a table. Using this prefix"
           + " multiple summarizers can be configured with options for each one. Each"
           + " summarizer configured should have a unique id, this id can be anything."
-          + " To add a summarizer set "
-          + "`table.summarizer.<unique id>=<summarizer class name>.` If the summarizer has options"
-          + ", then for each option set" + " `table.summarizer.<unique id>.opt.<key>=<value>`."),
-
+          + " To add a summarizer set table.summarizer.<unique id>=<summarizer class"
+          + " name>. If the summarizer has options, then for each option set"
+          + " table.summarizer.<unique id>.opt.<key>=<value>."),
```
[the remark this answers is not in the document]
jmark99: This was not an intentional change. Will revert.

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
jmark99: Corrected!

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -864,10 +864,11 @@
...
+  TABLE_OFFLINE_OPTS("table.offline.", null, PropertyType.PREFIX,
```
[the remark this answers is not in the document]
jmark99: I debated about adding the 'table.offline option'. It is not required for this ticket but was more of a side effect. In some of our past discussions we alluded to adding the capability of adding some operations that could work on offline tables. It was a simple change to allow a user to create a table that would be created in offline mode. At the moment there is no real activity that could be done with that initial offline table. This was intended to be there for that that eventuality. In its current state a user could choose to create a table in offline mode although there's nothing they could do with the table other than place it in online mode when needed.

It's easy enough to remove the option altogether and re-add it later if the need arises. Let me know your thoughts.

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
jmark99: added

core/src/main/java/org/apache/accumulo/core/client/admin/NewTableConfiguration.java
jmark99: added

core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
jmark99: Updated.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
jmark99: I went with the writing of splits to a temp file on the filesystem  after some discussions with @keith-turner. Given the size of some of the split sets it wouldn't be feasible to pass around the data within the FATE operations. Keith suggested writing them to a temporary file then then reading them in during the FATE operation itself. You are correct that there could be an issue if the server went down before the file was written, I think we decided that the odds of that happening would be low (hopefully).  I don't have a better solution at the moment. I'm open to alternative suggestions.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
jmark99: Added a log message and rethrow. Table Op will fail with ThriftTableOperationException if write fails.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
keith-turner: @milleruntime were you thinking this is `/tmp` on the local FS?  I think this may `/tmp` in an accumulo hdfs dir.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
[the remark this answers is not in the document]
keith-turner: > At the moment there is no real activity that could be done with that initial offline table.

Actually, can bulk import to offline tables at the moment. So with this PR a user should be able to create a table with splits offline, bulk import to it, and then bring it online.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
keith-turner: > You are correct that there could be an issue if the server went down before the file was written

Not sure there is a problem with the server going down.   The file is created before the FATE op is seeded.  If the servers goes down while its trying to setup the FATE op, then the client will retry RPC to seed fate op and in the process resend the splits.  Need to ensure the code that creates the split file is idempotent.  If the code is not idempotent, that could cause problems.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateTable.java
jmark99: Updated TableInfo to contain a splitInfo entry. Enabled simplication of a couple of classes by negating the need for some utility methods that were needed when writing the info to ZK.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
jmark99: Updated to utilize constants and helper utils from the MetadataTableUtil class. @keith-turner let me know if I used any of these incorrectly. The tests still passed so I'm assuming the updates are okay.

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
```diff
@@ -155,8 +155,8 @@ public void executeFateOperation(TInfo tinfo, TCredentials c, long opid, FateOpe
             splitFile = createSplitFile(opid, arguments, splitCount, SPLIT_OFFSET);
           } catch (IOException e) {
```
keith-turner: It seems like the exception e is just lost.   If not done elsewhere, would be good to log something about e for debugging purposes.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
```diff
@@ -99,10 +95,10 @@ private void writeLastRowToMetadataTable(SplitEntry splitEntry, BatchWriter bw)
       throws MutationsRejectedException {
     Text lastRow = new Text(tableInfo.tableId.toString() + "<");
     Mutation mut = new Mutation(lastRow);
-    mut.put(new Text("srv"), new Text("dir"), splitEntry.dirValue);
-    mut.put(new Text("srv"), new Text("lock"), splitEntry.lockValue);
-    mut.put(new Text("srv"), new Text("time"), splitEntry.timeValue);
-    mut.put(new Text("~tab"), new Text("~pr"),
+    MetadataSchema.TabletsSection.ServerColumnFamily.DIRECTORY_COLUMN.put(mut, splitEntry.dirValue);
```
keith-turner: Could do static imports of these to make code shorter.  In eclipse may be able to do this with Ctrl+Shift+M

server/master/src/main/java/org/apache/accumulo/master/tableOps/PopulateMetadata.java
```diff
@@ -42,8 +43,11 @@ public long isReady(long tid, Master environment) throws Exception {
     MetadataTableUtil.addTablet(extent, tableInfo.dir, environment, tableInfo.timeType,
         environment.getMasterLock());
 
-    return new FinishCreateTable(tableInfo);
-
+    if (tableInfo.props.containsKey(Property.TABLE_OFFLINE_OPTS + "create.initial.splits")) {
```
keith-turner: I cloned this branch to play around with moving adding splits into this class.   I got as far as writing the following function.  I think this function would work for 0 or more splits.  One thing I realized while attempting this exercise is the current code is using the same dir for each tablet which is not ideal (it may work).  So `dirFunction` below represents this TODO.

```java
  private void writeSplitsToMetadataTable(Table.ID tableId, SortedMap<Text> splits, Function<Text,Value> dirFunction, char timeType,
      ZooLock lock, BatchWriter bw) {
    Text prevSplit = null;
    for(Text split : Iterables.concat(splits, Collections.singleton(null))) {
      Mutation mut = new KeyExtent(tableId, split, prevSplit).getPrevRowUpdateMutation();
      ServerColumnFamily.DIRECTORY_COLUMN.put(mut, dirFunction.apply(split));
      ServerColumnFamily.TIME_COLUMN.put(mut, new Value(timeType+"0"));
      MetadataTableUtil.putLockID(lock, mut);
      prevSplit = split;

      bw.addMutation(mut);
    }
  }
```

One possible way to handle this is to do the following.

 * In the ChooseDir FATE step, read in the splits and choose a dir for each split.  Write out a new file with the dirs.
 * In the CreateDir FATE step, read the file with dirs and create them.
 * In the PopulateMetadata FATE step, read the file with dirs and use those for creating rows in metadata table.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
```diff
@@ -0,0 +1,170 @@
...
+    bw.addMutation(mut);
+  }
+
+  private void writeSplitsToMetadataTable(Master environment, String splitFile,
+      SplitEntry splitEntry, BatchWriter bw) throws IOException {
+    String tableId = tableInfo.tableId.toString();
+    try (FSDataInputStream splitStream = environment.getInputStream(splitFile)) {
+      try (BufferedReader br = new BufferedReader(new InputStreamReader(splitStream))) {
+        String split;
+        while ((split = br.readLine()) != null) {
```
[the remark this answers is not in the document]
jmark99: Verified that indeed does fail with binary data. Will refactor to handle binary correctly.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
jmark99: done

server/master/src/main/java/org/apache/accumulo/master/FateServiceHandler.java
jmark99: Updated to catch and throw exception if method fails.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateInitialSplits.java
jmark99: Discussed with @keith-turner and decided that initially will let the method throw an exception if it occurs. The FATE server should log the event. After initial release will examine behaviour of DFS and determine if we need to add some type of retry in the event of failure.

core/src/main/java/org/apache/accumulo/core/client/admin/TableCreationMode.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.accumulo.core.client.admin;
+
+public enum TableCreationMode {
```
keith-turner: need since tag

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateDir.java
```diff
@@ -47,5 +58,23 @@ public void undo(long tid, Master master) throws Exception {
...
+        fs.deleteRecursively(new Path(dirname.toString()));
+      }
+    }
+  }
+
+  private void createTabletDirectories(VolumeManager fs, SortedSet<Text> dirInfo)
+      throws IOException {
+
+    for (Text info : dirInfo) {
+      if (fs.exists(new Path(info.toString())))
```
keith-turner: If the Repo fails and is later run again, would this cause it to fail?

core/src/main/java/org/apache/accumulo/core/client/admin/TableCreationMode.java
jmark99: added

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateDir.java
jmark99: Thanks for catching that. There is no need to fail if the directory already exists upon retry.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateDir.java
```diff
@@ -39,6 +43,13 @@ public long isReady(long tid, Master environment) throws Exception {
   public Repo<Master> call(long tid, Master master) throws Exception {
     VolumeManager fs = master.getFileSystem();
     fs.mkdirs(new Path(tableInfo.dir));
```
keith-turner: I noticed the new code checks the boolean returned from mkdir, however the existing code does not.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateDir.java
```diff
@@ -47,5 +58,21 @@ public void undo(long tid, Master master) throws Exception {
...
+        fs.deleteRecursively(new Path(dirname.toString()));
+      }
+    }
+  }
+
+  private void createTabletDirectories(VolumeManager fs, SortedSet<Text> dirInfo)
+      throws IOException {
+
+    for (Text dir : dirInfo) {
+      if (!fs.mkdirs(new Path(dir.toString())))
```
keith-turner: I wish HDFS had async or batch operations.  Could possibly use a thread pool here to speed up the case of many dirs. I think the HDFS client will batch request from multiple threads into a single RPC to namenode. This could be a follow on issue. Since multiple places in Accumulo code need to create or delete lots of files, could build these high level operations into volume manager.

server/master/src/main/java/org/apache/accumulo/master/tableOps/PopulateMetadata.java
```diff
@@ -52,4 +95,16 @@ public void undo(long tid, Master environment) throws Exception {
         environment.getMasterLock());
   }
 
+  /**
+   * Create a map containing an association between each split directory and a split value.
+   */
+  private Map<Text,Text> createSplitDirectoryMap(SortedSet<Text> splits, SortedSet<Text> dirs) {
```
keith-turner: Could add `Preconditions.checkArgument(splits.size() == dirs.size());`

server/master/src/main/java/org/apache/accumulo/master/tableOps/TableInfo.java
```diff
@@ -32,6 +33,14 @@
...
+  TableCreationMode creationMode;
+
+  // Track information related to initial split creation
+  int initialSplitSize;
+  String splitFile;
+  String splitDirsFile;
+
   public Map<String,String> props;
 
   public String dir = null;
```
keith-turner: Renaming this to defaultTabletDir would have helped me understand some of the code a bit faster.

test/src/main/java/org/apache/accumulo/test/functional/CreateInitialSplitsIT.java
```diff
@@ -0,0 +1,218 @@
...
+  public void testCreateInitialSplitsWithEncodedSplits() throws TableExistsException,
+      AccumuloSecurityException, AccumuloException, TableNotFoundException {
+    tableName = getUniqueNames(1)[0];
+    SortedSet<Text> expectedSplits = generateNonBinarySplits(3000, 32, true);
+    NewTableConfiguration ntc = new NewTableConfiguration().withSplits(expectedSplits);
+    assertFalse(connector.tableOperations().exists(tableName));
+    connector.tableOperations().create(tableName, ntc);
+    assertTrue(connector.tableOperations().exists(tableName));
+    Collection<Text> createdSplits = connector.tableOperations().listSplits(tableName);
+    assertEquals(expectedSplits, new TreeSet<>(createdSplits));
```
keith-turner: It would also be nice to do the following to ensure a table created in this way does not have any quirks.

 1. Write some data to multiple tablets
 2. Verify data
 3. Compact table
 4. Verify data
 5. Delete table.

server/master/src/main/java/org/apache/accumulo/master/tableOps/CreateDir.java
keith-turner: I created #631

core/src/main/java/org/apache/accumulo/core/client/admin/TableCreationMode.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+package org.apache.accumulo.core.client.admin;
+
+/**
+ * Creation mode for table creation.
+ *
+ * @since 2.0.0
+ */
+public enum TableCreationMode {
```
keith-turner: Another possible name for this is `InitialTableState`

keith-turner:
@jmark99 I am finished reviewing this.  These changes look great.  The only major outstanding concern I have is making the test use the tables create in these new ways a bit to ensure they are functional.

keith-turner:
@jmark99  is there a test the creates an offline table?

jmark99:
@keith-turner , there is a test for creating offline tables in ShellServerIT

jmark99:
@keith-turner, thanks for the review. I've changed a couple of the variable/enum names based upon your comments. I'll work to create tests to verify functionality as outlined above.

keith-turner:
Ah ok, the test in ShellServerIT for offline looks good.

jmark99:
@keith-turner, I added an additional IT test to CreateInitialSplitsIT to address the steps you outlined above. If it looks ok do you think we are ready to close this PR?

milleruntime:
@jmark99 It looks like your last merge might have missed some changes... Travis CI is failing: https://travis-ci.org/apache/accumulo/jobs/425445192

jmark99:
Changes were merged into master closing #573

PR #581, opened 2018-08-02
Remove partially functional archive feature

keith-turner:


milleruntime:
I was looking at this code the other day and wondered what this "archive" stuff was all about.  Was there a use case for archiving at one time? Does the latest WAL fixes make this irrelevant?

keith-turner:
The archive stuff was created a long time ago to assist in debugging data loss problem observed when running continuous ingest + agitation.   For a  long time there were no data loss problems until recently.   I looked to use this feature recently and noticed it was not properly working.  Now HDFS has trash and this feature is probably no longer needed.  So I think its better to remove it rather than fix it.

Issue #582, opened 2018-08-07
Construct ServerContext once and pass it through codebase


mikewalch:
It would be better to pass `ServerContext`

PR #585, opened 2018-08-08
ServerContext is now a singleton class

mikewalch:
* ServerContext is only created using getInstance methods
* Increase passing of ServerContext to reduce getInstance calls

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
@@ -280,7 +280,6 @@ public class MetadataConstraints implements Constraint {
             violations = addViolation(violations, 7);
           }
         }
-
       }
     }
 
@@ -297,7 +296,7 @@ public class MetadataConstraints implements Constraint {
   }
 
   protected Arbitrator getArbitrator() {
-    return new ZooArbitrator();
+    return new ZooArbitrator(ServerContext.getInstance());
   }
 
   @Override
--- a/server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
@@ -57,8 +57,7 @@ public class TableManager {
       .synchronizedMap(new HashMap<>());
   private static final byte[] ZERO_BYTE = {'0'};
 
-  private static TableManager tableManager = null;
-
+  private final ServerContext context;
   private final String zkRoot;
   private final String instanceID;
   private ZooCache zooStateCache;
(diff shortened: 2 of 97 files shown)


mikewalch:
@ctubbsii, that is the ultimate goal. There are just parts of the code where it's not easy to remove ServerContext.  This is a good intermediate step.

server/base/src/main/java/org/apache/accumulo/server/tables/TableManager.java
```diff
@@ -99,14 +98,8 @@ public static void prepareNewTableState(String instanceId, Table.ID tableId,
         existsPolicy);
   }
 
-  public synchronized static TableManager getInstance() {
```
keith-turner: Nice to see this go away.

server/base/src/main/java/org/apache/accumulo/server/tablets/UniqueNameAllocator.java
```diff
@@ -71,14 +73,4 @@ public synchronized String getNextName() {
     return new String(FastFormat.toZeroPaddedString(next++, 7, Character.MAX_RADIX, new byte[0]),
         UTF_8);
   }
-
-  private static UniqueNameAllocator instance = null;
-
-  public static synchronized UniqueNameAllocator getInstance() {
```
keith-turner: Another nice removal.

server/base/src/main/java/org/apache/accumulo/server/util/MetadataTableUtil.java
```diff
@@ -114,7 +113,7 @@
 
   private MetadataTableUtil() {}
 
-  public synchronized static Writer getMetadataTable(ClientContext context) {
+  public synchronized static Writer getMetadataTable(ServerContext context) {
```
keith-turner: I like how you changed all of the methods in this class to accept server context.

server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
```diff
@@ -297,7 +296,7 @@ private static boolean isValidColumn(ColumnUpdate cu) {
   }
 
   protected Arbitrator getArbitrator() {
-    return new ZooArbitrator();
+    return new ZooArbitrator(ServerContext.getInstance());
```
keith-turner: I think it would be good to open a follow on issue to explore ways to remove this.  There was a lot to look at so I may have missed some, but this is the only call I noticed to ServerContext.getInstance() that was not in a main method.

mikewalch: I am not going to close #582 until all calls to getInstance() outside of main methods are removed and calls to getInstance() are converted to constructors.

PR #588, opened 2018-08-08
ACCUMULO-4629 add exact deletes

keith-turner:


keith-turner:
> Is there a write-up of what an "exact delete" is and what the overall design of this implementation is?

There is now : apache/accumulo-website#102

There is also what was on the JIRA issue.

mikewalch:
I really like exact deletes. I think it would be a great feature to add.  It just seems more useful if it was done on a per-mutation basis instead of changing the behavior of the table.  It would be nice if users could choose between range and exact deletes.  

If you are looking to improve performance by minimizing seeks for deletes, what about adding the option to turn off deletes for a table at creation time?

keith-turner:
@mikewalch and @ctubbsii I agree with your assessments.  I opened #598 based on your comments.

PR #590, opened 2018-08-09
ACCUMULO-4746 Fluent API for Mutation

mikewalch:
Replaces #347

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/data/Mutation.java
+++ b/core/src/main/java/org/apache/accumulo/core/data/Mutation.java
@@ -289,24 +289,24 @@ public class Mutation implements Writable {
     return row;
   }
 
-  private void put(byte b[]) {
-    put(b, b.length);
+  private void fill(byte b[]) {
+    fill(b, b.length);
   }
 
-  private void put(byte b[], int length) {
+  private void fill(byte b[], int length) {
     buffer.writeVLong(length);
     buffer.add(b, 0, length);
   }
 
-  private void put(boolean b) {
+  private void fill(boolean b) {
     buffer.add(b);
   }
 
-  private void put(int i) {
+  private void fill(int i) {
     buffer.writeVLong(i);
   }
 
-  private void put(long l) {
+  private void fill(long l) {
     buffer.writeVLong(l);
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -853,6 +853,475 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier,
...
+  /**
+   * Fluent API for putting or deleting to a Mutation that makes it easy use different types
+   * (i.e byte[], CharSequence, etc) when specifying the family, qualifier, value, etc.
+   *
+   * <p>
+   * Methods are optional but must follow this order: family, qualifier, visibility, timestamp.
+   *
+   * <p>
+   * The put and delete methods end the chain and finalize the Mutation by filling Mutation's
+   * UnsynchronizedBuffer.
```
keith-turner: Not sure how helpful it is to mention UnsynchronizedBuffer here since its an internal impl detail.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -853,6 +853,475 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier,
...
+   *
+   * @return a new FamilyOptions object, advancing the method chain
```
keith-turner: Changing advancing to starting or beginning makes more sense to me.  But I am also ok with current wording.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -853,6 +853,475 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier,
...
+    int columnVisibilityLength;
+
+    boolean hasTs = false;
+    long timestamp;
+
+    private Options() {}
+
+    // methods for changing the column family of a Mutation
+    /**
+     * Modifies the column family of a mutation.
```
keith-turner: Could be : `Sets the family to modify.`  I am ok with the current wording, thought this suggestions was slight more accurate.

core/src/main/java/org/apache/accumulo/core/data/Mutation.java
```diff
@@ -853,6 +853,475 @@ public void putDelete(byte[] columnFamily, byte[] columnQualifier,
...
+    TimestampOptions visibility(ColumnVisibility colVis);
+
+    TimestampOptions visibility(Text colVis);
+  }
+
+  /**
+   * Provides methods for setting the timestamp of a Mutation. The user must provide the timestamp
+   * as a long.
+   *
+   * All TimestampOptions methods return an instance derived from the MutationOptions interface,
```
keith-turner: If multiple paragraphs are desired in rendered javadoc then must have `<p>`.

PR #594, opened 2018-08-10
 #582 - Pass ServerContext to MetadataConstraints

mikewalch:
* Created SystemConstraint and made MetadataConstraints
  extend it
* Passed ServerContext to SystemEnvironment

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 
 import org.apache.accumulo.core.Constants;
 import org.apache.accumulo.core.constraints.Constraint;
@@ -51,7 +52,6 @@ import org.slf4j.LoggerFactory;
 
 public class MetadataConstraints implements Constraint {
 
-  private ServerContext context = null;
   private ZooCache zooCache = null;
   private String zooRoot = null;
 
@@ -108,7 +108,7 @@ public class MetadataConstraints implements Constraint {
 
   @Override
   public List<Short> check(Environment env, Mutation mutation) {
-    context = ((SystemEnvironment)env).getServerContext();
+    final ServerContext context = ((SystemEnvironment)env).getServerContext();
 
     ArrayList<Short> violations = null;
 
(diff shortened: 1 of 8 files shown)


server/base/src/main/java/org/apache/accumulo/server/constraints/SystemConstraint.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.accumulo.core.data.Mutation;
+import org.apache.accumulo.server.ServerContext;
+
+public interface SystemConstraint extends Constraint {
+
+  interface SystemEnvironment extends Environment {
+    ServerContext getServerContext();
+  }
+
+  default List<Short> check(SystemEnvironment env, Mutation mutation) {
```
keith-turner: I don't see where this is called.

server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
```diff
@@ -105,6 +105,12 @@ private static boolean isValidColumn(ColumnUpdate cu) {
     return lst;
   }
 
+  @Override
+  public List<Short> check(SystemEnvironment env, Mutation mutation) {
+    context = env.getServerContext();
```
keith-turner: Could possibly not have a SystemConstraint and only have a SystemEnv.  This class could just cast Env to SystemEnv.

server/base/src/main/java/org/apache/accumulo/server/constraints/SystemConstraint.java
mikewalch: Fixed in latest commit

server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
mikewalch: Fixed with the latest commit

server/base/src/main/java/org/apache/accumulo/server/constraints/MetadataConstraints.java
```diff
@@ -107,6 +108,7 @@ private static boolean isValidColumn(ColumnUpdate cu) {
 
   @Override
   public List<Short> check(Environment env, Mutation mutation) {
+    context = ((SystemEnvironment)env).getServerContext();
```
keith-turner: I think it would be a little cleaner to pass the context to `getArbitrator()` and not have it as an instance var.  The test that overrides  `getArbitrator()` could just ignore the context.

PR #609, opened 2018-08-21
Pass ServerContext to RFile. Closes #584

### LINKED ISSUE #584: Provide ServerContext to RFile
Internal refactor to connect the plumbing of RFile with the top level ServerContext object.  This would especially help with the new CryptoService, allowing elimination of the static objects in CryptoServiceFactory

milleruntime:
* This allows init of crypto in tserver, eliminating static instance in
CryptoServiceFactory

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/rfile/RFileScanner.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/rfile/RFileScanner.java
@@ -218,7 +218,7 @@ class RFileScanner extends ScannerOptions implements Scanner {
     if (null == this.dataCache) {
       this.dataCache = new NoopCache();
     }
-    this.cryptoService = CryptoServiceFactory.getConfigured(tableConf);
+    this.cryptoService = CryptoServiceFactory.newInstance(tableConf);
   }
 
   @Override
--- a/core/src/main/java/org/apache/accumulo/core/client/rfile/RFileSummariesRetriever.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/rfile/RFileSummariesRetriever.java
@@ -90,7 +90,7 @@ class RFileSummariesRetriever implements SummaryInputArguments, SummaryFSOptions
       for (RFileSource source : sources) {
         SummaryReader fileSummary = SummaryReader.load(conf, acuconf, source.getInputStream(),
             source.getLength(), summarySelector, factory,
-            CryptoServiceFactory.getConfigured(acuconf));
+            CryptoServiceFactory.newInstance(acuconf));
         SummaryCollection sc = fileSummary
             .getSummaries(Collections.singletonList(new Gatherer.RowRange(startRow, endRow)));
         all.merge(sc, factory);
--- a/core/src/main/java/org/apache/accumulo/core/file/rfile/RFileOperations.java
+++ b/core/src/main/java/org/apache/accumulo/core/file/rfile/RFileOperations.java
@@ -130,7 +130,13 @@ public class RFileOperations extends FileOperations {
 
       outputStream = fs.create(new Path(file), false, bufferSize, (short) rep, block);
     }
-    CryptoService cryptoService = CryptoServiceFactory.getConfigured(acuconf);
+
+    // calls to openWriter from the tserver will already have a crypto service initialized
+    // calls from clients will require a new crypto service
+    CryptoService cryptoService = options.cryptoService;
+    if (cryptoService == null) {
+      cryptoService = CryptoServiceFactory.newInstance(acuconf);
+    }
 
     BCFile.Writer _cbw = new BCFile.Writer(outputStream, options.getRateLimiter(), compression,
         conf, acuconf, cryptoService);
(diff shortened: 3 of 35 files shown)


core/src/main/java/org/apache/accumulo/core/client/rfile/RFileScanner.java
```diff
@@ -218,7 +218,7 @@ public void indexWeightChanged() {}
     if (null == this.dataCache) {
       this.dataCache = new NoopCache();
     }
-    this.cryptoService = CryptoServiceFactory.getConfigured(tableConf);
+    this.cryptoService = CryptoServiceFactory.newInstance(tableConf);
```
mikewalch: Not sure if tableConf can be used here.  Will it have your crypto configuration?

core/src/main/java/org/apache/accumulo/core/client/rfile/RFileSummariesRetriever.java
```diff
@@ -90,7 +90,7 @@ public SummaryOptions endRow(Text endRow) {
       for (RFileSource source : sources) {
         SummaryReader fileSummary = SummaryReader.load(conf, acuconf, source.getInputStream(),
             source.getLength(), summarySelector, factory,
-            CryptoServiceFactory.getConfigured(acuconf));
+            CryptoServiceFactory.newInstance(acuconf));
```
mikewalch: `acuconf` is default configuration. you should pass in your crypto configuration

core/src/main/java/org/apache/accumulo/core/file/rfile/RFileOperations.java
```diff
@@ -130,7 +130,13 @@ protected FileSKVWriter openWriter(FileOptions options) throws IOException {
 
       outputStream = fs.create(new Path(file), false, bufferSize, (short) rep, block);
     }
-    CryptoService cryptoService = CryptoServiceFactory.getConfigured(acuconf);
+
+    // calls to openWriter from the tserver will already have a crypto service initialized
+    // calls from clients will require a new crypto service
+    CryptoService cryptoService = options.cryptoService;
+    if (cryptoService == null) {
+      cryptoService = CryptoServiceFactory.newInstance(acuconf);
```
mikewalch: `acuconf` is table configuration

core/src/main/java/org/apache/accumulo/core/file/rfile/SplitLarge.java
```diff
@@ -60,7 +60,7 @@ public static void main(String[] args) throws Exception {
 
     for (String file : opts.files) {
       AccumuloConfiguration aconf = DefaultConfiguration.getInstance();
-      CryptoService cryptoService = CryptoServiceFactory.getConfigured(aconf);
+      CryptoService cryptoService = CryptoServiceFactory.newInstance(aconf);
```
mikewalch: aconf is default configuration

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -241,4 +256,11 @@ public synchronized UniqueNameAllocator getUniqueNameAllocator() {
     }
     return nameAllocator;
   }
+
+  public synchronized CryptoService getCryptoService() {
```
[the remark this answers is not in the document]
milleruntime: I wasn't sure. The scenario I was thinking was when something tries to get the CryptoService before tserver was started but I don't think making this method sychronized would help with that anyway.  I will remove it.

core/src/main/java/org/apache/accumulo/core/client/rfile/RFileScanner.java
milleruntime: The impl of RFile.withTableProperties() will just take whatever key values are passed and put them in a map.  It is almost as if this method should be renamed to "withProperties" since we allow more than just table properties.  Since it is already released as public API maybe we can just elaborate more in the javadoc.  I will create a follow on issue to investigate this.  

I don't think there is anything I can do for this PR though since we don't want to put the crypto stuff in the public API while its still experimental.

core/src/main/java/org/apache/accumulo/core/client/rfile/RFileSummariesRetriever.java
milleruntime: We are in the same boat here as RFileScanner

core/src/main/java/org/apache/accumulo/core/client/rfile/RFileScanner.java
milleruntime: Created #611

core/src/main/java/org/apache/accumulo/core/client/rfile/RFileSummariesRetriever.java
milleruntime: Nevermind, crypto won't actually work with the summaries API as-is.

milleruntime: Opened #612 to fix this.

PR #613, opened 2018-08-22
Removed all use of SiteConfiguration.getInstance()

mikewalch:
* Removed use of VolumeManagerImpl.get()
* Removed HdfsZooInstance class
* Removed use of ZooReaderWriter.getInstance()


server/base/src/main/java/org/apache/accumulo/server/security/handler/ZKAuthorizor.java
```diff
@@ -51,22 +52,21 @@
 
   private final String ZKUserAuths = "/Authorizations";
 
+  private ServerContext context;
   private String ZKUserPath;
-  private final ZooCache zooCache;
+  private ZooCache zooCache;
 
   public static synchronized Authorizor getInstance() {
```
milleruntime: Another place we could get rid of static state.  Could create a follow on.

server/base/src/main/java/org/apache/accumulo/server/zookeeper/ZooReaderWriter.java
```diff
@@ -20,25 +20,18 @@
...
   }
 
-  public static synchronized ZooReaderWriter getInstance() {
-    if (instance == null) {
-      AccumuloConfiguration conf = SiteConfiguration.getInstance();
-      instance = new ZooReaderWriter(conf.get(Property.INSTANCE_ZK_HOST),
-          (int) conf.getTimeInMillis(Property.INSTANCE_ZK_TIMEOUT),
-          conf.get(Property.INSTANCE_SECRET));
-    }
-    return instance;
```
milleruntime: Nice

PR #617, opened 2018-08-27
Removed duplicate calls and implementation of SecureRandom

PircDef:
Removed duplicate implementations and reduced the number of calls to SecureRandom.getInstance

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoUtils.java
+++ b/core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoUtils.java
@@ -37,11 +37,11 @@ public class CryptoUtils {
 
   private static final Logger log = LoggerFactory.getLogger(CryptoUtils.class);
 
-  public static SecureRandom getSha1SecureRandom() {
-    return getSecureRandom("SHA1PRNG", "SUN");
+  public static SecureRandom newSha1SecureRandom() {
+    return newSecureRandom("SHA1PRNG", "SUN");
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/accumulo/core/security/crypto/impl/AESCryptoService.java
```diff
@@ -67,6 +66,7 @@
   private String encryptingKeyManager = null;
   // Lets just load keks for reading once
   private static HashMap<String,Key> decryptingKeys = new HashMap<>();
+  private static SecureRandom sr = CryptoUtils.getSha1SecureRandom();
```
milleruntime: You could make this created in init() so it won't have to be static.  We are making a push to try and eliminate static state wherever we can.

core/src/main/java/org/apache/accumulo/core/security/crypto/CryptoUtils.java
```diff
@@ -41,23 +41,16 @@ public static SecureRandom getSha1SecureRandom() {
     return getSecureRandom("SHA1PRNG", "SUN");
   }
 
-  public static SecureRandom getSecureRandom(String secureRNG, String secureRNGProvider) {
+  private static SecureRandom getSecureRandom(String secureRNG, String secureRNGProvider) {
```
milleruntime: You could name this method "newSecureRandom" and the other "newSha1" to show that it is creating a new object every time it is called.  This kinda goes along with what I recently did in CryptoServiceFactory.

PR #630, opened 2018-09-05
Switched to using HADOOP_HOME for Hadoop 3.x

mikewalch:


mikewalch:
The related Hadoop issue is https://jira.apache.org/jira/browse/HADOOP-11393

While things were working, I was getting warnings from Hadoop when using HADOOP_PREFIX.

PR #644, opened 2018-09-12
Add experimental props to doc gen

milleruntime:
* Printed the experimental properties in a separate table
* Removed inner classes cluttering up ConfigurationDocGen


milleruntime:
Here is what they look like: https://github.com/apache/accumulo-website/pull/107

PR #648, opened 2018-09-14
Rename new Bulk import API

milleruntime:
Came up with this after reviewing #607

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
@@ -612,7 +612,7 @@ public interface TableOperations {
    * @throws TableNotFoundException
    *           when the table no longer exists
    *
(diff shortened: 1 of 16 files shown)


core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
```diff
@@ -709,7 +711,7 @@ void load()
    *
    * @since 2.0.0
    */
-  default ImportSourceArguments addFilesTo(String tableName) {
+  default ImportDestinationArguments importDirectory(String directory) {
```
keith-turner: Since its a fluent method and that might not be obvious at first, it would be nice to add an example of calling this to the javadoc.

keith-turner: This is a pre-existing issue, so it does not need to hold this up.  I think these changes look good.

PR #654, opened 2018-09-20
AccumuloFileOutputFormatITs need to look on local filesystem for results written into JUnit provided temp dir.

busbey:
Before this change the {mapred,mapreduce}.AccumuloFileOutputFormatITs would work against a minicluster backed by
Hadoop's LocalFileSystem, but fail if one followed the instructions for testing against a stand alone cluster.

The failure is because in the cluster case the FileSystem we get is for the HDFS instance underlying the
Accumulo cluster, which can't find the folder JUnit created for our test.


busbey:
the relevant change is going from `FileSystem.get(conf)` to `FileSystem.getLocal(conf)`; the formatting changes were done by maven.

PR #655, opened 2018-09-20
Remove httpclient dependency.

busbey:
* was present for Hadoop 1, which we don't support anymore
* was present for ThreadSafe note on our maven-plugin mojo, but that annotation is gone in current releases

Tested via running ITs locally and against standalone hadoop clusters

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,7 +130,6 @@
     <hadoop.version>2.6.4</hadoop.version>
     <htrace.hadoop.version>4.1.0-incubating</htrace.hadoop.version>
     <htrace.version>3.1.0-incubating</htrace.version>
-    <httpclient.version>3.1</httpclient.version>
     <it.failIfNoSpecifiedTests>false</it.failIfNoSpecifiedTests>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -189,7 +188,7 @@
       <dependency>
         <groupId>commons-codec</groupId>
         <artifactId>commons-codec</artifactId>
-        <version>1.4</version>
```
keith-turner: Is this change related to removing commons-httpclient in some way?

busbey: oh no. going from commons-codec 1.4 to 1.11 was part of a different set of changes I was doing that I must have missed cleaning up. let me push and update that moves that out.

keith-turner: Having it the same commit doesn't bother me. I was just curious if there was a connection. Could also update the commit message with a reason for bumping, if there is one.

keith-turner:
Fluo uses the Accumulo maven plugin for its ITs.  I ran Fluo's ITs using these changes w/o any issue.

pom.xml
busbey: pushed an update that left the commons-codec change out.

PR #660, opened 2018-09-24
Link to API definition on website

keith-turner:


keith-turner:
This goes with apache/accumulo-website#109

PR #673, opened 2018-10-03
Fixes #561 - Be explicit about aliases in shell

### LINKED ISSUE #561: Be explicit about aliases in shell
`deletetable` is the same command as `droptable`.  The help in shell should be more explicit that they are aliases

Currently:
```
deletetable           -  deletes a table
droptable             -  deletes a table
```

Help should look something like this:
```
deletetable           -  deletes a table (same as droptable)
droptable             -  deletes a table (same as deletetable)
```

cjmctague:
- Explicitly state that deletetable and droptable are the same

Screenshot from running in uno.
screenshot 2018-10-03 10 12 21


mikewalch:
The build failures are sporadic so I will merge this.

PR #674, opened 2018-10-03
Clarify RFile withTableProps javadoc. Fixes #611

### LINKED ISSUE #611: Passing non table props to RFile API
This situation came up in #609:  the RFile API has a method called "withTableProperties" that will take whatever key value properties you pass it, including instance properties.

milleruntime:
Not sure how to word this...

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/rfile/RFile.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/rfile/RFile.java
@@ -31,6 +31,7 @@ import org.apache.accumulo.core.client.summary.Summarizer;
 import org.apache.accumulo.core.client.summary.SummarizerConfiguration;
 import org.apache.accumulo.core.client.summary.Summary;
 import org.apache.accumulo.core.client.summary.Summary.FileStatistics;
+import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.core.data.Key;
 import org.apache.accumulo.core.data.Range;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/accumulo/core/client/rfile/RFile.java
```diff
@@ -171,7 +171,8 @@
 
     /**
      * Construct the {@link Scanner} with iterators specified in a tables properties. Properties for
-     * a table can be obtained by calling {@link TableOperations#getProperties(String)}
+     * a table can be obtained by calling {@link TableOperations#getProperties(String)}. Other
+     * properties are allowed as well in the props parameter.
```
keith-turner: What do you mean by `Other properties are allowed`?

milleruntime: From my comment on #609 ...
> The impl of RFile.withTableProperties() will just take whatever key values are passed and put them in a map. It is almost as if this method should be renamed to "withProperties" since we allow more than just table properties. Since it is already released as public API maybe we can just elaborate more in the javadoc.

keith-turner: It would be nice to expand on what `Other properties` means in the javadoc.  Below is my attempt to explain it :

Any property that impacts file behavior regardless of whether it has the `table.` prefix may be accepted and used .  For example cache and crypto properties could be passed here.

PR #692, opened 2018-10-11
Remove key provider from crypto

milleruntime:
* Removed hard coded SunJCE key provider from AESKeyUtils
* Removed key.provider property from AESCryptoService

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/security/crypto/impl/AESCryptoService.java
+++ b/core/src/main/java/org/apache/accumulo/core/security/crypto/impl/AESCryptoService.java
@@ -88,7 +88,8 @@ public class AESCryptoService implements CryptoService {
       default:
         throw new CryptoException("Unrecognized key manager");
     }
-
+    Objects.requireNonNull(this.encryptingKek,
+        "Encrypting Key Encryption Key was null, init failed");
   }
 
   @Override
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/accumulo/core/security/crypto/impl/AESCryptoService.java
```diff
@@ -72,12 +72,11 @@
 
   @Override
   public void init(Map<String,String> conf) throws CryptoException {
-    String keyLocation = conf.get("instance.crypto.opts.key.location");
-    String keyMgr = conf.get("instance.crypto.opts.key.provider");
+    String keyLocation = conf.get("instance.crypto.opts.key.uri");
+    // get key from URI for now, keyMgr framework could be expanded on in the future
+    String keyMgr = "uri";
     Objects.requireNonNull(keyLocation,
```
PircDef: If we're going to move the option to a properties value, such as "instance.crypto.opts.key.uri", we should probably also move the Objects.requireNonNull() to this.encryptingKek after the switch statement. That way as keyMgr's are added in the future, only the result is checked.

milleruntime: OK I will add an additional check after a call to loadKekFromUri().  I would think an exception would be thrown here but it wouldn't hurt to make sure null isn't returned.

PircDef: The check should be outside the switch (keyMgr) block, so that whichever method is used in the future, init will always make sure the key isn't null.

PR #697, opened 2018-10-11
ACCUMULO-4496 Update user manual and error messages for FATE upgrade …

EdColeman:
…failure.

Updated the user manual and the log message generated when outstanding
FATE transactions cause upgrade to fail.


mikewalch:
@EdColeman, this looks good.  Do you want me to merge it?

PR #726, opened 2018-10-26
log deprecated properties

mschnitzler:
Created `@ReplacedBy` annotation to denote deprecated properties.
Added log messages in `ConfigSanityCheck` for properties annotated with `@Deprecated` and `@ReplacedBy`.

Is this what you were looking for?

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/conf/ReplacedBy.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/ReplacedBy.java
@@ -29,5 +29,5 @@ import java.lang.annotation.RetentionPolicy;
 @Retention(RetentionPolicy.RUNTIME)
 @interface ReplacedBy {
 
-  Property replacedByProperty();
+  Property property();
 }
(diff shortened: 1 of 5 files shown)


mschnitzler:
issue #725

mikewalch:
@mschnitzler,  Thanks for the PR!  Looks great but I ran into 2 minor issues when I tested it:
* The warnings would print even when deprecated property was not set
* The warnings would print multiple times as the validate() method in ConfigSanityCheck is called multiples times (like when a table is created).

I created a pull request to your branch that should fix these issues.  Let me know what you think.  If you merge it, this pull request will be updated with my changes.

mikewalch:
This was merged in by d83837c81a29f253e41bac313ae768963320e71a

Thanks for the pull request!  If you would like to be added as a [contributor](https://accumulo.apache.org/people/), edit [this page](https://github.com/apache/accumulo-website/blob/master/pages/people.md) on GitHub and make a pull request.

core/src/main/java/org/apache/accumulo/core/conf/ReplacedBy.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * An annotation to denote that an {@link AccumuloConfiguration} {@link Property} key is replaced by
+ * a different {@link AccumuloConfiguration} {@link Property} key. This usually means that the
+ * {@link AccumuloConfiguration} {@link Property} key is {@link Deprecated}.
+ */
+@Inherited
+@Retention(RetentionPolicy.RUNTIME)
+@interface ReplacedBy {
```
keith-turner: I really like this annotation.

PR #727, opened 2018-10-26
fixes #587 Wait for tablets assigned to dead servers before deleting table

### LINKED ISSUE #587: Race condition with delete table and tablet server death result in metadata table garbage
The following may be possible.

 1. Tablet server X dies
 1. Master notices that Tablet T1 is assigned to tablet server X
 1. Table containing T1 is deleted 
 1. Master writes out walogs for tablet T1 to metadata table (which results in random junk in metadata table)
 1. Garbage collector and master notice junk in metadata table and decide to go out to lunch

Not 100% sure this can occur, need to investigate further.  If it can happen, it probably only impacts 1.8.0 and later.

keith-turner:
The tableOps.CleanUp.isReady() operation was not waiting on tablets that were
assigned to dead tablet servers.  The master will attempt to write logs and
delete the location for these tablets.  Therefore its best to wait for the
Master to make its metadata update before proceeding to clean up a deleted
table.

### DIFF
--- a/server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
@@ -204,7 +204,8 @@ abstract class TabletGroupWatcher extends Daemon {
           }
           Master.log.debug(store.name() + " location State: " + tls);
           // ignore entries for tables that do not exist in zookeeper
-          if (TableManager.getInstance().getTableState(tls.extent.getTableId()) == null)
+          TableState tableState = TableManager.getInstance().getTableState(tls.extent.getTableId());
+          if (tableState == null || tableState == TableState.DELETING)
             continue;
 
           if (Master.log.isTraceEnabled())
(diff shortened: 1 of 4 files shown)


server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
```diff
@@ -204,7 +204,8 @@ public void run() {
           }
           Master.log.debug(store.name() + " location State: " + tls);
           // ignore entries for tables that do not exist in zookeeper
-          if (TableManager.getInstance().getTableState(tls.extent.getTableId()) == null)
+          TableState tableState = TableManager.getInstance().getTableState(tls.extent.getTableId());
+          if (tableState == null || tableState == TableState.DELETING)
```
keith-turner: Thinking about this some more. Ingoring deleting tablets may prevent unloading them, which would cause delete table to hang.

keith-turner:
All ITs passed for this PR

PR #728, opened 2018-10-26
Revert commons-vfs2 version

matthpeterson:
Roll back to commons-vfs2 version 2.1 due to bugs with 2.2.  See dev mailing list thread: "commons-vfs2.jar 2.2 buggy"


matthpeterson:
Added a comment as requested.

PR #734, opened 2018-10-29
Make the ACCUMULO_MONITOR_BIND_ALL property case-insensitive checking

joshelser:


joshelser:
>  Did you test this?

Ran a quick test locally with `export ACCUMULO_MONITOR_BIND_ALL="True"` and it worked as expected. Pulled it from Stackoverflow, essentially. There was another solution suggested that was pure bash, but that required Bash4 which I didn't think was appropriate. The comments indicated that the `shopt` on `nocasematch` solution should work, but I didn't do a lot of digging either.

joshelser:
Thanks for the quick review, Keith. Pulling this in and merging up. Expecting conflicts on 2.0+

joshelser:
Looks like this whole construct was removed in 2.0 and, by default, services bind to `0.0.0.0`. No special handling required.

PR #750, opened 2018-11-06
Updated accumulo-env.sh to include existing CLASSPATH

mikewalch:


mikewalch:
I think prepend makes a little more sense but I could go either way.

PR #756, opened 2018-11-08
Add additional crypto unit tests

PircDef:
Added additional crypto unit tests including:

	Test for missing KEK uri file
	Test for empty KEK uri file
	Test for unwrapping with an incorrect KEK
	Added IllegalArgumentException to catch in AESKeyUtils.loadKekFromUri
	Renamed references to KeyManager to AESKeyUtils in CryptoTest.java


core/src/test/java/org/apache/accumulo/core/security/crypto/CryptoTest.java
```diff
@@ -294,7 +298,7 @@ public void testKeyManagerGeneratesKey() throws NoSuchAlgorithmException, NoSuch
   }
 
   @Test
-  public void testKeyManagerWrapAndUnwrap()
+  public void testAESKeyUtilsWrapAndUnwrap()
```
milleruntime: Nice catch on the rename.

PR #772, opened 2018-11-19
Verifying whether JAVA_HOME is set or not

deeyes24:
PATCH:  Asserting whether JAVA_HOME is set or not when running "mvn package". When the JAVA_HOME is not set, we are seeing an error that says "bin/javac" not found, "bin/java" not found.
Adding the Assertion that checks for JAVA_HOME is set or not and Errors out with a more intuitive Error Message


keith-turner:
Thanks for the contribution @deeyes24 .  If you would like to be added as a contributor make a PR at https://github.com/apache/accumulo-website/edit/master/pages/people.md

PR #777, opened 2018-11-21
Remove hard coded values from BatchWriterConfig

milleruntime:
Replace hard coded defaults and key names in BatchWriterConfig with values in ClientProperty.
Also improved BATCH_WRITER_DURABILITY javadoc.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/BatchWriterConfig.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/BatchWriterConfig.java
@@ -231,15 +231,15 @@ public class BatchWriterConfig implements Writable {
     // write this out in a human-readable way
     ArrayList<String> fields = new ArrayList<>();
     if (maxMemory != null)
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/apache/accumulo/core/client/BatchWriterConfig.java
```diff
@@ -209,15 +231,15 @@ public void write(DataOutput out) throws IOException {
     // write this out in a human-readable way
     ArrayList<String> fields = new ArrayList<>();
     if (maxMemory != null)
-      addField(fields, "maxMemory", maxMemory);
+      addField(fields, MAX_MEM.getKey(), maxMemory);
```
mikewalch: Could leave this method unchanged.  Not sure if there is a benefit to changing these keys

milleruntime: Yeah its not work breaking serialized data.  I will undo the changes to the serialized code.

PR #779, opened 2018-11-23
Create types in ClientProperty. Closes #778

### LINKED ISSUE #778: Loss of precision configuring BW with new AccumuloClient
The new AccumuloClient builder allows setting Batch writer properties using the BatchWriterConfig object.  This is great but BatchWriterConfig allows setting of different TimeUnits when configuring the batch writer. The AccumuloClient builder always assuming seconds as the TimeUnit so if you set something like milliseconds, there will be loss of precision and user may get unexpected results.  See https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java#L376

milleruntime:
* Change ClientProperty to use millis for batch writer properties

One solution to fix #778.  I also thought of adding different property types to ClientProperty, similar to the way Property uses PropertyType.TIMEDURATION but that required a lot more work. Perhaps that could be a future improvement.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
@@ -356,7 +356,7 @@ public class AccumuloClientImpl implements AccumuloClient {
 
     @Override
     public ConnectionOptions zkTimeout(int timeout) {
-      ClientProperty.INSTANCE_ZOOKEEPERS_TIMEOUT.setTime(properties, (long) timeout);
+      ClientProperty.INSTANCE_ZOOKEEPERS_TIMEOUT.setTimeInMillis(properties, (long) timeout);
       return this;
     }
 
@@ -375,9 +375,9 @@ public class AccumuloClientImpl implements AccumuloClient {
     @Override
     public ConnectionOptions batchWriterConfig(BatchWriterConfig batchWriterConfig) {
       ClientProperty.BATCH_WRITER_MEMORY_MAX.setBytes(properties, batchWriterConfig.getMaxMemory());
-      ClientProperty.BATCH_WRITER_LATENCY_MAX.setTime(properties,
+      ClientProperty.BATCH_WRITER_LATENCY_MAX.setTimeInMillis(properties,
           batchWriterConfig.getMaxLatency(TimeUnit.MILLISECONDS));
-      ClientProperty.BATCH_WRITER_TIMEOUT_MAX.setTime(properties,
+      ClientProperty.BATCH_WRITER_TIMEOUT_MAX.setTimeInMillis(properties,
           batchWriterConfig.getTimeout(TimeUnit.MILLISECONDS));
       setProperty(ClientProperty.BATCH_WRITER_THREADS_MAX, batchWriterConfig.getMaxWriteThreads());
       setProperty(ClientProperty.BATCH_WRITER_DURABILITY,
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
@@ -376,9 +376,9 @@ public class AccumuloClientImpl implements AccumuloClient {
     public ConnectionOptions batchWriterConfig(BatchWriterConfig batchWriterConfig) {
       setProperty(ClientProperty.BATCH_WRITER_MAX_MEMORY_BYTES, batchWriterConfig.getMaxMemory());
       setProperty(ClientProperty.BATCH_WRITER_MAX_LATENCY_SEC,
-          batchWriterConfig.getMaxLatency(TimeUnit.SECONDS));
+          batchWriterConfig.getMaxLatency(TimeUnit.MILLISECONDS));
       setProperty(ClientProperty.BATCH_WRITER_MAX_TIMEOUT_SEC,
-          batchWriterConfig.getTimeout(TimeUnit.SECONDS));
+          batchWriterConfig.getTimeout(TimeUnit.MILLISECONDS));
       setProperty(ClientProperty.BATCH_WRITER_MAX_WRITE_THREADS,
           batchWriterConfig.getMaxWriteThreads());
       setProperty(ClientProperty.BATCH_WRITER_DURABILITY,
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
@@ -376,9 +376,9 @@ public class AccumuloClientImpl implements AccumuloClient {
     public ConnectionOptions batchWriterConfig(BatchWriterConfig batchWriterConfig) {
       setProperty(ClientProperty.BATCH_WRITER_MAX_MEMORY_BYTES, batchWriterConfig.getMaxMemory());
       setProperty(ClientProperty.BATCH_WRITER_MAX_LATENCY_SEC,
-          batchWriterConfig.getMaxLatency(TimeUnit.MILLISECONDS));
+          batchWriterConfig.getMaxLatency(TimeUnit.MILLISECONDS) + "ms");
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -53,10 +53,10 @@
   // BatchWriter
   BATCH_WRITER_MAX_MEMORY_BYTES("batch.writer.max.memory.bytes", "52428800",
       "Max memory (in bytes) to batch before writing"),
-  BATCH_WRITER_MAX_LATENCY_SEC("batch.writer.max.latency.sec", "120",
```
keith-turner: Could use time units like `2s` or `150ms`.  Can use the function `ConfigurationTypeHelper.getTimeInMillis(String)` to do this.  This is what is used by server config.

milleruntime: Do you think I should add types to keep track, like we have in Property?

keith-turner: That would be nice if it avoids making each user of a config prop have to know how to parse it.

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -37,55 +37,64 @@
...
 
   // BatchWriter
-  BATCH_WRITER_MAX_MEMORY_BYTES("batch.writer.max.memory.bytes", "52428800",
-      "Max memory (in bytes) to batch before writing"),
-  BATCH_WRITER_MAX_LATENCY_SEC("batch.writer.max.latency.sec", "120",
-      "Max amount of time (in seconds) to hold data in memory before flushing it"),
-  BATCH_WRITER_MAX_TIMEOUT_SEC("batch.writer.max.timeout.sec", "0",
+  BATCH_WRITER_MAX_MEMORY_BYTES("batch.writer.memory.max", "50M", PropertyType.BYTES,
+      "Max memory (in bytes) to batch before writing", "2.0.0", false),
+  BATCH_WRITER_MAX_LATENCY_SEC("batch.writer.latency.max", "120s", PropertyType.TIMEDURATION,
```
mikewalch: You should update the enum name to match the property name.

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -37,55 +37,64 @@
...
+  AUTH_TOKEN("auth.token", "", PropertyType.STRING,
+      "Authentication token (ex. mypassword, /path/to/keytab)", "2.0.0", true),
 
   // BatchWriter
-  BATCH_WRITER_MAX_MEMORY_BYTES("batch.writer.max.memory.bytes", "52428800",
-      "Max memory (in bytes) to batch before writing"),
-  BATCH_WRITER_MAX_LATENCY_SEC("batch.writer.max.latency.sec", "120",
-      "Max amount of time (in seconds) to hold data in memory before flushing it"),
-  BATCH_WRITER_MAX_TIMEOUT_SEC("batch.writer.max.timeout.sec", "0",
+  BATCH_WRITER_MAX_MEMORY_BYTES("batch.writer.memory.max", "50M", PropertyType.BYTES,
```
keith-turner: Since names changed need to ensure the props file that ends up in the tarball is updated.

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -178,6 +189,22 @@ public Long getLong(Properties properties) {
     return Long.parseLong(value);
   }
 
+  public Long getBytes(Properties properties) {
+    String value = getValue(properties);
```
keith-turner: Could do `checkState(type == PropertyType.BYTES)`

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -178,6 +189,22 @@ public Long getLong(Properties properties) {
...
+    if (value.isEmpty()) {
+      return null;
+    }
+    return ConfigurationTypeHelper.getMemoryAsBytes(value);
+  }
+
+  public Long getTimeInMillis(Properties properties) {
+    String value = getValue(properties);
```
keith-turner: Could do `checkState(type == PropertyType.TIMEDURATION)`

core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
```diff
@@ -376,9 +376,9 @@ public SaslOptions useSasl() {
     public ConnectionOptions batchWriterConfig(BatchWriterConfig batchWriterConfig) {
       setProperty(ClientProperty.BATCH_WRITER_MAX_MEMORY_BYTES, batchWriterConfig.getMaxMemory());
       setProperty(ClientProperty.BATCH_WRITER_MAX_LATENCY_SEC,
-          batchWriterConfig.getMaxLatency(TimeUnit.SECONDS));
+          batchWriterConfig.getMaxLatency(TimeUnit.MILLISECONDS) + "ms");
```
keith-turner: For other props that now have a timeduration type did you check how those are used to see if units need to be added anywhere else?

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
milleruntime: I had this check throwing an exception if it failed but it seemed unnecessary since a Number Format exception will get thrown.  Did you have any other idea how to handle a failure?

keith-turner: This check would be in addition to the checks in `getValue`.  The main purpose of this check is to catch programmer error.  If I call `BATCH_WRITER_MAX_LATENCY_SEC.getBytes(props)` the code may work sometimes.

milleruntime: Oh yeah good point, it won't always fail.  I will add them back.

core/src/main/java/org/apache/accumulo/core/clientImpl/AccumuloClientImpl.java
milleruntime: I'm working on it, thanks.  The ZK timeout is used in a lot more places.  You think I should create a method in ClientProperty that sets the units?  That might be better than having a bunch of concats.

milleruntime: I created setTime() and setBytes() in 80675e6e85.  They are kinda nice.

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
milleruntime: I believe [ClientConfigGenerate](https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/conf/ClientConfigGenerate.java) will handle that.  But I will make sure I update the website.

milleruntime:
@mikewalch I created setTime() and setBytes() in my latest update 80675e6.  Let me know what you think and I will merge

core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -194,6 +219,18 @@ public boolean getBoolean(Properties properties) {
     return Boolean.valueOf(value);
   }
 
+  public void setBytes(Properties properties, Long bytes) {
+    checkState(PropertyType.BYTES == getType(), "Invalid type setting " + "bytes. Type must be "
+        + PropertyType.BYTES + ", not " + getType());
+    properties.setProperty(getKey(), bytes.toString());
+  }
+
+  public void setTime(Properties properties, Long milliseconds) {
```
mikewalch: could be `setMilliseconds` to be clear but I am OK with it as is

mikewalch: or maybe `setTimeInMillis` as there is a `getTimeInMillis`

PR #781, opened 2018-11-27
fixes ACCUMULO-4410: Master didn't not resume balancing after adminis…

ivakegg:
…trative tserver shutdown

  - Updated the master to reduce the serversToShutdown to only those that are current.
  - Removed the statefulness of the ShutdownTServer fate operation as that state is never actually maintained.

### DIFF
--- a/server/master/src/main/java/org/apache/accumulo/master/Master.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/Master.java
@@ -1554,49 +1554,57 @@ public class Master extends AccumuloServerContext
   @Override
   public void update(LiveTServerSet current, Set<TServerInstance> deleted,
       Set<TServerInstance> added) {
-    DeadServerList obit = new DeadServerList(
-        ZooUtil.getRoot(getInstance()) + Constants.ZDEADTSERVERS);
-    if (added.size() > 0) {
-      log.info("New servers: " + added);
-      for (TServerInstance up : added)
-        obit.delete(up.hostPort());
-    }
-    for (TServerInstance dead : deleted) {
-      String cause = "unexpected failure";
-      if (serversToShutdown.contains(dead))
-        cause = "clean shutdown"; // maybe an incorrect assumption
-      if (!getMasterGoalState().equals(MasterGoalState.CLEAN_STOP))
-        obit.post(dead.hostPort(), cause);
-    }
-
-    Set<TServerInstance> unexpected = new HashSet<>(deleted);
-    unexpected.removeAll(this.serversToShutdown);
-    if (unexpected.size() > 0) {
-      if (stillMaster() && !getMasterGoalState().equals(MasterGoalState.CLEAN_STOP)) {
-        log.warn("Lost servers " + unexpected);
+    // if we have deleted or added tservers, then adjust our dead server list
+    if (!deleted.isEmpty() || !added.isEmpty()) {
+      DeadServerList obit = new DeadServerList(
+          ZooUtil.getRoot(getInstance()) + Constants.ZDEADTSERVERS);
+      if (added.size() > 0) {
+        log.info("New servers: " + added);
+        for (TServerInstance up : added)
+          obit.delete(up.hostPort());
       }
-    }
-    serversToShutdown.removeAll(deleted);
-    badServers.keySet().removeAll(deleted);
-    // clear out any bad server with the same host/port as a new server
-    synchronized (badServers) {
-      cleanListByHostAndPort(badServers.keySet(), deleted, added);
-    }
-    synchronized (serversToShutdown) {
-      cleanListByHostAndPort(serversToShutdown, deleted, added);
-    }
+      for (TServerInstance dead : deleted) {
+        String cause = "unexpected failure";
+        if (serversToShutdown.contains(dead))
+          cause = "clean shutdown"; // maybe an incorrect assumption
+        if (!getMasterGoalState().equals(MasterGoalState.CLEAN_STOP))
+          obit.post(dead.hostPort(), cause);
+      }
 
-    synchronized (migrations) {
-      Iterator<Entry<KeyExtent,TServerInstance>> iter = migrations.entrySet().iterator();
-      while (iter.hasNext()) {
-        Entry<KeyExtent,TServerInstance> entry = iter.next();
-        if (deleted.contains(entry.getValue())) {
-          log.info("Canceling migration of " + entry.getKey() + " to " + entry.getValue());
-          iter.remove();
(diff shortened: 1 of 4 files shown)


ivakegg:
I have tested and verified this on a relatively large test system.

server/master/src/main/java/org/apache/accumulo/master/tserverOps/ShutdownTServer.java
```diff
@@ -52,12 +52,7 @@ public long isReady(long tid, Master master) throws Exception {
     }
 
     // Inform the master that we want this server to shutdown
-    // We don't want to spam the master with shutdown requests, so
-    // only send this request once
-    if (!requestedShutdown) {
-      master.shutdownTServer(server);
-      requestedShutdown = true;
```
ivakegg: This state is never actually persisted as a new one of this object is created between each isReady call.

mjwall:
Thanks @ivakegg, am seeing this currently in 1.9.2.

ivakegg:
I don't know what to say about the Jenkins fail.  The log says it failed to merge, yet I have no problem merging and the merge check succeeds.

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1586,6 +1586,11 @@ public void update(LiveTServerSet current, Set<TServerInstance> deleted,
       cleanListByHostAndPort(serversToShutdown, deleted, added);
     }
```
ivakegg: I am wondering if I can simply remove all of the removes above this line now since doing a retainAll of the current servers theoretically would take care of those.

ivakegg: Upon further examination, I believe I may be able to only remove line 1579.  However it is certainly not hurting anything.  I will leave this as is for now.

keith-turner: I was wondering the same thing.  It is a bit of a mess.

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1586,6 +1586,11 @@ public void update(LiveTServerSet current, Set<TServerInstance> deleted,
...
 
+    // clear out any servers that are no longer current
+    // this is needed when we are using a fate operation to shutdown a tserver as it
+    // will continue to add the server to the serversToShutdown (ACCUMULO-4410)
+    serversToShutdown.retainAll(current.getCurrentServers());
```
keith-turner: I was looking at how this `update()` method is called an noticed that its conditionally called when the set of tservers changes.  This made me think that could be still be race conditions where junk lingers in `serversToShutdown`.

This race condition made me wonder if `Master.StatusThread.updateStatus()` could be more defensive.   I thought this would be a really simple change but realized the need to synchronize complicated it.   Below is the result of my poking at the code.  I don't like the complexity below, wish I could just call `Sets.intersect().isEmpty()` directly in the if stmt.

```java
    private long updateStatus()
        throws AccumuloException, AccumuloSecurityException, TableNotFoundException {
      Set<TServerInstance> currentServers = tserverSet.getCurrentServers();
      tserverStatus = gatherTableInformation(currentServers);
      checkForHeldServer(tserverStatus);

      // See ACCUMULO-XYZ
      boolean stsEmpty;
      Collection<TServerInstance> stsView = null;
      synchronized (serversToShutdown) {
        //need to sync to avoid concurent mod exception while iterating (impl of Sets.intersection iterates)
        SetView<TServerInstance> intersection = Sets.intersection(serversToShutdown, currentServers);
        stsEmpty = intersection.isEmpty();
        if(!stsEmpty) {
          stsView = new ArrayList<>(intersection);
        }
      }

      if (!badServers.isEmpty()) {
        log.debug(
            "not balancing because the balance information is out-of-date " + badServers.keySet());
      } else if (notHosted() > 0) {
        log.debug("not balancing because there are unhosted tablets: " + notHosted());
      } else if (getMasterGoalState() == MasterGoalState.CLEAN_STOP) {
        log.debug("not balancing because the master is attempting to stop cleanly");
      } else if (!stsEmpty) {
        log.debug("not balancing while shutting down servers " + stsView);
      } else {
        for (TabletGroupWatcher tgw : watchers) {
          if (!tgw.isSameTserversAsLastScan(currentServers)) {
            log.debug("not balancing just yet, as collection of live tservers is in flux");
            return DEFAULT_WAIT_FOR_WATCHER;
          }
        }
        return balanceTablets();
      }
      return DEFAULT_WAIT_FOR_WATCHER;
    }
```

I also thought of doing `serversToShutdown.retainAll(currentServers)` in `Master.StatusThread.updateStatus()` instead of using `Sets.intersect()`.  However this just feels wrong to me.   I like the idea of `Master.StatusThread.updateStatus()` only reading these sets.

server/master/src/main/java/org/apache/accumulo/master/tserverOps/ShutdownTServer.java
```diff
@@ -68,8 +63,6 @@ public long isReady(long tid, Master master) throws Exception {
           TabletServerStatus status = connection.getTableMap(false);
           if (status.tableMap != null && status.tableMap.isEmpty()) {
             log.info("tablet server hosts no tablets " + server);
-            connection.halt(master.getMasterLock());
```
keith-turner: This may have been called in isReady intentionally in-order to get some re-try type behavior, not sure.  The command later on in the method made me think this.

server/master/src/main/java/org/apache/accumulo/master/Master.java
ivakegg: I suppose we could do this in addition to the retainAll in the update() call as a safety, but I question whether the additional synchronization is worth it.

server/master/src/main/java/org/apache/accumulo/master/tserverOps/ShutdownTServer.java
ivakegg: Alternatively, in addition to sending a halt we could remove the zookeeper lock here.

ivakegg: Working through the logic, I believe the halt was actually better placed in the isReady call.  Reverting that part of the change.

server/master/src/main/java/org/apache/accumulo/master/Master.java
ivakegg: Updated the LiveTServerSet to call the Master.update(...) method everytime, and pushing the check for changes into that method.  This will ensure that the serversToShutdown is cleaned up and avoids having to add this piece of code.

PR #786, opened 2018-11-29
fixes #785 avoid using system classloader in client rfile code

### LINKED ISSUE #785: AccumuloClassLoader is being used in client API to create RFiles
While the creating an RFile in bulk import randomwalk test, the follow warnings were printed.  

```
2018-11-28 16:20:29,126 [start.classloader.AccumuloClassLoader] WARN : Failed to load Accumulo configuration 'accumulo.properties' from classpath
2018-11-28 16:20:29,126 [start.classloader.AccumuloClassLoader] WARN : Using default value '/tmp' for 'general.vfs.cache.dir' as there is no Accumulo configuration on classpath
2018-11-28 16:20:29,128 [start.classloader.AccumuloClassLoader] WARN : Using default value 'null' for 'general.classpaths' as there is no Accumulo configuration on classpath
2018-11-28 16:20:29,128 [start.classloader.AccumuloClassLoader] WARN : Using default value '' for 'general.vfs.classpaths' as there is no Accumulo configuration on classpath
2018-11-28 16:20:29,129 [start.classloader.AccumuloClassLoader] WARN : Using default value '$ACCUMULO_HOME/lib/ext/[^.].*.jar' for 'general.dynamic.classpaths' as there is no Accumulo configuration on classpath
2018-11-28 16:20:29,131 [start.classloader.AccumuloClassLoader] WARN : Using default value '/tmp' for 'general.vfs.cache.dir' as there is no Accumulo configuration on classpath
```

Should probably avoid using `accumulo-start` module code in API.

keith-turner:
Properly making this change cascaded into wider changes.  Trying to avoid the
system classloader led to the discovery of code in BCFile that created a
crypto service using static config and the system classloader.  I realized
this code needed to go which resulted in making the crypto service mandatory
when opening an RFile.  This led to pushing the  ServerContext further into
code, which is a good thing.

The following actions were taken :

 * Made CryptoService a required argument for opening a file in FileOperations
 * MAde CryptoServiceFactory require the caller to specify if accumulo
   classloader should be used.
 * Added CryptoServiceFactory.newDefaultInstance() method and called this in
   places like test code that just need the default crypto service.
 * Pushed ServerContext further into code.  One nice thing about this is that
   things available from ServerContext no longer needed to passed.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
+++ b/core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
@@ -24,22 +24,21 @@ import org.apache.accumulo.core.spi.crypto.CryptoService;
 
 public class CryptoServiceFactory {
 
-  /**
-   * Create a new crypto service configured in {@link Property#INSTANCE_CRYPTO_SERVICE}.
-   *
-   * @param useAccumuloClassloader
-   *          Determine if Accumulo class loader is used. When calling from accumulo server code,
-   *          pass true. When calling from Accumulo client code, pass false.
-   */
-  public static CryptoService newInstance(AccumuloConfiguration conf,
-      boolean useAccumuloClassloader) {
+  public enum ClassloaderType {
+    // Use the Accumulo custom classloader. Should only be used by Accumulo server side code.
+    ACCUMULO,
+    // Use basic Java classloading mechanism. Should be use by Accumulo client code.
+    JAVA
+  }
+
+  public static CryptoService newInstance(AccumuloConfiguration conf, ClassloaderType ct) {
 
     CryptoService newCryptoService;
 
-    if (useAccumuloClassloader) {
+    if (ct == ClassloaderType.ACCUMULO) {
       newCryptoService = Property.createInstanceFromPropertyName(conf,
           Property.INSTANCE_CRYPTO_SERVICE, CryptoService.class, new NoCryptoService());
-    } else {
+    } else if (ct == ClassloaderType.JAVA) {
       String clazzName = conf.get(Property.INSTANCE_CRYPTO_SERVICE);
       if (clazzName == null || clazzName.trim().isEmpty()) {
         newCryptoService = new NoCryptoService();
@@ -51,6 +50,8 @@ public class CryptoServiceFactory {
           throw new RuntimeException(e);
         }
       }
+    } else {
+      throw new IllegalArgumentException();
     }
 
     newCryptoService.init(conf.getAllPropertiesWithPrefix(Property.INSTANCE_CRYPTO_PREFIX));
@@ -58,6 +59,6 @@ public class CryptoServiceFactory {
   }
 
   public static CryptoService newDefaultInstance() {
-    return newInstance(DefaultConfiguration.getInstance(), false);
+    return newInstance(DefaultConfiguration.getInstance(), ClassloaderType.JAVA);
   }
 }
--- a/core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
+++ b/core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
@@ -17,6 +17,7 @@
 package org.apache.accumulo.core.crypto;
 
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
+import org.apache.accumulo.core.conf.DefaultConfiguration;
 import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.core.cryptoImpl.NoCryptoService;
 import org.apache.accumulo.core.spi.crypto.CryptoService;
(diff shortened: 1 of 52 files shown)


core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
```diff
@@ -25,11 +26,38 @@
...
+   *
+   * @param useAccumuloClassloader
+   *          Determine if Accumulo class loader is used. When calling from accumulo server code,
+   *          pass true. When calling from Accumulo client code, pass false.
    */
-  public static CryptoService newInstance(AccumuloConfiguration conf) {
-    CryptoService newCryptoService = Property.createInstanceFromPropertyName(conf,
-        Property.INSTANCE_CRYPTO_SERVICE, CryptoService.class, new NoCryptoService());
+  public static CryptoService newInstance(AccumuloConfiguration conf,
+      boolean useAccumuloClassloader) {
```
milleruntime: Could make the parameter an enum and then set useAccumuloClassloader to the appropriate value.  I was thinking something like Caller.CLIENT or Scope.CLIENT, similar to what we did with Scope in CryptoEnvironment.  There is some dislike for booleans as parameters.  I found myself having to go back and check the javadoc that false = client.

core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
```diff
@@ -25,11 +26,38 @@
...
+        }
+      }
+    }
+
     newCryptoService.init(conf.getAllPropertiesWithPrefix(Property.INSTANCE_CRYPTO_PREFIX));
     return newCryptoService;
   }
+
+  public static CryptoService newDefaultInstance() {
+    return newInstance(DefaultConfiguration.getInstance(), false);
```
milleruntime: I think this should get what is configured and not just the default.  Calling newDefaultInstance is the same as using NoCryptoService, which is what we have now but may break special parts of the code when crypto is configured.

core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
```diff
@@ -25,11 +26,38 @@
...
+  }
```
milleruntime: Could also have a static method that sets NoCryptoService.  Which would mainly be used for tests but at least it wouldn't call static code.

server/base/src/main/java/org/apache/accumulo/server/util/FileUtil.java
```diff
@@ -105,8 +106,8 @@ private static Path createTmpDir(AccumuloConfiguration acuConf, VolumeManager fs
     return result;
   }
 
-  public static Collection<String> reduceFiles(AccumuloConfiguration acuConf, Configuration conf,
-      VolumeManager fs, Text prevEndRow, Text endRow, Collection<String> mapFiles, int maxFiles,
+  public static Collection<String> reduceFiles(ServerContext context, AccumuloConfiguration acuConf,
+      Configuration conf, Text prevEndRow, Text endRow, Collection<String> mapFiles, int maxFiles,
```
milleruntime: Nice use of ServerContext

core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
keith-turner: I have not seen a situation in the test where this would be needed.

core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
keith-turner: > I think this should get what is configured and not just the default.

Configured where?  The purpose of this method is for code that has no meaningful config can call it.

>  Calling newDefaultInstance is the same as using NoCryptoService

At first I was creating NoCryptoService everywhere in the code.  Then I realized that if we ever changed the default that all of this code would need to be changed.  So I created this method which at first just returned `new NoCryptoService()`.  Then I realized that if we ever change the default that we will need to change this method also.  So I changed it to use the default config.  That way if the default is ever changed this code does not need to change.

milleruntime: I was just concerned that DefaultConfiguration.getInstance() is calling static code that we are trying to avoid.  But if its not, then yeah it would be better to get the default everywhere.

server/base/src/main/java/org/apache/accumulo/server/util/FileUtil.java
keith-turner: I made more use of it in bd64447

core/src/main/java/org/apache/accumulo/core/crypto/CryptoServiceFactory.java
keith-turner: This would make reading the code easier.

keith-turner: done in 455f2ca

PR #790, opened 2018-11-30
Make durability propertyType case insensitive

milleruntime:
* Durability has uppercase values but the PropertyType.isValidFromat check fails if the
enum is set as a value.

This will fix the currently failing AccumuloOutputFormatIT

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/conf/PropertyType.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/PropertyType.java
@@ -116,7 +116,7 @@ public enum PropertyType {
       "A list of fully qualified java class names representing classes on the classpath.\n"
           + "An example is 'java.lang.String', rather than 'String'"),
 
-  DURABILITY("durability", in(true, null, "default", "none", "log", "flush", "sync"),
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/accumulo/core/conf/PropertyType.java
```diff
@@ -116,7 +116,7 @@
       "A list of fully qualified java class names representing classes on the classpath.\n"
           + "An example is 'java.lang.String', rather than 'String'"),
 
-  DURABILITY("durability", in(true, null, "default", "none", "log", "flush", "sync"),
+  DURABILITY("durability", in(false, null, "default", "none", "log", "flush", "sync"),
```
keith-turner: If making this sanity check case insensitive then need to ensure that code that handles durability is also case insensitive.

milleruntime: This seems to be the method most used, from [DurabilityImpl.java](https://github.com/apache/accumulo/blob/master/core/src/main/java/org/apache/accumulo/core/clientImpl/DurabilityImpl.java#L39): 
```java 
public static Durability fromString(String value) {
    return Durability.valueOf(value.toUpperCase());
  }
```

PR #795, opened 2018-12-04
Unit test and validation for client properties

mikewalch:


core/src/main/java/org/apache/accumulo/core/conf/ClientProperty.java
```diff
@@ -302,4 +302,44 @@ public static void setAuthenticationToken(Properties properties, AuthenticationT
...
   }
+
+  public static boolean isPropertySet(Properties properties, ClientProperty prop) {
+    return properties.containsKey(prop.getKey()) && !prop.getValue(properties).isEmpty();
+  }
+
+  /**
+   * @return True if Properties contains all required properties
+   */
+  public static boolean hasRequired(Properties properties) {
```
keith-turner: Seems this is only called in test code?

mikewalch: It is. I'll remove it.

PR #796, opened 2018-12-04
Replace ClientInfo with properties in MR

milleruntime:
Work towards fixing #788

### DIFF
--- a/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloInputFormat.java
+++ b/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloInputFormat.java
@@ -41,7 +41,7 @@ import org.slf4j.LoggerFactory;
  * job using the {@link #configure()} method, which provides a fluent API. For Example:
  *
  * <pre>
- * AccumuloInputFormat.configure().clientProperties(info).table(name).auths(auths) // required
+ * AccumuloInputFormat.configure().clientProperties(props).table(name).auths(auths) // required
  *     .addIterator(iter1).ranges(ranges).fetchColumns(columns).executionHints(hints)
  *     .samplerConfiguration(sampleConf).disableAutoAdjustRanges() // enabled by default
  *     .scanIsolation() // not available with batchScan()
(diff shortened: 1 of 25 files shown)


hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloInputFormat.java
```diff
@@ -41,7 +41,7 @@
  * job using the {@link #configure()} method, which provides a fluent API. For Example:
  *
  * <pre>
- * AccumuloInputFormat.configure().clientInfo(info).table(name).auths(auths) // required
+ * AccumuloInputFormat.configure().clientProperties(info).table(name).auths(auths) // required
```
mikewalch: could use props instead of info

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloRowInputFormat.java
```diff
@@ -43,7 +43,7 @@
  * For Example:
  *
  * <pre>
- * AccumuloRowInputFormat.configure().clientInfo(info).table(name).auths(auths) // required
+ * AccumuloRowInputFormat.configure().clientProperties(info).table(name).auths(auths) // required
```
mikewalch: props instead of info

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoopImpl/mapreduce/OutputFormatBuilderImpl.java
```diff
@@ -39,7 +40,7 @@
   boolean simulationMode = false;
 
   @Override
-  public OutputFormatBuilder.OutputOptions<T> clientInfo(ClientInfo clientInfo) {
+  public OutputFormatBuilder.OutputOptions<T> clientProperties(Properties clientProperties) {
```
mikewalch: ClientInfo.from() is needed here

milleruntime: Thanks, fixed in 1a5df1a

PR #797, opened 2018-12-05
Improved tablet metadata abstraction layer and used it more.

keith-turner:


core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
```diff
@@ -283,57 +251,154 @@ public Options saveKeyValues() {
...
+    Options fetchCompactId();
+
+    Options fetchDir();
+
+    Options fetchFiles();
+
+    Options fetchFlushId();
+
+    Options fetchLast();
+
```
mikewalch: Could add javadoc to all of these options.

core/src/main/java/org/apache/accumulo/core/util/Merge.java
```diff
@@ -218,55 +212,23 @@ protected void merge(AccumuloClient client, String table, List<Size> sizes, int
...
     Table.ID tableId;
-    Scanner scanner;
+    TabletsMetadata tablets;
     try {
       ClientContext context = new ClientContext(client.info());
       tableId = Tables.getTableId(context, tablename);
-      scanner = client.createScanner(MetadataTable.NAME, Authorizations.EMPTY);
+      tablets = TabletsMetadata.builder().scanMetadataTable()
+          .overRange(new KeyExtent(tableId, end, start).toMetadataRange()).fetchFiles().fetchPrev()
+          .build(context);
```
milleruntime: This is great!

core/src/main/java/org/apache/accumulo/core/tabletserver/log/LogEntry.java
```diff
@@ -80,21 +80,25 @@ public static LogEntry fromBytes(byte bytes[]) throws IOException {
...
+    parts = value.split("\\|")[0].split(";");
     String filename = parts[parts.length - 1];
     long timestamp = key.getTimestamp();
     return new LogEntry(extent, timestamp, server, filename);
   }
 
+  public static LogEntry fromKeyValue(Key key, Value value) {
+    return fromKeyValue(key, value.toString());
+  }
+
```
milleruntime: This just a general fix, unrelated to your bigger changes?

keith-turner: In the code that parses a row to a TabletMetadata object I had already converted the value to string and I wanted to avoid duplicating that work.

core/src/main/java/org/apache/accumulo/core/clientImpl/ConcurrentKeyExtentCache.java
```diff
@@ -85,8 +85,8 @@ protected void updateCache(KeyExtent e) {
   @VisibleForTesting
   protected Stream<KeyExtent> lookupExtents(Text row)
       throws TableNotFoundException, AccumuloException, AccumuloSecurityException {
-    return MetadataScanner.builder().from(ctx).scanMetadataTable().overRange(tableId, row, null)
-        .checkConsistency().fetchPrev().build().stream().limit(100).map(TabletMetadata::getExtent);
+    return TabletsMetadata.builder().forTable(tableId).overlapping(row, null).checkConsistency()
+        .fetchPrev().build(ctx).stream().limit(100).map(TabletMetadata::getExtent);
```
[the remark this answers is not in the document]
keith-turner: Actually the test is not in 1.9.  I will still fix separately.

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
keith-turner: I think it would be better to document the metadata table schema in one place and have this class refer to that.

Issue #749, opened 2018-11-06
Add Multi table functionality to new MapReduce API


milleruntime:
Follow on work for #743 to maintain functionality that AccumuloMultiTableInputFormat provided in the old mapreduce API.

milleruntime:
I am wondering what the best way to do this would be... have an intermediate method such as "add()" to signal more tables are going to be added?
```java
AccumuloInputFormat.configure().clientInfo(info).table(name).auths(auths).add()
    .table(name2).auths(auths2).ranges(myRanges).add()
    .table(name3).auths(auths3).fetchColumns(cols).samplerConfiguration(conf).store(job);
```
This would allow for looping, if tables have the same configuration:
```java
TableParams opts = AccumuloInputFormat.configure().clientInfo(info);
for (int x = 0; x < 10; x++) {
    opts = opts.table("table" + x).auths(myAuths).ranges(myRanges).add()
}
opts.store(job);
```

Or maybe call the method "next()"?

keith-turner:
Seems like the following could work without `add()`.  Each time `table()` is called it implies you are starting a new table.

```java
AccumuloInputFormat.configure().clientInfo(info)
    .table(name).auths(auths)
    .table(name2).auths(auths2).ranges(myRanges)
    .table(name3).auths(auths3).fetchColumns(cols).samplerConfiguration(conf)
    .store(job);
```

keith-turner:
Another possibility is that the entire fluent chain is called for each table.

```java
AccumuloInputFormat.configure().clientInfo(info).table(name).auths(auths).store(job);
AccumuloInputFormat.configure().clientInfo(info).table(name2).auths(auths2).ranges(myRanges).store(job);
AccumuloInputFormat.configure().clientInfo(info).table(name3).auths(auths3).fetchColumns(cols).samplerConfiguration(conf).store(job);
```

milleruntime:
> Seems like the following could work without add(). Each time table() is called it implies you are starting a new table.

OK I like this one.

PR #812, opened 2018-12-07
Make new MR API accept booleans. Closes #803

### LINKED ISSUE #803: Pass boolean variables to Accumulo MapReduce builder methods
I was trying to convert [ContinuousVerify](https://github.com/apache/accumulo-testing/blob/master/src/main/java/org/apache/accumulo/testing/continuous/ContinuousVerify.java) to use the new builder.  In some cases, `scanOffline` should be set for AccumuloInputFormat but all of the time.  The builder could be created twice in an `if-else` or it could be saved off into variable mid build but what would be better is if you could pass in a boolean to the `scanOffline()` method and have just one use of the builder.  

All methods like this in the builder should be converted to accept a boolean as I think this a common issue.

milleruntime:


hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloInputFormat.java
```diff
@@ -43,7 +43,7 @@
  * <pre>
  * AccumuloInputFormat.configure().clientProperties(props).table(name) // required
  *     .auths(auths).addIterator(iter1).ranges(ranges).fetchColumns(columns).executionHints(hints)
- *     .samplerConfiguration(sampleConf).disableAutoAdjustRanges() // enabled by default
+ *     .samplerConfiguration(sampleConf).autoAdjustRanges() // enabled by default
```
mikewalch: Need to update javadoc to use boolean

hadoop-mapreduce/src/test/java/org/apache/accumulo/hadoop/its/mapreduce/AccumuloInputFormatIT.java
```diff
@@ -148,22 +151,21 @@ public void testGetSplits() throws Exception {
...
     AccumuloInputFormat.configure().clientProperties(getClientInfo().getProperties()).table(table)
-        .auths(Authorizations.EMPTY).batchScan().store(job);
-    try {
-      inputFormat.getSplits(job);
-      fail("An exception should have been thrown");
-    } catch (IOException e) {}
+        .auths(Authorizations.EMPTY).batchScan(true).store(job);
+
+    exception.expect(IllegalArgumentException.class);
+    inputFormat.getSplits(job);
```
mikewalch: Could add an illegal arg exception to store also

milleruntime: Yeah I like that.  I will move the checks to a method to share.

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloInputFormat.java
milleruntime: Thanks fixed in 338bb74

hadoop-mapreduce/src/test/java/org/apache/accumulo/hadoop/its/mapreduce/AccumuloInputFormatIT.java
milleruntime: Actually, I will do this in #749 since they are verified on a per table basis.

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloInputFormat.java
mikewalch: It still needs to be fixed in AccumuloInputFormat & AccumuloRowInputFormat

PR #830, opened 2018-12-14
#820 - Fix broken ITs due to client issues

mikewalch:


mikewalch:
This fixes the ITs that are failing.  Hide whitespace when you review.

PR #833, opened 2018-12-17
Merge ClientContext and AccumuloClientImpl

mikewalch:
* Removed AccumuloClientImpl and kept ClientContext
  class which implements AccumuloClient
* Removed use of context.getClient() as ClientContext
  implements AccumuloClient
* Limited instantiation of ClientContext by casting
  AccumuloClient to ClientContext


server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -644,8 +644,8 @@ public TableManager getTableManager() {
     return context.getTableManager();
   }
 
-  public AccumuloClient getClient() throws AccumuloSecurityException, AccumuloException {
```
keith-turner: The master has a getContext method.  So this method could be dropped entirely.  May be able to change the impl of getClient to call getContext and then inline the method.

mikewalch: Done in c491aff

PR #838, opened 2018-12-18
Fix some codestyle issues that are not automated.

cjmctague:
- Adjust the comparisons so that the constant is on the right side of
of the comparison. This was the pattern found throughout the majority
of the project. Also logically makes more sense.
- Add missing @Override annotations


keith-turner:
@cjmctague what do you mean by `that are not automated` in your commit message?  Does that refer to these not being automatically done in the maven build?  What tool did you use to make these changes?

cjmctague:
@keith-turner Yes, that they are not checked during the maven build with the formatter or the checkstyles. Intellij said there were about 650ish instances of the constant being on the left and 6500ish of them being on the right in comparisons. So I had it refactor the 650 instances to match the rest of the project.

PR #840, opened 2018-12-18
fixes #819 Handle bad locality group config.

### LINKED ISSUE #819: Creating inconsistent locality groups causes compaction failures
If the set of locality groups on a table are not consistent (e.g. overlapping), then compactions will start to fail.  Normally this is fine in the short term except that in 1.8, 1.9, and possibly 2.0 this will actually cause a situation that cannot be remedied until the tservers are restarted (documented in another ticket).  This ticket is to suggest that we prevent creating inconsistent locality group configurations in the first place.

keith-turner:
Two changes were made for bad locality group config.  First, when setting
properties a warning  is issued if the change results in bad LG config.
Second, minor compactions will log a warning and proceed with no LG config
if the config is bad.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/impl/InstanceOperationsImpl.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/impl/InstanceOperationsImpl.java
@@ -43,12 +43,15 @@ import org.apache.accumulo.core.tabletserver.thrift.TabletClientService.Client;
 import org.apache.accumulo.core.trace.Tracer;
 import org.apache.accumulo.core.util.AddressUtil;
 import org.apache.accumulo.core.util.HostAndPort;
+import org.apache.accumulo.core.util.LocalityGroupUtil;
+import org.apache.accumulo.core.util.LocalityGroupUtil.LocalityGroupConfigurationError;
 import org.apache.accumulo.core.zookeeper.ZooUtil;
 import org.apache.accumulo.fate.zookeeper.ZooCache;
 import org.apache.accumulo.fate.zookeeper.ZooCacheFactory;
 import org.apache.thrift.TException;
 import org.apache.thrift.transport.TTransport;
 import org.apache.thrift.transport.TTransportException;
+import org.slf4j.LoggerFactory;
 
 /**
  * Provides a class for administering the accumulo instance
@@ -72,6 +75,8 @@ public class InstanceOperationsImpl implements InstanceOperations {
         client.setSystemProperty(Tracer.traceInfo(), context.rpcCreds(), property, value);
       }
     });
+
+    checkLocalityGroups(property);
   }
 
   @Override
@@ -84,6 +89,22 @@ public class InstanceOperationsImpl implements InstanceOperations {
         client.removeSystemProperty(Tracer.traceInfo(), context.rpcCreds(), property);
       }
     });
+
+    checkLocalityGroups(property);
+  }
+
+  void checkLocalityGroups(String propChanged) throws AccumuloSecurityException, AccumuloException {
+    if (LocalityGroupUtil.isLocalityGroupProperty(propChanged)) {
+      try {
+        LocalityGroupUtil.checkLocalityGroups(getSystemConfiguration().entrySet());
+      } catch (LocalityGroupConfigurationError | RuntimeException e) {
+        LoggerFactory.getLogger(this.getClass()).warn("Changing '" + propChanged
+            + "' resulted in bad locality group config. This may be a transient situation since "
+            + "the config spreads over multiple properties. Setting properties in a different "
+            + "order may help. Even though this warning was displayed, the property was updated. "
+            + "Please check your config to ensure consistency.", e);
+      }
+    }
   }
 
   @Override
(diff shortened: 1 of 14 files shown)


core/src/main/java/org/apache/accumulo/core/client/impl/InstanceOperationsImpl.java
```diff
@@ -84,6 +89,22 @@ public void execute(MasterClientService.Client client) throws Exception {
...
+  void checkLocalityGroups(String propChanged) throws AccumuloSecurityException, AccumuloException {
+    if (LocalityGroupUtil.isLocalityGroupProperty(propChanged)) {
+      try {
+        LocalityGroupUtil.checkLocalityGroups(getSystemConfiguration().entrySet());
+      } catch (LocalityGroupConfigurationError | RuntimeException e) {
+        LoggerFactory.getLogger(this.getClass()).warn("Changing '" + propChanged
+            + "' resulted in bad locality group config. This may be a transient situation since "
+            + "the config spreads over multiple properties. Setting properties in a different "
+            + "order may help. Even though this warning was displayed, the property was updated. "
+            + "Please check your config to ensure consistency.", e);
```
milleruntime: Why update the property if it is going to result in a bad locality group?  If it is transient, then I would think that would be less reason to update the property.  If it is not transient, then I would think you'd want to keep warning until its fixed.

test/src/main/java/org/apache/accumulo/test/functional/BadLocalityGroupMincIT.java
```diff
@@ -0,0 +1,89 @@
...
+import org.apache.accumulo.core.data.Mutation;
+import org.apache.accumulo.core.data.Value;
+import org.apache.accumulo.core.security.Authorizations;
+import org.apache.accumulo.harness.AccumuloClusterHarness;
+import org.apache.hadoop.io.Text;
+import org.junit.Test;
+
+import com.google.common.collect.Iterables;
+
+public class BadLocalityGroupMincIT extends AccumuloClusterHarness {
```
milleruntime: Since we have some long running tests, it would be nice to have a javadoc that this test should not time out.  Or something mentioning that the last delete should not hang.

core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
```diff
@@ -1000,17 +1032,21 @@ public void setLocalityGroups(String tableName, Map<String,Set<Text>> groups)
...
+      }
+
       all.addAll(entry.getValue());
     }
 
     for (Entry<String,Set<Text>> entry : groups.entrySet()) {
       Set<Text> colFams = entry.getValue();
       String value = LocalityGroupUtil.encodeColumnFamilies(colFams);
-      setProperty(tableName, Property.TABLE_LOCALITY_GROUP_PREFIX + entry.getKey(), value);
+      setPropertyNoChecks(tableName, Property.TABLE_LOCALITY_GROUP_PREFIX + entry.getKey(), value);
```
milleruntime: Wouldn't you also want to check here?

core/src/main/java/org/apache/accumulo/core/client/impl/InstanceOperationsImpl.java
keith-turner: > Why update the property if it is going to result in a bad locality group?

It's because locality group config is spread across multiple properties.  So there could be existing user code like the following.

```
 setProp(k1,v1); //makes LG group config incomplete/bad
 setProp(k2,v2); //makes LG group config incomplete/bad
 setProp(k3,v3); //makes LG config complete/good
```

So the transient bad state is very short lived and the user may have never noticed any problems from it.  I do not want to break existing code that is working for 1.9.3.  However it would be good to bring this to the users attention.

core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
keith-turner: This code is setting properties in the correct order and I did not want it to read all properties for each property set by this code.

test/src/main/java/org/apache/accumulo/test/functional/BadLocalityGroupMincIT.java
keith-turner: Oh I should remove that comment because I don't expect delete table to hang.  That comment came from me starting off by copying a test and I noticed it had that delete at the end so I left it.

core/src/main/java/org/apache/accumulo/core/client/impl/TableOperationsImpl.java
```diff
@@ -929,35 +930,66 @@ public void setProperty(final String tableName, final String property, final Str
...
-      });
+      removePropertyNoChecks(tableName, property);
+
+      checkLocalityGroups(tableName, property);
     } catch (TableNotFoundException e) {
       throw new AccumuloException(e);
     }
   }
 
+  private void removePropertyNoChecks(final String tableName, final String property)
```
milleruntime: Could rename this method to "removePropNoLocalityChecks".  Or something to indicate locality groups.

PR #842, opened 2018-12-19
More cleanup for WAL

milleruntime:
* Actually log MUTATION for single mutation
* Make TabletMutations constructor take CommitSession
* Simplify calls to defineTablet
* Remove unused datafile path param
* Add checks for debug

Only synchronize on write method

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletMutations.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletMutations.java
@@ -27,7 +27,8 @@ public class TabletMutations {
   private final List<Mutation> mutations;
   private final Durability durability;
 
-  public TabletMutations(CommitSession commitSession, List<Mutation> mutations, Durability durability) {
+  public TabletMutations(CommitSession commitSession, List<Mutation> mutations,
+      Durability durability) {
     this.commitSession = commitSession;
     this.mutations = mutations;
     this.durability = durability;
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletMutations.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletMutations.java
@@ -20,16 +20,15 @@ import java.util.List;
 
 import org.apache.accumulo.core.client.Durability;
 import org.apache.accumulo.core.data.Mutation;
+import org.apache.accumulo.tserver.tablet.CommitSession;
 
 public class TabletMutations {
-  private final int tid;
-  private final long seq;
+  private CommitSession commitSession;
   private final List<Mutation> mutations;
   private final Durability durability;
 
-  public TabletMutations(int tid, long seq, List<Mutation> mutations, Durability durability) {
-    this.tid = tid;
-    this.seq = seq;
+  public TabletMutations(CommitSession commitSession, List<Mutation> mutations, Durability durability) {
+    this.commitSession = commitSession;
     this.mutations = mutations;
     this.durability = durability;
   }
@@ -39,11 +38,11 @@ public class TabletMutations {
   }
 
   public int getTid() {
-    return tid;
+    return commitSession.getLogId();
   }
 
   public long getSeq() {
-    return seq;
+    return commitSession.getWALogSeq();
   }
 
   public Durability getDurability() {
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/log/DfsLogger.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/log/DfsLogger.java
@@ -21,6 +21,7 @@ import static org.apache.accumulo.tserver.logger.LogEvents.COMPACTION_FINISH;
 import static org.apache.accumulo.tserver.logger.LogEvents.COMPACTION_START;
 import static org.apache.accumulo.tserver.logger.LogEvents.DEFINE_TABLET;
 import static org.apache.accumulo.tserver.logger.LogEvents.MANY_MUTATIONS;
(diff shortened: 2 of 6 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -365,7 +366,8 @@ private void write(final Collection<CommitSession> sessions, boolean mincFinish,
             if (commitSession.beginUpdatingLogsUsed(copy, mincFinish)) {
               try {
                 // Scribble out a tablet definition and then write to the metadata table
-                defineTablet(commitSession, writeRetry);
+                write(singletonList(commitSession), false,
+                    logger -> logger.defineTablet(commitSession), writeRetry);
```
keith-turner: Is the `NO_WAIT_LOGGER_OP` behavior preserved?  Not sure if its possible at this depth in the code, but it would be more clear if define tablet were written using `Durability.LOG`.  For define tablet we want to write to the log w/o sync because we know define tablet will be followed by subsequent writes that may or may not sync.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/DfsLogger.java
```diff
@@ -398,10 +401,12 @@ public static DFSLoggerInputStreams readHeaderAndReturnStream(FSDataInputStream
    */
   public synchronized void open(String address) throws IOException {
     String filename = UUID.randomUUID().toString();
-    log.debug("Address is {}", address);
+    if (log.isDebugEnabled())
```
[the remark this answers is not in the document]
milleruntime: I had added as an optimization but forgot the log4j parameterized statements make these checks obsolete.  I will remove.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletMutations.java
```diff
@@ -39,11 +38,11 @@ public TabletMutations(int tid, long seq, List<Mutation> mutations, Durability d
   }
 
   public int getTid() {
-    return tid;
+    return commitSession.getLogId();
```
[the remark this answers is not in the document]
milleruntime: Yes but not in Commit session.  I think it gets the value from Tablet.getLogId() which returns it's private final int logId.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
milleruntime: Yeah I see what you are saying but I am not sure how we could do that.

keith-turner: Ok.  Looking at that code I think we could have a `logFileData(LogFileKey key, LogFileValue value, Durability d)` and defineTablet could call that with a durability of `LOG`.  Also there are some other method that log a single key value that could call it.   Impl could be something like :

```java
logFileData(LogFileKey key, LogFileValue value, Durability d) {
   logFileData(singletonList(new Pair(k,v)), d);
}
```

This would avoid the duplicate exception handling in defineTablet.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -271,11 +271,13 @@ public void run() {
         while (!nextLogMaker.isShutdown()) {
           DfsLogger alog = null;
           try {
-            log.debug("Creating next WAL");
+            if (log.isDebugEnabled())
```
keith-turner: Seem like this check should be removed.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -271,11 +271,13 @@ public void run() {
...
+              log.debug("Creating next WAL");
             alog = new DfsLogger(tserver.getContext(), conf, syncCounter, flushCounter);
             alog.open(tserver.getClientAddressString());
             String fileName = alog.getFileName();
-            log.debug("Created next WAL " + fileName);
+            if (log.isDebugEnabled())
```
keith-turner: This could also be removed

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
```diff
@@ -271,11 +271,13 @@ public void run() {
...
+              log.debug("Created next WAL " + fileName);
```
keith-turner: could do `log.debug("Created next WAL {}", fileName);` to avoid string concat when debug not enabled.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/DfsLogger.java
```diff
@@ -537,22 +538,14 @@ public void close() throws IOException {
       }
   }
 
-  public synchronized void defineTablet(long seq, int tid, KeyExtent tablet) throws IOException {
+  public LoggerOperation defineTablet(CommitSession cs) throws IOException {
```
keith-turner: I like this change to only pass CommitSession

server/tserver/src/main/java/org/apache/accumulo/tserver/log/TabletServerLogger.java
milleruntime: Yeah thanks.  I missed the slf4j in this class.

PR #843, opened 2018-12-19
Remove extra loop in InMemoryMap

milleruntime:
* The number of total keys across mutations was counted twice so just
pass the count down to the InMemoryMap


milleruntime:
> This is neat. How did you find this?

The commit method was one that was showing up a lot using flight recorder so this lead me to dive down to calls in it and noticed that value was counted twice.

PR #845, opened 2018-12-19
fixes #823 unreference closed WALs that were never written to

### LINKED ISSUE #823: Handle empty write ahead logs more gracefully.
For certain error situations a tablet server may get into a retry loop and create lots of empty write ahead logs.  An empty write ahead log may have a header and no data, so its not actually an empty file.  All of these empty write ahead logs can cause slow recovery.  There are at least two possible way to handle this problem.  The tablet server could not add empty write ahead logs to zookeeper OR clean up known empty write ahead logs from zookeeper.  If that is not a workable solution, then maybe empty write ahead logs could be filtered during recovery so that they are not assigned to tablets.

keith-turner:


keith-turner:
All ITs passed for this PR.

PR #857, opened 2018-12-21
Updates to DumpZookeeper utility

mikewalch:
* Created human readable option which is now default
* Xml can still be output by passing --xml argument
* Utility can now be called using 'accumulo-util dumpzoo'


assemble/bin/accumulo-util
```diff
@@ -263,6 +264,9 @@ function main() {
     build-native)
       build_native "${@:2}"
       ;;
+    dumpzoo)
```
milleruntime: Maybe make it zoodump or zoo-dump similar to how we have hadoop-jar.  Although, our commands aren't that consistently named so its no big deal, either way.

mikewalch: I agree. It should probably have a `-` for consistency

PR #858, opened 2018-12-21
Create builder to clean up CachableBlockFile

milleruntime:
I always found the constructors for the CachableBlockFile.Reader difficult to track down what was being set.  I think this makes it a lot more clear, without adding too much code.

### DIFF
--- a/core/src/test/java/org/apache/accumulo/core/file/rfile/RFileTest.java
+++ b/core/src/test/java/org/apache/accumulo/core/file/rfile/RFileTest.java
@@ -70,7 +70,7 @@ import org.apache.accumulo.core.file.blockfile.cache.impl.BlockCacheConfiguratio
 import org.apache.accumulo.core.file.blockfile.cache.impl.BlockCacheManagerFactory;
 import org.apache.accumulo.core.file.blockfile.cache.lru.LruBlockCache;
 import org.apache.accumulo.core.file.blockfile.cache.lru.LruBlockCacheManager;
-import org.apache.accumulo.core.file.blockfile.impl.CachableBlockFile;
+import org.apache.accumulo.core.file.blockfile.impl.CachableBlockFile.CachableBuilder;
 import org.apache.accumulo.core.file.rfile.RFile.Reader;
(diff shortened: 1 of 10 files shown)


core/src/test/java/org/apache/accumulo/core/file/rfile/RFileTest.java
```diff
@@ -304,10 +304,10 @@ public void openReader(boolean cfsi) throws IOException {
       LruBlockCache indexCache = (LruBlockCache) manager.getBlockCache(CacheType.INDEX);
       LruBlockCache dataCache = (LruBlockCache) manager.getBlockCache(CacheType.DATA);
 
-      CachableBlockFile.Reader _cbr = new CachableBlockFile.Reader("source-1", in, fileLength, conf,
-          dataCache, indexCache,
-          CryptoServiceFactory.newInstance(accumuloConfiguration, ClassloaderType.JAVA));
-      reader = new RFile.Reader(_cbr);
+      CachableBuilder cb = new CachableBuilder().cacheId("source-1").input(in).length(fileLength)
```
mikewalch: It looks like you are always creating a reader.  Could you have your builder just create one (instead of CachableBuilder) ? Like the following:

```java
      reader = RFile.reader().cacheId("source-1").input(in).length(fileLength)
          .conf(conf).data(dataCache).index(indexCache).cryptoService(
              CryptoServiceFactory.newInstance(accumuloConfiguration, ClassloaderType.JAVA));
```

milleruntime: Yeah I should be able to do that.  There are a few places where the CachableBlockFile reader are used without the RFile but they might just be for tests.

milleruntime: So I figured out that Summaries use just the BCFile reader and not the entire RFile.  I think this was an optimization done for reading summary data.  So for now, we still need it separate from RFile.

mikewalch: Ok sounds good

PR #860, opened 2018-12-27
Fixes #854 handle many tablets referencing many WALs

### LINKED ISSUE #854: Tablet server should compact all tablets referencing older logs
The tablet server minor compacts any tablet that reference more than 5 write ahead logs.  With per tablet server write ahead log tracking all write ahead logs referenced by any tablet are assigned to all tablets at recovery time.  This means that individual tablets may end up trying to recover from man write ahead logs.

When a tablet server has more than 5 write ahead logs then it should probably minor compact/flush any tablet referencing the WALs created before the 5th one.  These tablets themselves may have less than 5 write ahead logs.

keith-turner:


test/src/main/java/org/apache/accumulo/test/functional/WALSunnyDayIT.java
```diff
@@ -258,17 +258,27 @@ private int getWaitFactor() {
...
+        ZooReaderWriter zk = new ZooReaderWriter(i.getZooKeepers(), i.getZooKeepersSessionTimeOut(),
+            "");
+        WalStateManager wals = new WalStateManager(c.getInstance(), zk);
+        for (Entry<Path,WalState> entry : wals.getAllState().entrySet()) {
+          // WALs are in use if they are not unreferenced
+          result.put(entry.getKey().toString(), entry.getValue());
+        }
+        return result;
+      } catch (WalMarkerException wme) {
+        if (wme.getCause() instanceof NoNodeException) {
```
keith-turner: TODO this should probably be handled in WalStateManager.. need to open an issue.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
```diff
@@ -2497,6 +2497,38 @@ public void removeInUseLogs(Set<DfsLogger> candidates) {
...
   }
 
+  public void checkIfMinorCompactionNeededForLogs(List<DfsLogger> closedLogs) {
+
+    // grab this outside of tablet lock.
+    int maxLogs = tableConfiguration.getCount(Property.TABLE_MINC_LOGS_MAX);
+
+    String reason = null;
+    synchronized (this) {
+      if (currentLogs.size() >= maxLogs) {
```
jschmidt10: Will this effectively subtract 1 from the max logs configuration? If currentLogs.size() is equal to maxLogs, then we'll force a MinC.

test/src/main/java/org/apache/accumulo/test/functional/WALSunnyDayIT.java
keith-turner: I determined this should probably not be handled in WalStateManager.  However while looking into the issue I discovered #866

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
keith-turner: > If currentLogs.size() is equal to maxLogs, then we'll force a MinC.

Thats correct.  I believe this was the behavior of the code before my change.  The tablet server code used to do the following.  For that case I was trying to maintain the current behavior.

```java
if (tablet.getLogCount() >= maxLogEntriesPerTablet) {
```

jschmidt10: Ahh, okay. Thanks for the clarification.

PR #861, opened 2018-12-27
Remove unused param from Summaries

milleruntime:
* Also replace duplicate code in SummaryReader with method
* Inline print method only called by other print method


core/src/main/java/org/apache/accumulo/core/summary/SummaryReader.java
```diff
@@ -197,19 +194,11 @@ public static SummaryReader load(FileSystem fs, Configuration conf, AccumuloConf
       bcReader = new CachableBlockFile.Reader(cb);
       return load(bcReader, summarySelector, factory);
     } catch (FileNotFoundException fne) {
-      SummaryReader sr = new SummaryReader();
```
keith-turner: Nice

PR #866, opened 2019-01-02
Fix race condition when getting WALs for dead tserver

keith-turner:
When a tablet server dies the master gets its WALs from ZK.  In ZK there
is a list of WALs per tserver. Each WAL in ZK has state that is either
OPEN, CLOSED, or UNREFERENCED.  The master needs a list of OPEN and
CLOSED logs for dead tservers.  While the master is trying to obtain
this list its possible that the Accumulo GC may delete an UNREFERENCED
WAL.  If this happened then the code before this commit would return an
empty list of WALs. This could result in OPEN and CLOSED logs being
ignored for recovery which could result in data loss.  This patch fixes
the race condition.

This bug was observed while looking into an exception I noticed while
writing test for #860.  In the IT I was frequently calling another
WalStateManager function and noticed NoNodeExceptions.  As a result I
examined how the entire class handled NoNode race conditions and found
this bug.  I noticed some other possible NoNode race condition, but do
not think this these would occur with the current way the code is called.

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/log/WalStateManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/log/WalStateManager.java
@@ -163,9 +163,22 @@ public class WalStateManager {
       String zpath = root() + "/" + tsi.toString();
       zoo.sync(zpath);
       for (String child : zoo.getChildren(zpath)) {
-        Pair<WalState,Path> parts = parse(zoo.getData(zpath + "/" + child, null));
-        if (parts.getFirst() != WalState.UNREFERENCED) {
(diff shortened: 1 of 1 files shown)


keith-turner:
I have been looking into this and I think it could only happen if the GC pulled info about an unreferenced WAL into memory before a tserver died AND then deleted the unreferenced WAL after the tserver died (while the master is trying to get the list).  The GC seems to completely ignore WALs for any dead tservers that have tablets assigned to them.

keith-turner:
Looking at the history of WalStateManager it seems like this may have been a bug since 1.8.0.

PR #868, opened 2019-01-02
Make replication services start when configured

milleruntime:
* Only start the replication services in TabletServer if the required
properties are set at startup

I noticed tserver will always start multiple threads for replication, even if replication name is not configured.  As far as I can tell, this is the only required property for replication and if it is not set at startup anyway, there is a good chance not all data will be replicated.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/conf/Property.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/Property.java
@@ -456,8 +456,6 @@ public enum Property {
       "The amount of time an assignment can run before the server will print a"
           + " warning along with the current stack trace. Meant to help debug stuck"
           + " assignments"),
-  TSERV_REPLICATION_ENABLED("tserver.replication.services.enabled", "false", PropertyType.BOOLEAN,
-      "Start the replication services when tserver starts up."),
   TSERV_REPLICATION_REPLAYERS("tserver.replication.replayer.", null, PropertyType.PREFIX,
       "Allows configuration of implementation used to apply replicated data"),
   TSERV_REPLICATION_DEFAULT_HANDLER("tserver.replication.default.replayer",
@@ -1190,7 +1188,7 @@ public enum Property {
 
   private static final EnumSet<Property> fixedProperties = EnumSet.of(Property.TSERV_CLIENTPORT,
       Property.TSERV_NATIVEMAP_ENABLED, Property.TSERV_SCAN_MAX_OPENFILES,
-      Property.TSERV_REPLICATION_ENABLED, Property.MASTER_CLIENTPORT, Property.GC_PORT);
+      Property.MASTER_CLIENTPORT, Property.GC_PORT);
 
   /**
    * Checks if the given property may be changed via Zookeeper, but not recognized until the restart
--- a/core/src/main/java/org/apache/accumulo/core/conf/Property.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/Property.java
@@ -456,6 +456,8 @@ public enum Property {
       "The amount of time an assignment can run before the server will print a"
           + " warning along with the current stack trace. Meant to help debug stuck"
           + " assignments"),
+  TSERV_REPLICATION_ENABLED("tserver.replication.services.enabled", "false", PropertyType.BOOLEAN,
+      "Start the replication services when tserver starts up."),
   TSERV_REPLICATION_REPLAYERS("tserver.replication.replayer.", null, PropertyType.PREFIX,
       "Allows configuration of implementation used to apply replicated data"),
   TSERV_REPLICATION_DEFAULT_HANDLER("tserver.replication.default.replayer",
@@ -1188,7 +1190,7 @@ public enum Property {
 
   private static final EnumSet<Property> fixedProperties = EnumSet.of(Property.TSERV_CLIENTPORT,
       Property.TSERV_NATIVEMAP_ENABLED, Property.TSERV_SCAN_MAX_OPENFILES,
-      Property.MASTER_CLIENTPORT, Property.GC_PORT);
+      Property.TSERV_REPLICATION_ENABLED, Property.MASTER_CLIENTPORT, Property.GC_PORT);
 
   /**
    * Checks if the given property may be changed via Zookeeper, but not recognized until the restart
--- a/server/master/src/main/java/org/apache/accumulo/master/Master.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/Master.java
@@ -20,6 +20,7 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.apache.accumulo.fate.util.UtilWaitThread.sleepUninterruptibly;
 
 import java.io.IOException;
+import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -1261,22 +1262,6 @@ public class Master
     clientService = sa.server;
     log.info("Started Master client service at {}", sa.address);
 
-    // Start the replication coordinator which assigns tservers to service replication requests
-    MasterReplicationCoordinator impl = new MasterReplicationCoordinator(this);
-    ReplicationCoordinator.Iface haReplicationProxy = HighlyAvailableServiceWrapper.service(impl,
-        this);
-    // @formatter:off
-    ReplicationCoordinator.Processor<ReplicationCoordinator.Iface> replicationCoordinatorProcessor =
-      new ReplicationCoordinator.Processor<>(TraceWrap.service(haReplicationProxy));
-    // @formatter:on
-    ServerAddress replAddress = TServerUtils.startServer(context, hostname,
-        Property.MASTER_REPLICATION_COORDINATOR_PORT, replicationCoordinatorProcessor,
-        "Master Replication Coordinator", "Replication Coordinator", null,
-        Property.MASTER_REPLICATION_COORDINATOR_MINTHREADS,
-        Property.MASTER_REPLICATION_COORDINATOR_THREADCHECK, Property.GENERAL_MAX_MESSAGE_SIZE);
-
-    log.info("Started replication coordinator service at " + replAddress.address);
-
     // block until we can obtain the ZK lock for the master
     getMasterLock(zroot + Constants.ZMASTER_LOCK);
 
@@ -1391,6 +1376,81 @@ public class Master
       sleepUninterruptibly(100, TimeUnit.MILLISECONDS);
     }
 
+    // if the replication name is ever set, then start replication services
+    final ReplTServer replServer = new ReplTServer();
+    SimpleTimer.getInstance(getConfiguration()).schedule(() -> {
+      try {
+        if (!getConfiguration().get(Property.REPLICATION_NAME).isEmpty()) {
+          if (replServer.server == null) {
+            log.info(Property.REPLICATION_NAME.getKey() + " was set, starting repl services.");
+            replServer.setServer(setupReplication());
+          }
+        }
+      } catch (Exception e) {
+        log.error("Replication name was set but error occurred starting services. ", e);
+      }
+    }, 1000, 5000);
+
+    // The master is fully initialized. Clients are allowed to connect now.
+    masterInitialized.set(true);
+
+    while (clientService.isServing()) {
+      sleepUninterruptibly(500, TimeUnit.MILLISECONDS);
+    }
+    log.info("Shutting down fate.");
+    fate.shutdown();
+
+    log.info("Shutting down timekeeping.");
+    timeKeeper.shutdown();
+
+    final long deadline = System.currentTimeMillis() + MAX_CLEANUP_WAIT_TIME;
+    statusThread.join(remaining(deadline));
+    if (!getConfiguration().get(Property.REPLICATION_NAME).isEmpty()) {
+      if (replicationWorkAssigner != null)
+        replicationWorkAssigner.join(remaining(deadline));
+      if (replicationWorkDriver != null)
+        replicationWorkDriver.join(remaining(deadline));
+      TServerUtils.stopTServer(replServer.server);
+    }
+    // Signal that we want it to stop, and wait for it to do so.
+    if (authenticationTokenKeyManager != null) {
+      authenticationTokenKeyManager.gracefulStop();
+      authenticationTokenKeyManager.join(remaining(deadline));
+    }
+
+    // quit, even if the tablet servers somehow jam up and the watchers
+    // don't stop
+    for (TabletGroupWatcher watcher : watchers) {
+      watcher.join(remaining(deadline));
+    }
+    log.info("exiting");
+  }
+
+  class ReplTServer {
+    TServer server;
+
+    public void setServer(TServer server) {
+      this.server = server;
+    }
+  }
+
+  private TServer setupReplication()
+      throws UnknownHostException, KeeperException, InterruptedException {
+    // Start the replication coordinator which assigns tservers to service replication requests
+    MasterReplicationCoordinator impl = new MasterReplicationCoordinator(this);
+    ReplicationCoordinator.Iface haReplicationProxy = HighlyAvailableServiceWrapper.service(impl,
+        this);
+    // @formatter:off
+    ReplicationCoordinator.Processor<ReplicationCoordinator.Iface> replicationCoordinatorProcessor =
+            new ReplicationCoordinator.Processor<>(TraceWrap.service(haReplicationProxy));
+    // @formatter:on
+    ServerAddress replAddress = TServerUtils.startServer(context, hostname,
+        Property.MASTER_REPLICATION_COORDINATOR_PORT, replicationCoordinatorProcessor,
+        "Master Replication Coordinator", "Replication Coordinator", null,
+        Property.MASTER_REPLICATION_COORDINATOR_MINTHREADS,
+        Property.MASTER_REPLICATION_COORDINATOR_THREADCHECK, Property.GENERAL_MAX_MESSAGE_SIZE);
+
+    log.info("Started replication coordinator service at " + replAddress.address);
     // Start the daemon to scan the replication table and make units of work
     replicationWorkDriver = new ReplicationDriver(this);
     replicationWorkDriver.start();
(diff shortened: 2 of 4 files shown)


joshelser:
Pardon my ignorance: do we differentiate properties that can be changed dynamically and those which require a restart? That's my biggest concern -- I think your change would require a config-change+restart instead of just at config-change that we have now.

ctubbsii:
> Pardon my ignorance: do we differentiate properties that can be changed dynamically and those which require a restart?

In the past, I don't think we've done a good job at this differentiation. I think there's a natural differentiation, which we could use: properties which can be set in ZooKeeper, and those that can't, but even those that can be set in ZooKeeper, there's not well-defined behavior as to when they will go into effect. This is something that we should think about doing better in future. That's outside the scope of this issue, in the general case, but in this narrow case, I think Josh makes a good point: if you can change the configuration for a replication receiver on a running system, we should probably make sure that the services respond to that configuration change... and we can't do that if they are disabled on startup and don't monitor for configuration changes to re-enable.

joshelser:
> if you can change the configuration for a replication receiver on a running system, we should probably make sure that the services respond to that configuration change... and we can't do that if they are disabled on startup and don't monitor for configuration changes to re-enable.

Yes, this was half of my point. The other half is that we need to have documentation that would tell people that they need to restart Accumulo after making this change, especially since this changes previous semantics. Probably release notes to advertise the immediate change and user manual to tell people for the future that they need to do it.

ctubbsii:
It looks like there's a "fixed Zoo property" concept, which we label on some properties, which indicates things that can be changed in ZK, but won't be recognized until restart of some service/daemon... not sure if that is used in the generated docs for properties... but it seems like it easily could be.

milleruntime:
Yeah a fixed property will get a little snippet in the generated doc like `gc.port.client` does here:
"zk mutable: yes but requires restart of the gc"

milleruntime:
With this change, we could throw an illegal state exception if the user sets a property on the fly, expecting data to get replicated.  I am not sure where that would be done though.

mikewalch:
It's easy to make `replication.name` a [fixed property](https://github.com/apache/accumulo/blob/90dfad3b341a5098113ea8fdce28a0bc4511d067/core/src/main/java/org/apache/accumulo/core/conf/Property.java#L1189) but it looks like [this method](https://github.com/apache/accumulo/blob/ce1aa1f76577caed521f9668147f1ca81239b137/core/src/main/java/org/apache/accumulo/core/conf/ConfigurationDocGen.java#L135) will need to updated for the docs to say Accumulo needs to be restarted (as right now it will say replication needs to be restarted.)

milleruntime:
> It's easy to make replication.name a fixed property but it looks like this method will need to updated for the docs to say Accumulo needs to be restarted (as right now it will say replication needs to be restarted.)

This makes me wonder if the REPLICATION prefix properties should all be TSERVER.  The names would be longer but seems more accurate to me.

ctubbsii:
> This makes me wonder if the REPLICATION prefix properties should all be TSERVER. The names would be longer but seems more accurate to me.

If it affects the remote replication receiver service built in to the tserver to act as a destination for replicas, then probably.

mikewalch:
> This makes me wonder if the REPLICATION prefix properties should all be TSERVER. The names would be longer but seems more accurate to me.

I am not too familiar with replication but it's properties might be used in tserver and master.  I think the documentation should not call out the process to be restarted by using the prefix of the property.  Either we should make sure that this prefix is an Accumulo process (i.e master, tserver, monitor, etc) or just have a general message in the docs like `zk mutable: yes but requires a restart`. Users can determine which processes should be restarted.

joshelser:
> I am not too familiar with replication but it's properties might be used in tserver and master.

That's correct. There's logic across master and tabletserver for replication. The intent to have its own prefix was to keep all replication configuration in "one place".

> just have a general message in the docs like zk mutable: yes but requires a restart

+1

> With this change, we could throw an illegal state exception if the user sets a property on the fly, expecting data to get replicated. I am not sure where that would be done though.

That would be client side validation, right? I suppose it's not unreasonable for this case to let client-side validation be sufficient (e.g. it's not bad if a user circumvents our own code and tries to set a zk-mutable property -- worst case, it's just ineffectual). That sounds like a good idea. Even if not an exception, a WARN would go a long way for operators to realize they straying from the "norm"

ctubbsii:
We can certainly fix the message without fixing the property names, but if it a configuration knob for tserver, then it should be prefixed accordingly. The convention we have is that properties for specific processes are prefixed with the process type, if they affect all of Accumulo, they are prefixed with either `general.` or `instance.` (the latter being used for instance-identifying information... information that together establishes the an identity for a cluster and must be the same across the entire cluster in order for each component to identify as part of the same Accumulo instance).

ctubbsii:
> Let's do one thing at a time, please.

Agreed. My thinking is that there are multiple possible follow-on actions which may affect the decision about the initial action. Have we reached consensus on the initial set of actions here? If so, is it the following?

1. enable/disable replication based on startup config,
2. used fixed properties to document that changes in ZK require restart, and
3. clean up wording of said documentation so it's not based on property name

Or something else?

milleruntime:
> > Let's do one thing at a time, please.
> 
> Agreed. My thinking is that there are multiple possible follow-on actions which may affect the decision about the initial action. Have we reached consensus on the initial set of actions here? If so, is it the following?
> 
> 1. enable/disable replication based on startup config,
> 2. used fixed properties to document that changes in ZK require restart, and
> 3. clean up wording of said documentation so it's not based on property name
> 
> Or something else?

I also suggest throwing an error on the client if replication properties are set (through the shell I guess) when the tserver doesn't have a replication.name.  Perhaps this could be done in ConfigSanityCheck and print a warning?

ctubbsii:
It's not clear to me how all the replication properties work together... which ones are on the send side and which ones affect the receive side.

It might be best to just add a new fixed property (requires restart), `tserver.replicationservices.enabled=true/false` and leave all the replication property stuff for future cleanup/improvements.

milleruntime:
> It's not clear to me how all the replication properties work together... which ones are on the send side and which ones affect the receive side.
> 
> It might be best to just add a new fixed property (requires restart), `tserver.replicationservices.enabled=true/false` and leave all the replication property stuff for future cleanup/improvements.

+1  I like this the best.  This also won't require changing the documentation for the other properties since it would get generated to say "zk mutable: yes but requires restart of the tserver".

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -456,6 +456,8 @@
       "The amount of time an assignment can run before the server will print a"
           + " warning along with the current stack trace. Meant to help debug stuck"
           + " assignments"),
+  TSERV_REPLICATION_ENABLED("tserver.replication.services.enabled", "false", PropertyType.BOOLEAN,
```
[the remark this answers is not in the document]
milleruntime: Look at the services that the Master starts up for replication, it is a bit trickier.  And I am not sure how much we save since there should only be one Master running versus X TabletServers.  I think making it a tserver property makes the most sense here.  There may also be a use case where you want certain tservers to participate in replication and disabled on others.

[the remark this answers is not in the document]
mikewalch: Even if you are not going to touch replication in other processes, I would keep the name `replication.enabled`.  It's either enabled or it's not. If can confusing if we have other similar properties like `master.replication.enabled` as you don't want users enabling it in tserver but not the master.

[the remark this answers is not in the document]
joshelser: >  It's either enabled or it's not

This.

Either replication is on (and all the necessary services are running) or it's off (and none of the services are running). I thought you were advocating to moving 100% of the way to the latter with this change, @milleruntime.

keith-turner:
Could do the following :  
 * If replication.name is set to empty string then don't start up replication related services in the tserver.
 * In the tserver create a timer that checks if replication.name transitions from empty to non-empty then automatically start the services in the tserver.

This change should hopefully be transparent to the user.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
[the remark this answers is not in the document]
milleruntime: Ideally, yes.

milleruntime:
> Could do the following :
> 
> * If replication.name is set to empty string then don't start up replication related services in the tserver.
> * In the tserver create a timer that checks if replication.name transitions from empty to non-empty then automatically start the services in the tserver.  This is similar to how there are timers that check if thread pool sizes need to be adjusted.
> 
> This change should hopefully be transparent to the user.

Ok I think this would work the best.  And also, the more I thought about it, it doesn't make sense to have a new property (or have a tserver property) so I will go back to how I was checking initially on replication.name.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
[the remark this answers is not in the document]
ctubbsii: As I understand it, the current design doesn't make a single `replication.enabled` option as simple to do as it first appears. The elements in the master that are replication related appear to be part of the replica "send" code path.... the elements in the tserver that we're talking about appear related to the "AccumuloReplicaSystem", which is specifically a tserver feature for handling received replication data at the destination (the replica "receive" code path). It should be entirely possible to enable these features independently (to support basically any use case that isn't bi-directional replication between two Accumulo instances).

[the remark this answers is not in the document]
joshelser: > the current design doesn't make a single replication.enabled option as simple to do as it first appears.

I'm not sure what you think is untenable here. There are multiple services that are started, in different processes, but it all uses the single configuration option: a global on/off.

> It should be entirely possible to enable these features independently (to support basically any use case that isn't bi-directional replication between two Accumulo instances).

The current implementation is definitely not intended to be used in this fashion. I'd caution against trying to separate these as it's going to open a can of worms. If nothing else, if this is an itch you want to scratch, do that on its own to avoid conflating these issues.

milleruntime:
OK I think these changes work best for all use cases and require no action from users.  I made a thread in each Master and tserver that checks every 5 seconds if replication.name is set.  Then if it is set, startup all the services.  That seemed to be a reasonable time to check since any longer, and the tests began failing.  

The only other part in question was whether to kill the threads after starting the services.  I didn't see a graceful way to do that with SimpleTimer though.

I kicked off a jenkins job to run all ITs: https://jenkins.revelc.net/job/Accumulo-PR/127/

milleruntime:
MultiTserverReplicationIT was failing due to this change (it assumed services were always running) but was easily fixed in 01f89e2.  @joshelser were your concerns resolved with my latest changes?

joshelser:
Taking a look at the latest you have pushed now, @milleruntime .

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1391,27 +1376,20 @@ public void run() {
...
+        if (!getConfiguration().get(Property.REPLICATION_NAME).isEmpty()) {
+          if (replServer.server == null) {
+            log.info(Property.REPLICATION_NAME.getKey() + " was set, starting repl services.");
+            replServer.setServer(setupReplication());
+          }
+        }
+      } catch (UnknownHostException | KeeperException | InterruptedException e) {
+        log.error("Error occurred starting replication services. ", e);
+      }
+    }, 1000, 5000);
```
joshelser: Change we make this initial delay `0` to mimic the current semantics (we start the service right away)?

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1427,9 +1405,13 @@ public void run() {
 
     final long deadline = System.currentTimeMillis() + MAX_CLEANUP_WAIT_TIME;
     statusThread.join(remaining(deadline));
-    replicationWorkAssigner.join(remaining(deadline));
-    replicationWorkDriver.join(remaining(deadline));
-    replAddress.server.stop();
+    if (!getConfiguration().get(Property.REPLICATION_NAME).isEmpty()) {
```
joshelser: What about always checking if these are non-null and stopping them if they are?

Checking the configuration for the property (while it should always be accurate) has the potential to introduce a bug where we could just check everything and be certain it's all stopped.

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1444,6 +1426,55 @@ public void run() {
     log.info("exiting");
   }
 
+  class ReplTServer {
```
joshelser: What value is this class giving us? Just seems like an extra object we create.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -2841,35 +2839,14 @@ public void run() {
...
-        }
+    // if the replication name is ever set, then start replication services
+    SimpleTimer.getInstance(aconf).schedule(() -> {
+      if (!getConfiguration().get(Property.REPLICATION_NAME).isEmpty()) {
+        log.info(Property.REPLICATION_NAME.getKey() + " was set, starting repl services.");
+        setupReplication(aconf);
       }
-    };
-    SimpleTimer.getInstance(aconf).schedule(replicationWorkThreadPoolResizer, 10000, 30000);
+    }, 1000, 5000);
```
joshelser: `0` initial delay here too, please.

test/src/main/java/org/apache/accumulo/test/replication/MultiTserverReplicationIT.java
```diff
@@ -47,6 +48,7 @@
 
   @Override
   public void configure(MiniAccumuloConfigImpl cfg, Configuration hadoopCoreSite) {
+    cfg.setProperty(Property.REPLICATION_NAME.getKey(), "test");
```
joshelser: Can you add a comment here as to why this is needed, please?

server/master/src/main/java/org/apache/accumulo/master/Master.java
milleruntime: This was so we can pass a final object to the timer thread.

joshelser: What about just using `AtomicReference` instead of defining our own class?

server/master/src/main/java/org/apache/accumulo/master/Master.java
milleruntime: Ahhh I see.  Yes, that would be better.  Good idea.

server/master/src/main/java/org/apache/accumulo/master/Master.java
milleruntime: Oh yeah good idea, that would be better.

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1427,9 +1406,12 @@ public void run() {
...
     final long deadline = System.currentTimeMillis() + MAX_CLEANUP_WAIT_TIME;
     statusThread.join(remaining(deadline));
-    replicationWorkAssigner.join(remaining(deadline));
-    replicationWorkDriver.join(remaining(deadline));
-    replAddress.server.stop();
+    if (replicationWorkAssigner != null)
+      replicationWorkAssigner.join(remaining(deadline));
+    if (replicationWorkDriver != null)
+      replicationWorkDriver.join(remaining(deadline));
+    TServerUtils.stopTServer(replServer.get());
```
keith-turner: Is it ok to pass stopTServer null?

keith-turner: could do 

```java
if(replServer.get() != null){
     replicationWorkAssigner.join(remaining(deadline));		￼   
￼     replicationWorkDriver.join(remaining(deadline));		￼ 
￼     replAddress.server.stop();
}

```

server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1391,27 +1377,20 @@ public void run() {
...
-    try {
-      replicationMetrics.register();
-    } catch (Exception e) {
-      log.error("Failed to register replication metrics", e);
-    }
+    // if the replication name is ever set, then start replication services
+    final AtomicReference<TServer> replServer = new AtomicReference<>();
+    SimpleTimer.getInstance(getConfiguration()).schedule(() -> {
+      try {
+        if (!getConfiguration().get(Property.REPLICATION_NAME).isEmpty()) {
```
keith-turner: Could check if `replServer.get() == null` before checking config as this may be a much faster check and the timer has less to do after enabling.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -2952,8 +2929,10 @@ public void run() {
         }
       }
     }
-    log.debug("Stopping Replication Server");
-    TServerUtils.stopTServer(this.replServer);
+    if (this.replServer != null) {
```
keith-turner: If `replServer` is not volatile then may not see changes from another thread.

server/master/src/main/java/org/apache/accumulo/master/Master.java
milleruntime: Yeah the TServerUtils method checks for null initially. I had something similar earlier but Josh suggest to always stop the thread.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
milleruntime: Fixed in 0016dce1a80f7c8f8552252f148ba77f94da2725

server/master/src/main/java/org/apache/accumulo/master/Master.java
milleruntime: Fixed in 0016dce1a80f7c8f8552252f148ba77f94da2725.  Also did the same in Tserver

joshelser: Only concern with how this is written now is that it's subject to a race condition if we ever had multiple invocations of this "block" being executed concurrently (we could start two replication Thrift servers). I think `SimpleTimer` does not allow this to happen though.

joshelser:
@milleruntime one comment up above on SimpleTimer (if you could double-check that?), then I'm good. Thanks for the review ping.

server/master/src/main/java/org/apache/accumulo/master/Master.java
milleruntime: The race condition exists prior to this update, when we were just checking if the name was empty, because there is no synchronization.  But I think we are OK because the ScheduleExecutorService should prevent two concurrent threads from ever running for the same task.  The javadoc states:
"If any execution of this task
     * takes longer than its period, then subsequent executions
     * may start late, but will not concurrently execute."

milleruntime:
> one comment up above on SimpleTimer (if you could double-check that?), then I'm good. 

I looked into it a bit and replied to your comment.  I think we are OK.

milleruntime:
All ITs in accumulo/test/src/main/java/org/apache/accumulo/test/replication are still passing

PR #871, opened 2019-01-04
Check transaction are alive for new bulk import RPCs

keith-turner:


server/base/src/main/java/org/apache/accumulo/server/zookeeper/TransactionWatcher.java
```diff
@@ -122,28 +122,37 @@ public boolean transactionComplete(String type, long tid) throws Exception {
     }
   }
 
+  public void runQuietly(String ztxBulk, long tid, Runnable task) {
```
milleruntime: Can you add a javadoc description of what this run method does?

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -136,6 +143,52 @@ void start(Path bulkDir, Master master, long tid, boolean setTime) throws Except
...
+            ThriftUtil.returnClient(client);
+          }
+        });
+
+        loadQueue.clear();
+        queuedDataSize = 0;
+      }
+    }
+
+    private void queueLoad(HostAndPort server, KeyExtent extent,
```
milleruntime: This is a confusing name, since we have different meanings for the term "load".  I assume you are "loading the queue" here but perhaps a name like addToQueue may be less confusing.

milleruntime:
I think these changes are good and definitely needed.  My only concern is the complexity of the LoadFiles FATE transaction.  I think at the very least we need another class to abstract some of the logic, similar to how we have OfflineLoader and OnlineLoader.  We could add something to the Loader abstract class for queueing up load messages.

Another idea would be to break up LoadFiles into multiple FATE steps.  We could have something like QueueFiles and the SendFiles.

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -128,6 +129,12 @@ void start(Path bulkDir, Master master, long tid, boolean setTime) throws Except
     // track how many tablets were sent load messages per tablet server
     MapCounter<HostAndPort> loadMsgs;
 
+    // Each RPC to a tablet server needs to check in zookeeper to see if the transaction is still
+    // active. The purpose of this map is to group load request by tablet servers inorder to do less
+    // RPCs. Less RPCs will result in less calls to Zookeeper.
+    Map<HostAndPort,Map<TKeyExtent,Map<String,MapFileInfo>>> loadQueue;
```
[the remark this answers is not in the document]
keith-turner: I don't find it confusing.  I think the type parameters make it very explicit.  Did you have a suggestion for a replacement?

keith-turner:
> We could add something to the Loader abstract class for queueing up load messages.

 I don't think pushing the logic up makes sense because OfflineLoader does not need this logic.  The queuing behavior is specific to the OnlineLoader.

> Another idea would be to break up LoadFiles into multiple FATE steps. We could have something like QueueFiles and the SendFiles.

Seems like this would require pushing another file to HDFS that is sorted differently, which I don't think would simplify things.

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
keith-turner: Changed  in f2d62d7

server/base/src/main/java/org/apache/accumulo/server/zookeeper/TransactionWatcher.java
keith-turner: Added in f2d62d7

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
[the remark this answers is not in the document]
milleruntime: Could create a new class but that seems like overkill since its only used in a few spots.

milleruntime:
I think the real problem comes with the complexity of the LoadFiles class itself and not these changes.  I can try and come up with some ideas how to make it easier to understand and maintain, separate from this PR.

PR #874, opened 2019-01-07
Fix MapReduce bug

mikewalch:
* AccumuloOutputFormat is using wrong getClientInfo method

### DIFF
--- a/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloOutputFormat.java
+++ b/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloOutputFormat.java
@@ -16,8 +16,6 @@
  */
 package org.apache.accumulo.hadoop.mapreduce;
 
-import static org.apache.accumulo.hadoopImpl.mapreduce.AbstractInputFormat.getClientInfo;
-
 import java.io.IOException;
 
 import org.apache.accumulo.core.client.Accumulo;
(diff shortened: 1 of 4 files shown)


hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/AccumuloOutputFormat.java
```diff
@@ -16,8 +16,6 @@
  */
 package org.apache.accumulo.hadoop.mapreduce;
 
-import static org.apache.accumulo.hadoopImpl.mapreduce.AbstractInputFormat.getClientInfo;
```
milleruntime: Oh yeah bad import haha.  Nice catch!

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoopImpl/mapreduce/AccumuloOutputFormatImpl.java
```diff
@@ -88,7 +88,7 @@ public static void setClientInfo(Job job, ClientInfo info) {
    *
    * @since 2.0.0
    */
-  protected static ClientInfo getClientInfo(JobContext context) {
+  public static ClientInfo getClientInfo(JobContext context) {
```
milleruntime: Should just drop this method and call the OutputConfigurator.

mikewalch: I am not sure why this should be done.  Do you want to create a separate issue?

milleruntime: Yeah I will open an issue.  There is just a lot of redundant code now in the new map reduce impl that we needed before because of so many methods being in the public API.

PR #881, opened 2019-01-08
During tablet recovery filter logs out that do not define tablet.

keith-turner:
In a situation where a tablet had lots of write ahead logs, log
recovery consumed a lot of memory.  The reason for the memory
consumption was that all recovery logs are opened at once.  This
means the index for each file and a few key/values from each file
are read into memory.

This patch modifies the code to first open logs one at time to see if
they define the tablet. Then only the logs that define the tablet are
opened at the same time.  This has the potential to use less memory
when a subset of the logs define a tablet.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogsIterator.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogsIterator.java
@@ -52,7 +52,15 @@ public class RecoveryLogsIterator implements CloseableIterator<Entry<LogFileKey,
 
     try {
       for (Path log : recoveryLogPaths) {
-        iterators.add(new RecoveryLogReader(fs, log, start, end));
+        LOG.debug("Opening recovery log {}", log.getName());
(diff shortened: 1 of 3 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogsIterator.java
```diff
@@ -52,7 +52,15 @@
 
     try {
       for (Path log : recoveryLogPaths) {
-        iterators.add(new RecoveryLogReader(fs, log, start, end));
+        LOG.debug("Opening recovery log {}", log.getName());
+        RecoveryLogReader rlr = new RecoveryLogReader(fs, log, start, end);
```
alerman: Should this reader be opened in a try-with-resources or closed in a finally block

keith-turner: No, because rlr is added to a list of readers that the class keeps later in the method.

alerman: got it, thanks

PR #884, opened 2019-01-09
Clean up new Map Reduce internals

milleruntime:
Separated into 3 commits.
1 - Clean up the Input Classes by making AbstractRecordReader into AccumuloRecordReader
2 - Clean up the Output Classes by making AbstractRecordWriter into AccumuloRecordWriter
3 - Clean up FileOutput classes

AccumuloRecordReader and AccumuloRecordWriter are still abstract but I thought the names would make what gets serialized and logged more user friendly.


milleruntime:
Few ITs are failing so I will have to push up some fixes.

milleruntime:
@mikewalch I made some changes in 7d0a967.  We had been passing a hard coded class to the serialization in the RecordReader and writer but that broke with this PR.  This next change just passes the calling class all the way down to the serialization and doesn't make any assumptions about who is calling getSplits() and initialize().

PR #890, opened 2019-01-10
Organize new Bulk Import code

milleruntime:
My attempt to make the new code easier to follow and maintain. 

* Move bulk import clientImpl code to its own bulk package
* Pull some FaTE logic out of BulkImport
* Made the LoadMappingIterator its own class and push some of the reader logic into it
* Added comments to BulkSerialize
* Removed some VisibileForTesting annotations from ConcurrentKeyExtentCache where its used in BulkImport
* Pulled some logic out of LoadFiles.loadFiles and created some new methods for readability

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/LoadMappingIterator.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/LoadMappingIterator.java
@@ -70,8 +70,6 @@ public class LoadMappingIterator
 
   @Override
   public Map.Entry<KeyExtent,Bulk.Files> next() {
-    if (!hasNext())
-      return null;
     Bulk.Mapping bm = gson.fromJson(reader, Bulk.Mapping.class);
     if (renameMap != null) {
       return new AbstractMap.SimpleEntry<>(bm.getKeyExtent(tableId),
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/LoadMappingIterator.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/LoadMappingIterator.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.accumulo.core.clientImpl.bulk;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.apache.accumulo.core.clientImpl.bulk.BulkSerialize.createGson;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.UncheckedIOException;
+import java.util.AbstractMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import org.apache.accumulo.core.clientImpl.Table;
+import org.apache.accumulo.core.dataImpl.KeyExtent;
+
+import com.google.gson.Gson;
+import com.google.gson.stream.JsonReader;
+
+/**
+ * Iterator for reading the Bulk Load Mapping JSON.
+ */
+public class LoadMappingIterator
+    implements Iterator<Map.Entry<KeyExtent,Bulk.Files>>, AutoCloseable {
+  private Table.ID tableId;
+  private JsonReader reader;
+  private Gson gson = createGson();
+  private Map<String,String> renameMap;
+
+  LoadMappingIterator(Table.ID tableId, InputStream loadMapFile) throws IOException {
+    this.tableId = tableId;
+    this.reader = new JsonReader(new BufferedReader(new InputStreamReader(loadMapFile, UTF_8)));
+    this.reader.beginArray();
+  }
+
+  public void setRenameMap(Map<String,String> renameMap) {
+    this.renameMap = renameMap;
+  }
+
+  @Override
+  public void close() throws IOException {
+    reader.close();
+  }
+
+  @Override
+  public boolean hasNext() {
+    try {
+      return reader.hasNext();
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @Override
+  public Map.Entry<KeyExtent,Bulk.Files> next() {
+    if (!hasNext())
+      return null;
+    Bulk.Mapping bm = gson.fromJson(reader, Bulk.Mapping.class);
+    if (renameMap != null) {
+      return new AbstractMap.SimpleEntry<>(bm.getKeyExtent(tableId),
+          bm.getFiles().mapNames(renameMap));
+    } else {
+      return new AbstractMap.SimpleEntry<>(bm.getKeyExtent(tableId), bm.getFiles());
+    }
+  }
+
+}
--- a/server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
@@ -312,7 +312,7 @@ class LoadFiles extends MasterRepo {
    */
   private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator loadMapIter,
       Master master, long tid) throws Exception {
-    PeekingIterator<Map.Entry<KeyExtent,Bulk.Files>> lmi = new PeekingIterator(loadMapIter);
+    PeekingIterator<Map.Entry<KeyExtent,Bulk.Files>> lmi = new PeekingIterator<>(loadMapIter);
     Map.Entry<KeyExtent,Bulk.Files> loadMapEntry = lmi.peek();
 
     Text startRow = loadMapEntry.getKey().getPrevEndRow();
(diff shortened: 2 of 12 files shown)


core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/LoadMappingIterator.java
```diff
@@ -0,0 +1,84 @@
...
+      return reader.hasNext();
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
+    }
+  }
+
+  @Override
+  public Map.Entry<KeyExtent,Bulk.Files> next() {
+    if (!hasNext())
+      return null;
```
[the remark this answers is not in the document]
ctubbsii: Not really a fan of the null-oriented programming here.  Would prefer caller check hasNext and we throw an exception here. However, I'm not sure what would be better.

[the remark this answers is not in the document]
keith-turner: Guava has a static `Iterators.peekingIterator()` that could possibly be used to simply the loop that needs this null behavior by avoiding the call to next before the loop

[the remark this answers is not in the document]
milleruntime: >  Would prefer caller check hasNext and we throw an exception here. However, I'm not sure what would be better.

This is basically what we are currently doing as the JSON will throw an exception if you try to get more from the gson object.

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -328,38 +333,45 @@ private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator lmi,
...
-      }
+      // get first tablet in range
+      currentTablet = getFirstTablet(tabletIter, currentTablet, fileTablet.getPrevEndRow());
       tablets.add(currentTablet);
 
-      while (!Objects.equals(currentTablet.getEndRow(), fke.getEndRow())) {
-        currentTablet = tabletIter.next();
-        tablets.add(currentTablet);
-      }
+      // get all tablets between currentTablet and EndRow given in mapping file
```
keith-turner: Personally I think these loops inlined are easier to understand.  The method names do not give me any concrete idea what the method will do with the parameters.  So I still need to go look at the method impl to understand the big picture.  

I think the loops inlined with comments you put before the methods would be good.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/LoadMappingIterator.java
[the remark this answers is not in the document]
milleruntime: The PeekingIterator (I used the one in our source that has been there since day one FYI) seems to work well.  Used it in 8d933ab

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
keith-turner: Those names would be better.  The words `skip` and `advance` are other possible method name prefixes.

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -328,38 +334,45 @@ private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator lmi,
...
+    while (lmi.hasNext()) {
+      loadMapEntry = lmi.next();
+      KeyExtent fileTablet = loadMapEntry.getKey();
       tablets.clear();
 
-      while (!Objects.equals(currentTablet.getPrevEndRow(), fke.getPrevEndRow())) {
-        currentTablet = tabletIter.next();
-      }
+      // get first tablet in range
+      currentTablet = getFirstTablet(tabletIter, currentTablet, fileTablet.getPrevEndRow());
```
keith-turner: I think collapsing these two functions into one would make this code more readable. Then this loop could look like

```java
// note in this code tabletIters is a peeking iter also
    while (lmi.hasNext()) {
      loadMapEntry = lmi.next();
      KeyExtent loadRange = loadMapEntry.getKey();

      List<TabletMetadata> tablets = findOverlapping(loadRange, tabletIter);

      loader.load(tablets, loadMapEntry.getValue());
    }
```

With a find overlapping that looks like the following.

```java
  private List<TabletMetadata> findOverlapping(KeyExtent loadRange,
      PeekingIterator<TabletMetadata> tabletIter) {

    List<TabletMetadata> tablets = new ArrayList<>();

    // skip tablets until have the same prev end row
    while (!Objects.equals(tabletIter.peek().getPrevEndRow(), loadRange.getPrevEndRow())) {
      tabletIter.next();
    }

    // add tablets until the end row matches the load range
    while (true) {
      TabletMetadata tablet = tabletIter.next();
      tablets.add(tablet);
      if (Objects.equals(tablet.getEndRow(), loadRange.getEndRow()))
        break;
    }

    return tablets;
  }
```

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -328,38 +334,45 @@ private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator lmi,
...
-        }
-        loadMapEntry = lmi.next();
-      }
-      KeyExtent fke = loadMapEntry.getKey();
-      Files files = loadMapEntry.getValue();
-      loadMapEntry = null;
-
+    while (lmi.hasNext()) {
+      loadMapEntry = lmi.next();
+      KeyExtent fileTablet = loadMapEntry.getKey();
```
keith-turner: I think this should be called `loadRange`

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
keith-turner: Actually `tabletIter` does not have to be a peeking iter if the code is written as it currently is.

```java
 private List<TabletMetadata> findOverlapping(KeyExtent loadRange,
      Iterator<TabletMetadata> tabletIter) {

    List<TabletMetadata> tablets = new ArrayList<>();
    TabletMetadata tablet = tabletIter.next();
    
    // skip tablets until have the same prev end row
    while (!Objects.equals(tablet.getPrevEndRow(), loadRange.getPrevEndRow())) {
      tablet = tabletIter.next();
    }
    tablets.add(tablet);
    
    // add tablets until the end row matches the load range
    while(!Objects.equals(tablet.getEndRow(), loadRange.getEndRow())) {
      tablet = tabletIter.next();
      tablets.add(tablet);
    };

    return tablets;
  }
```

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -304,10 +310,10 @@ long finish() throws Exception {
    * scan the metadata table getting Tablet range and location information. It will return 0 when
    * all files have been loaded.
    */
-  private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator lmi, Master master,
-      long tid) throws Exception {
-
-    Map.Entry<KeyExtent,Bulk.Files> loadMapEntry = lmi.next();
+  private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator loadMapIter,
+      Master master, long tid) throws Exception {
+    PeekingIterator<Map.Entry<KeyExtent,Bulk.Files>> lmi = new PeekingIterator(loadMapIter);
```
keith-turner:
```suggestion
    PeekingIterator<Map.Entry<KeyExtent,Bulk.Files>> lmi = new PeekingIterator<>(loadMapIter);
```

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
milleruntime: That is probably a good idea to combine the two into one function with a singular purpose. What about the name "findOverlappingTablets"?  It is a bit longer but states what is being found.

keith-turner: `findOverlappingTablets` sounds good.  I was not sure what to name it.

milleruntime:
Good suggestions, thanks @keith-turner .  You OK with the other changes? Can you take a quick look at the scanTime math in the finish method to make sure I preserved the calculations.

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/PrepBulkImport.java
```diff
@@ -162,11 +162,9 @@ private void checkForMerge(final Master master) throws Exception {
 
       Iterators.transform(lmi, entry -> entry.getKey());
```
keith-turner:
```suggestion
```

server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java
```diff
@@ -328,38 +331,35 @@ private long loadFiles(Table.ID tableId, Path bulkDir, LoadMappingIterator lmi,
...
+    // skip tablets until we find the prevEndRow of loadRange
+    while (!Objects.equals(currentTablet.getPrevEndRow(), loadRange.getPrevEndRow())) {
+      currentTablet = tabletIter.next();
     }
-    long t2 = System.currentTimeMillis();
+    // we have found the first tablet in the range, add it to the list
+    tablets.add(currentTablet);
 
-    long sleepTime = loader.finish();
-    if (sleepTime > 0) {
```
keith-turner: Personally I would leave this code as is in order to avoid duplicating the code for dealing with scan time.

milleruntime: I guess I thought it was confusing that we had multiple calculations of the sleep time, one returned from the finish method and then more calculations dealing with the scan time.  I was trying to put it all in the one method (finish), even if it was in two places, offline and online.

keith-turner:
@milleruntime  442fce0 is what I was thinking about for inlining.

PR #894, opened 2019-01-11
Fix #883 Specify client props as HDFS path in new M/R API

### LINKED ISSUE #883: Specify client properties as path in new MapReduce API
The new MapReduce API for 2.0 needs a way to specify client properties as path (that can be cached in HDFS) to provide users with way to avoid serializing properties in Job configuration.

mikewalch:
* Also cleaned up new M/R API by removing unnecessary methods
  and using properties in place of ClientInfo

### DIFF
--- a/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/InputFormatBuilder.java
+++ b/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/InputFormatBuilder.java
@@ -45,14 +45,26 @@ public interface InputFormatBuilder {
    * @since 2.0
    */
   interface ClientParams<T> {
+
     /**
-     * Set the connection information needed to communicate with Accumulo in this job.
-     * clientProperties param can be created using {@link Accumulo#newClientProperties()}
+     * Set client properties needed to communicate with Accumulo for this job. This information will
+     * be serialized into the configuration. Therefore, it is more secure to use
+     * {@link #clientPropertiesPath(String)}. Client properties can be created using
+     * {@link Accumulo#newClientProperties()}
      *
      * @param clientProperties
      *          Accumulo connection information
      */
     TableParams<T> clientProperties(Properties clientProperties);
+
+    /**
+     * Set path to HDFS location containing accumulo-client.properties file. This setting is more
+     * secure than {@link #clientProperties(Properties)}
+     *
+     * @param clientPropsPath
+     *          HDFS path to accumulo-client.properties
+     */
+    TableParams<T> clientPropertiesPath(String clientPropsPath);
   }
 
   /**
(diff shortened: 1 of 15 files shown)


hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoopImpl/mapreduce/lib/ConfiguratorBase.java
```diff
@@ -102,69 +78,26 @@ protected static String enumToConfKey(Enum<?> e) {
...
-      Properties props) {
-    StringWriter writer = new StringWriter();
-    try {
-      props.store(writer, "client properties");
-    } catch (IOException e) {
-      throw new IllegalStateException(e);
+      Properties props, String clientPropsPath) {
+    if (clientPropsPath != null) {
+      try {
+        DistributedCacheHelper.addCacheFile(new URI(clientPropsPath), conf);
```
milleruntime: Should update the DistributedCacheHelper to not use deprecated methods so it can be un-deprecated.  This could be a separate PR though

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/InputFormatBuilder.java
```diff
@@ -45,14 +45,26 @@
...
      *          Accumulo connection information
      */
     TableParams<T> clientProperties(Properties clientProperties);
+
+    /**
+     * Set path to HDFS location containing accumulo-client.properties file. This setting is more
+     * secure than {@link #clientProperties(Properties)}
+     *
+     * @param clientPropsPath
+     *          HDFS path to accumulo-client.properties
```
milleruntime: This could be from a DFS location and doesn't have to be HDFS right?

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoopImpl/mapreduce/lib/ConfiguratorBase.java
mikewalch: I need something that works with both mapreduce & mapred.  I am not sure what a better option is.  We might be able to do the following but I think this should be a separate issue (I will create one):
```java
    String files = conf.get("mapreduce.job.cache.files");
    conf.set("mapreduce.job.cache.files", files == null ? uri.toString() : files + "," + uri.toString());
```

hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce/InputFormatBuilder.java
mikewalch: Yes it doesn't have be HDFS.  I update the docs.

PR #899, opened 2019-01-15
Move encrypted log flush to synch thread. Fixes #839

### LINKED ISSUE #839: Crypto flushes too frequently
While working on #829 I looked into why there is a separate flush for the encrypting stream.  This is required, along with the NoFlushOutputStream so that data will be written to HDFS without having to flush the FSDataOutputStream.  But the flush on the encryption stream should be done in the sync thread and not every time a key/value pair is written.

milleruntime:


server/tserver/src/main/java/org/apache/accumulo/tserver/log/DfsLogger.java
```diff
@@ -558,7 +559,6 @@ public LoggerOperation defineTablet(CommitSession cs) throws IOException {
   private synchronized void write(LogFileKey key, LogFileValue value) throws IOException {
     key.write(encryptingLogFile);
     value.write(encryptingLogFile);
-    encryptingLogFile.flush();
```
keith-turner: This moved from a synchronzied block to an unsynchronized block.   Could possibly be problematic, but I am not completely sure.  Depends a lot on the stream impl.

milleruntime: Isn't this the same behavior of the WAL normally, when it is not encrypted?  Flushing/syncing in a separate unsychronized thread, while many threads write key values in a synchronized method?

keith-turner: It's not quite the same.  What you described is for the lower level hdfs output stream.  This change makes the higher level encryption stream do that.

milleruntime: I tested this with Uno and the encryption plugin enabled.  I was able to ingest 3 million rows and scan through the shell.  So I believe this works with our AES crypto impl.  I guess if someone discovered an impl that it doesn't work with, we can address it then.

milleruntime:
Looking further at the impl of CryptoOutputStream, it seems like its OK to keep it as is.

PR #902, opened 2019-01-15
Add exclusion to test for vuln in mini-kdc lib

milleruntime:
Hadoop mini-kdc brings in a vulnerable dependency we don't need so we can just exclude it.


milleruntime:
This is to avoid https://nvd.nist.gov/vuln/detail/CVE-2017-12974

PR #904, opened 2019-01-16
Exclude vuln thrift dep from build

milleruntime:
Snyk CLI tool was flagging this Thrift dependency.  Excluding it from our pom seems to have no effect.


milleruntime:
@ctubbsii Can you check this PR and get this in with the release?  With this update the tool should report 0 maven dependency vulnerabilities upon release.

PR #905, opened 2019-01-16
fixes #813 close server side session when scanner closed

### LINKED ISSUE #813: Close scan session when not all data is read.
If  only some data is read from a scanner and then the scanner is closed then this may leave server side scan sessions open. Normally this does not matter because these scan sessions do not use much memory.  However in the case of an isolated scan then scan session can use a lot of memory.  While working on #797 I had refactored a metadata scan from `Range(x,y)` to `Range(x, null)` and did not read all data.  This resulted in tablet servers dying with an out of memory error.

Currently scanners have a close method that is a noop.  However it would be nice if this close method cleaned up any server side scan sessions.

keith-turner:


core/src/main/java/org/apache/accumulo/core/clientImpl/ScannerIterator.java
```diff
@@ -58,18 +59,24 @@
...
   private static ThreadPoolExecutor readaheadPool = new ThreadPoolExecutor(0, Integer.MAX_VALUE, 3L,
       TimeUnit.SECONDS, new SynchronousQueue<>(),
       new NamingThreadFactory("Accumulo scanner read ahead thread"));
 
+  private boolean closed = false;
+
   ScannerIterator(ClientContext context, Table.ID tableId, Authorizations authorizations,
       Range range, int size, long timeOut, ScannerOptions options, boolean isolated,
-      long readaheadThreshold) {
+      long readaheadThreshold, Set<ScannerIterator> activeIters) {
```
milleruntime: I see you are using a SynchronizedSet in ScannerImpl but this constructor only takes a set.  Isn't it possible to pass in a non synchronized set?

core/src/main/java/org/apache/accumulo/core/clientImpl/ScannerIterator.java
```diff
@@ -112,16 +120,32 @@ public boolean hasNext() {
     throw new NoSuchElementException();
   }
 
+  void close() {
+    // run actual close operation in the background so this does not block.
+    readaheadPool.execute(() -> {
+      synchronized (scanState) {
```
milleruntime: Is it possible to consume all the threads in the pool while waiting for ThriftScanner to close?  Like if I were to call close in a loop.

keith-turner: No,  I think that pool will created an unlimited number of threads.  Your question made me realize I was probably tracking a max of too many iterators.  I lowered it from 1000 to 32.  The use case I am trying to support in someone creating a scanner, creating a few iterators, and then closing.  If someone is created a large number of iterators, I am not really trying to support that.

core/src/main/java/org/apache/accumulo/core/clientImpl/ScannerIterator.java
```diff
@@ -99,6 +106,7 @@ public boolean hasNext() {
     iter = getNextBatch().iterator();
     if (!iter.hasNext()) {
       finished = true;
+      activeIters.remove(this);
```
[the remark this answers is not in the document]
keith-turner: I changed how this was implemented.

core/src/main/java/org/apache/accumulo/core/clientImpl/ScannerIterator.java
keith-turner: I changed how this was implemented.

core/src/main/java/org/apache/accumulo/core/clientImpl/ScannerIterator.java
keith-turner: I am thinking ideally the close messages would be grouped by tablets server and then the messages sent.  

This makes me think it would be nice to have a service for making async calls.  I can just queue up async calls I want to make and the service will group them by tablet server.

mikewalch:
I think this PR broke SessionBlockVerifyIT

keith-turner:
@mikewalch  I think I fixed it in 43626decef4577e1f38d281a94aa91eab4848852

PR #906, opened 2019-01-16
Fix #876 - Refactor WorkAssigner Tests to use AccumuloClient

### LINKED ISSUE #876: Refactor WorkAssigner Tests to use AccumuloClient better
DistributedWorkQueueWorkAssigner has two implementations, SequentialWorkAssigner and UnorderedWorkAssigner.  Recent changes allow these classes to take an AccumuloClient as an argument to the constructors. AccumuloClient should contain objects that their parent class (DistributedWorkQueueWorkAssigner) has getters and setters for, such as client and zooCache.  The following tests should be refactored so we no longer need these getters and setters in DistributedWorkQueueWorkAssigner.

- SequentialWorkAssignerTest
- UnorderedWorkAssignerTest
- SequentialWorkAssignerIT
- UnorderedWorkAssignerIT

It also might be possible to refactor so we don't need setWorkQueue as well since this should be initialized automatically.

cjmctague:
- Remove the use of setClient() and setZooCache()
- remove setClient() and setZooCache() from
DistributedWorkQueueWorkAssigner as they are no longer used
- refactor and remove setClient and setZooCache from IT's
- refactor workQueue

### DIFF
--- a/server/master/src/test/java/org/apache/accumulo/master/replication/SequentialWorkAssignerTest.java
+++ b/server/master/src/test/java/org/apache/accumulo/master/replication/SequentialWorkAssignerTest.java
@@ -66,30 +66,29 @@ public class SequentialWorkAssignerTest {
 
     queuedWork.put("cluster1", cluster1Work);
 
-    assigner.setClient(client);
-    assigner.setZooCache(zooCache);
-    assigner.setWorkQueue(workQueue);
+    assigner.zooCache = zooCache;
+    assigner.workQueue = workQueue;
     assigner.setQueuedWork(queuedWork);
 
     expect(client.getInstanceID()).andReturn("instance");
 
     // file1 replicated
-    expect(zooCache.get(ZooUtil.getRoot("instance") + ReplicationConstants.ZOO_WORK_QUEUE + "/"
-        + DistributedWorkQueueWorkAssignerHelper.getQueueKey("file1",
+    expect(assigner.zooCache.get(ZooUtil.getRoot("instance") + ReplicationConstants.ZOO_WORK_QUEUE
+        + "/" + DistributedWorkQueueWorkAssignerHelper.getQueueKey("file1",
             new ReplicationTarget("cluster1", "1", Table.ID.of("1"))))).andReturn(null);
     // file2 still needs to replicate
     expect(
-        zooCache
+        assigner.zooCache
             .get(ZooUtil.getRoot("instance") + ReplicationConstants.ZOO_WORK_QUEUE + "/"
                 + DistributedWorkQueueWorkAssignerHelper.getQueueKey("file2",
                     new ReplicationTarget("cluster1", "2", Table.ID.of("2")))))
                         .andReturn(new byte[0]);
 
-    replay(workQueue, zooCache, client);
+    replay(assigner.workQueue, assigner.zooCache, assigner.client);
 
     assigner.cleanupFinishedWork();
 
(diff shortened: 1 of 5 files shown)


milleruntime:
@cjmctague thanks for the contribution but this is not _quite_ what I had in mind with the refactor.  I was thinking only have calls to the constructor and then the assignWork() method, which is the typical behavior of the class.  The assignWork() method should do the work to initialize the workQueue and create the zooCache, instead of just setting them directly in the test.  This will require more mocking up of objects like ZooUtil but it would be a more realistic test.

server/master/src/test/java/org/apache/accumulo/master/replication/SequentialWorkAssignerTest.java
```diff
@@ -66,30 +66,29 @@ public void basicZooKeeperCleanup() {
 
     queuedWork.put("cluster1", cluster1Work);
 
-    assigner.setClient(client);
-    assigner.setZooCache(zooCache);
-    assigner.setWorkQueue(workQueue);
+    assigner.zooCache = zooCache;
+    assigner.workQueue = workQueue;
```
milleruntime: These should be replaced with a single call to `assigner.assignWork()`

cjmctague: Gotcha that would create the zooCache and workQueue but we would still need to assign them to a "Mock" version of it right?

cjmctague:
@milleruntime Sorry for the misunderstanding. Can you clarify the use case you are talking about for the more realistic case as well as the "Mock" up of ZooUtil

server/master/src/test/java/org/apache/accumulo/master/replication/SequentialWorkAssignerTest.java
milleruntime: You would have to do it a bit differently and setup the EasyMock objects first in the init() method, similar to how we do it in other tests (see [InMemoryMapTest](https://github.com/milleruntime/accumulo/blob/c0f9e322111c8f6bcbfcaf1c67160756153b63a5/server/tserver/src/test/java/org/apache/accumulo/tserver/InMemoryMapTest.java#L112)).  Then call replay at the end of the init method.  You will have to mock each object first calling _EasyMock.expect(..)_ with _andReturn(..)_ and _anyTimes()_ using the EasyMock framework for any objects that the code being called requires.  This usually takes some trial and error, running the test to see what gets called.  The EasyMock framework will throw exceptions telling you what it is missing.  

For example, looking at `DistributedWorkQueueWorkAssigner.initializeWorkQueue()`  you will need to add this to init to mock up the InstanceID:
`    EasyMock.expect(client.getInstanceID()).andReturn("testID").anyTimes();
`

milleruntime:
> @milleruntime Sorry for the misunderstanding. Can you clarify the use case you are talking about for the more realistic case as well as the "Mock" up of ZooUtil

No worries.  The EasyMock stuff can be weird and I am probably not doing a great job explaining it.  So please don't hesitate to ask more questions :+1:

cjmctague:
Sorry for the delay on this. I was tasked with some time sensitive tasks that prevented me from focusing on this.

milleruntime:
No worries.  This was just an idea that I thought would be straight forward but turns out wasn't really.  I won't be offended if you close the PR :smile:

PR #909, opened 2019-01-18
Reduce use of CachedConfiguration for #735

milleruntime:
* Add hadoop conf to ClientContext and utilize where available
* Drop use of CachedConfiguration in Volume Manager
* Remove unused instances of CachedConfiguration from Tests
* Replace with new object in unit tests where reuse doesn't matter


milleruntime:
BulkImportMonitoringIT, CountNameNodeOpsBulkIT and BulkIT are passing with these changes.

PR #912, opened 2019-01-23
Replace use of CachedConfiguration

milleruntime:
* Make some ServerConstants methods take ServerContext
* Removed method that uses CachedConfigruation from VolumeConfiguraiton
* Removed CachedConfigruation from utils and initialize
* Add hadoop conf to LogCloser
* Pull hadoop conf from server context in ITs

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
@@ -274,8 +274,7 @@ public class VolumeUtil {
 
     VolumeChooserEnvironment chooserEnv = new VolumeChooserEnvironment(extent.getTableId(),
         context);
-    Path newDir = new Path(vm.choose(chooserEnv,
-        ServerConstants.getBaseUris(context.getConfiguration(), context.getHadoopConf()))
+    Path newDir = new Path(vm.choose(chooserEnv, ServerConstants.getBaseUris(context))
         + Path.SEPARATOR + ServerConstants.TABLE_DIR + Path.SEPARATOR + dir.getParent().getName()
         + Path.SEPARATOR + dir.getName());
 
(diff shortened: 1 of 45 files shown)


server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
```diff
@@ -166,7 +165,7 @@ public TabletFiles(String dir, List<LogEntry> logEntries,
   public static String switchRootTableVolume(ServerContext context, String location)
       throws IOException {
     String newLocation = switchVolume(location, FileType.TABLE,
-        ServerConstants.getVolumeReplacements(context.getConfiguration()));
+        ServerConstants.getVolumeReplacements(context.getConfiguration(), context.getHadoopConf()));
```
keith-turner: Could only context be passed here?

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
```diff
@@ -275,10 +274,10 @@ private static String decommisionedTabletDir(ServerContext context, ZooLock zooL
 
     VolumeChooserEnvironment chooserEnv = new VolumeChooserEnvironment(extent.getTableId(),
         context);
-    Path newDir = new Path(
-        vm.choose(chooserEnv, ServerConstants.getBaseUris(context.getConfiguration()))
-            + Path.SEPARATOR + ServerConstants.TABLE_DIR + Path.SEPARATOR
-            + dir.getParent().getName() + Path.SEPARATOR + dir.getName());
+    Path newDir = new Path(vm.choose(chooserEnv,
+        ServerConstants.getBaseUris(context.getConfiguration(), context.getHadoopConf()))
```
keith-turner: Could only context be passed here?

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
milleruntime: In some cases yes but we still use this in Initialize where we don't have a good way to create ServerContext yet.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
milleruntime: In some cases yes but we still use this in Initialize where we don't have a good way to create ServerContext yet.

keith-turner: Ok..  Could possibly have two methods.  One that takes context and one that takes everything.  This would simplify most code while leaving the init code as is.

PR #934, opened 2019-02-01
Add default impl to new IteratorEnvironment method

milleruntime:
This will help with 2.0 upgrades


milleruntime:
> Are there any implementations that can be removed?

This could replace some that were done for #513.  Like SimpleIterEnv and BaseIteratorEnvironment

PR #935, opened 2019-02-01
Merged multiple compaction strategies into one.

keith-turner:
While testing 2.0.0-alpha-2 I found that I wanted the functionality of
the size limit and two tier compactions strategies. However could only
configure one.  This PR merges those two into one.  Two tier can be
dropped because it is new in 2.0.0.


server/tserver/src/main/java/org/apache/accumulo/tserver/compaction/strategies/BasicCompactionStrategy.java
```diff
@@ -0,0 +1,164 @@
...
+ *
+ * <p>
+ * To use this strategy with Minor Compactions set {@code table.file.compress.type=snappy} and set a
+ * different compress type in {@code table.majc.compaction.strategy.opts.large.compress.type} for
+ * larger files.
+ *
+ * <p>
+ * The options that take sizes are in bytes and the suffixes K,M,and G can be used.
+ */
+public class BasicCompactionStrategy extends DefaultCompactionStrategy {
```
milleruntime: Could call it ConfigurableCompactionStrategy and rename the current ConfigurableCompactionStrategy to ShellCompactionStrategy since its less configurable and seems to be only used by the shell compact command.

keith-turner: I thought about that, but did not like creating something new with same name as something old for user configurable pluggable thing.

mikewalch: I am not a fan of this name either but I don't know a better name.

mikewalch:
@keith-turner, you should update https://accumulo.apache.org/docs/2.x/getting-started/table_configuration#compaction-strategies

PR #939, opened 2019-02-06
Add ScannerBase.fetchColumn/Family using CharSequence

mikewalch:
* Modified some but not all tests to use new method

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
@@ -105,6 +105,7 @@ public interface ScannerBase extends Iterable<Entry<Key,Value>>, AutoCloseable {
    *
    * @param colFam
    *          the column family to be fetched
+   * @since 2.0.0
    */
   void fetchColumnFamily(CharSequence colFam);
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
```diff
@@ -121,6 +139,37 @@
    */
   void fetchColumn(Text colFam, Text colQual);
 
+  /**
+   * Adds a column to the list of columns that will be fetched by this scanner. The column is
+   * identified by family and qualifier. By default when no columns have been added the scanner
+   * fetches all columns.
+   *
+   * <p>
+   * <b>WARNING</b>. Using this method with custom iterators may have unexpected results. Iterators
```
keith-turner: Could say see the warning on {@link #fetchColumn(Text, Text)}

core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
```diff
@@ -121,6 +139,37 @@
...
+   *
+   * <p>
+   * tl;dr If using a custom iterator with a seek method that adds column families, then may want to
+   * avoid using this method.
+   *
+   * @param colFam
+   *          the column family of the column to be fetched
+   * @param colQual
+   *          the column qualifier of the column to be fetched
+   */
```
keith-turner: Need a `@since` tag.

core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
```diff
@@ -90,6 +90,24 @@
...
+   *
+   * <p>
+   * When used in conjunction with custom iterators, the set of column families fetched is passed to
+   * the top iterator's seek method. Custom iterators may change this set of column families when
+   * calling seek on their source.
+   *
+   * @param colFam
+   *          the column family to be fetched
+   */
+  void fetchColumnFamily(CharSequence colFam);
```
keith-turner: Need a `@since` tag.

core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
mikewalch: Fixed. See 787590acc34707

core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
mikewalch: Fixed. See 787590acc34707

core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
mikewalch: Fixed. See 787590acc34707

test/src/main/java/org/apache/accumulo/test/ConditionalWriterIT.java
```diff
@@ -203,7 +203,7 @@ public void testBasic() throws Exception {
         assertEquals(Status.REJECTED, cw.write(cm5).getStatus());
 
         // ensure rejected mutations did not write
-        scanner.fetchColumn(new Text("name"), new Text("last"));
```
keith-turner: So much nicer.

milleruntime:
Can you add a default implementation since it is a new method on an existing interface?  Looks good otherwise.

PR #963, opened 2019-02-14
Fix #910 - Add setClasspathItems() to MiniAccumuloConfig

### LINKED ISSUE #910: Make Accumulo maven plugin only use public API or move it back into Accumulo
Opening an issue for apache/accumulo-maven-plugin#2 in Accumulo because I think it needs to be resolved before releasing 2.0.0.

mikewalch:
* This allows accumulo-maven-plugin to only use public API

See related PR: apache/accumulo-maven-plugin#3

### DIFF
--- a/minicluster/src/main/java/org/apache/accumulo/minicluster/MiniAccumuloConfig.java
+++ b/minicluster/src/main/java/org/apache/accumulo/minicluster/MiniAccumuloConfig.java
@@ -264,7 +264,7 @@ public class MiniAccumuloConfig {
    * @return the current instance
    * @since 2.0.0
    */
-  public MiniAccumuloConfig setClasspathItems(String... classpathItems) {
+  public MiniAccumuloConfig setClasspath(String... classpathItems) {
     impl.setClasspathItems(classpathItems);
     return this;
   }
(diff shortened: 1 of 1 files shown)


minicluster/src/main/java/org/apache/accumulo/minicluster/MiniAccumuloConfig.java
```diff
@@ -255,4 +255,17 @@ public MiniAccumuloConfig setNativeLibPaths(String... nativePathItems) {
...
+
+  /**
+   * Sets the classpath elements to use when spawning processes.
+   *
+   * @param classpathItems
+   *          the classpathItems to set
+   * @return the current instance
+   * @since 2.0.0
+   */
+  public MiniAccumuloConfig setClasspathItems(String... classpathItems) {
```
keith-turner: Could call this `setClasspath` and drop `Items`.  I don't think `Items` adds any useful information.

mikewalch: Pushed an update for this

mikewalch:
> This is okay. It exposes something which is currently being used, but as public API. However, I think it might be better if we simply expose something like `addEnvironment(key, value)`, so it's a bit more future-proof. Then, the user can set the `CLASSPATH` environment variable when mini launches processes, but they can also set things like `USER` and `HOME` and `LD_LIBRARY_PATH`, and others, as needed.

I think this is an interesting idea that could be explored in a future issue.

PR #972, opened 2019-02-19
Support dispatching and prioritizing by user scan types

keith-turner:
While working on apache/fluo#1055 I found it cumbersome to configure
Fluo and Accumulo to dispatch Fluo notification scans to a dedicated
executor.  With these changes its much simpler to do that.  Fluo can
set an execution hint  like `scan_type=fluo-ntfy` and Accumulo can be
configured to execute the scan based on the type without changing Fluo
config or source code.

Before these changes, Fluo would have required either a custom
dispatcher or custom Fluo config for Accumulo executors. With these
changes nothing needs to be done in Fluo.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
@@ -341,6 +341,11 @@ public interface ScannerBase extends Iterable<Entry<Key,Value>>, AutoCloseable {
    * scan, only how quickly it is returned.
    *
    * <p>
+   * Using the hint {@code scan_type=<type>} and documenting all of the types for your application
+   * is one strategy to consider. This allows administrators to adjust executor and prioritizer
+   * config for your application scan types without having to change the application source code.
+   *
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/accumulo/core/client/ScannerBase.java
```diff
@@ -341,6 +341,11 @@ default void fetchColumn(CharSequence colFam, CharSequence colQual) {
    * scan, only how quickly it is returned.
    *
    * <p>
+   * Using the hint {@code scan_type=<type>} and documenting all of the types for your application
+   * is one strategy to consider. This allows administrators to adjust executor and prioritizer
+   * config for your application scan types without having to change the application source code.
```
mikewalch: You should also update the website (if needed): https://accumulo.apache.org/docs/2.x/administration/scan-executors

keith-turner: apache/accumulo-website#157

PR #973, opened 2019-02-20
fixes #768 fix race condition in generation of table map

### LINKED ISSUE #768: Handle TableOps exists() issues

keith-turner:


core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
```diff
@@ -98,4 +106,8 @@ public TableMap(Instance instance, ZooCache zooCache) {
   public Map<String,String> getIdtoNameMap() {
     return tableIdToNameMap;
   }
+
+  public boolean isCurrent(ZooCache zc) {
+    return this.zooCache == zc && this.updateCount == zc.getUpdateCount();
+  }
```
milleruntime: This method is nice.  Would it be better to use equals comparison?

core/src/main/java/org/apache/accumulo/core/client/impl/Tables.java
```diff
@@ -132,22 +111,35 @@ public static String getTableName(Instance instance, String tableId)
...
     TableMap map;
+
+    final ZooCache zc = getZooCache(instance);
+
+    map = getTableMap(instance, zc);
+
+    if (!map.isCurrent(zc)) {
+      instanceToMapCache.invalidate(instance.getInstanceID());
+      map = getTableMap(instance, zc);
+    }
```
milleruntime: Do we need to synchronize on the TableMap cache?  Or does guava handle that?

keith-turner: Guava handles that

core/src/main/java/org/apache/accumulo/core/client/impl/TableMap.java
keith-turner: In this case I wanted to make sure it was the exact same zoocache instance

milleruntime: Ah right, gotcha.

Issue #978, opened 2019-02-22
Log waits beyond a threshold in FileManager.reserveReaders()


FineAndDandy:
While calling FileManager.reserveReaders() filePermits.acquireUninterruptibly(files.size()) is called. This method will block until the filePermits are available. Using tryAcquire(int, long, TimeUnit) in a loop until successful would allow logging beyond a configurable threshold while waiting.

keith-turner:
In addition to logging, could also have metrics ... see #946

FineAndDandy:
closed via #981

PR #981, opened 2019-02-22
Add logging for slow file permits requests. (#978)

FineAndDandy:


server/tserver/src/main/java/org/apache/accumulo/tserver/FileManager.java
```diff
@@ -287,7 +290,15 @@ private void closeReaders(Collection<FileSKVIterator> filesToClose) {
     Map<FileSKVIterator,String> readersReserved = new HashMap<>();
 
     if (!tablet.isMeta()) {
+      long start = System.currentTimeMillis();
       filePermits.acquireUninterruptibly(files.size());
```
keith-turner: Could use the tryAcquire method on the semaphore.  Something like

```java
      boolean acquired = filePermits.tryAcquire(files.size(), slowFilePermitMillis, TimeUnit.MILLISECONDS);
      if(!acquired) {
        //log message
        filePermits.acquireUninterruptibly(files.size());
      }
```

It would make the code slightly more complicated

server/tserver/src/main/java/org/apache/accumulo/tserver/FileManager.java
```diff
@@ -287,7 +290,15 @@ private void closeReaders(Collection<FileSKVIterator> filesToClose) {
...
+      long waitTime = System.currentTimeMillis() - start;
+
+      if (waitTime >= slowFilePermitMillis) {
+        String message = new StringBuilder(128).append("Slow file permits request: ")
+            .append(waitTime).append(" ms, files requested: ").append(files.size()).toString();
+        log.info(message);
```
keith-turner: This could iuse slf4j substitution instead of the string builder.  May be useful to include the key extent in the log message so the tablet is known.

Any particular reason for logging at info? I don't think the level should be changed, just curious.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -525,6 +525,9 @@
   TSERV_SLOW_FLUSH_MILLIS("tserver.slow.flush.time", "100ms", PropertyType.TIMEDURATION,
       "If a flush to the write-ahead log takes longer than this period of time,"
           + " debugging information will written, and may result in a log rollover."),
+  TSERV_SLOW_FILEPERMIT_MILLIS("tserver.slow.filepermit.time", "100ms", PropertyType.TIMEDURATION,
```
keith-turner: A larger default may be good.  If its too low it will fire too often and will be ignored.  Maybe at least one second?  or 500ms?

server/tserver/src/main/java/org/apache/accumulo/tserver/FileManager.java
FineAndDandy: I initially implemented it that way, but the code bloat was a bit larger than that. tryAcquire returns false after the timeout so a while loop has to be used, time has to be aggregated over all the loops, and interrupts have to be handled. Due to the time aggregating I found myself using a start/stop time in addition to the timeout time for total wait time anyway.

server/tserver/src/main/java/org/apache/accumulo/tserver/FileManager.java
FineAndDandy: TSERV_SLOW_FLUSH_MILLIS logs at INFO so I was trying to follow the lead there. I will add the key extent and convert to slf4j sbustitution

core/src/main/java/org/apache/accumulo/core/conf/Property.java
FineAndDandy: I would think that a system that is blocking at all waiting for file permits would want to know since it ends up being rolled into the scan time. Since proper configuration of the tserver readahead/maxfiles open and the various compaction settings for files/ratios will prevent this log message from ever firing. I'm happy to make the default whatever value you prefer though.

keith-turner: > I would think that a system that is blocking at all waiting for file permits would want to know since it ends up being rolled into the scan time.

Reading that made me think it may be nice to have this time as a metric.

server/tserver/src/main/java/org/apache/accumulo/tserver/FileManager.java
keith-turner: > I initially implemented it that way, but the code bloat was a bit larger than that.

I agree its not worth the bloat because I can't think of any strong need for immediacy.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
keith-turner: > Since proper configuration of the tserver readahead/maxfiles open and the various compaction settings for files/ratios will prevent this log message from ever firing.

I was thinking it may happen occasionally under normal circumstances, but its not a concern unless it too long.  But what is too long,  I don't know.  The thinking that it should never happen is a valid view point and may suggest a lower default value.  Now I am not sure what the default should be, because I see merit in both view points.  This makes me like the idea of a metric a bit more.

keith-turner: Also, I am fine with leaving at 100ms.  That time seems like a good default for we don't expect this to ever happen, but if it does its ok as long as its not too long.

keith-turner: I think it would be good to add this to the set HOT_PATH_PROPERTIES as it will be checked on a per scan basis.

FineAndDandy: okay

PR #993, opened 2019-02-27
Consolidate option parsing classes

mikewalch:


milleruntime:
What are these changes doing?

mikewalch:
> What are these changes doing?

We have a lot of option parsing classes that subclass each other.  It's confusing.  This change reduces the number of these classes and removes code.

milleruntime:
OK I see... you are just removing a bunch of similar CLI classes and replacing them with ClientOpts.

server/tracer/src/main/java/org/apache/accumulo/tracer/TraceTableStats.java
```diff
@@ -33,14 +33,15 @@
...
  * Reads the trace table and prints out some stats about the spans found.
  */
 public class TraceTableStats {
-  static class Opts extends ClientOnDefaultTable {
-    public Opts() {
-      super("trace");
-    }
+  static class Opts extends ClientOpts {
+    @Parameter(names = "--table", description = "table to use")
+    String tableName = "trace";
```
milleruntime: Why not just add tableName to ClientOpts?

mikewalch: Options don't always require/use a table.

milleruntime: Right but can't you set it to null like some of the options that aren't always used?

mikewalch: I know you can set a parameter to default to null but that doesn't remove it as option. It'll still print out on the command line if user prints the help.

milleruntime: Oh right, I forgot about the Help.

PR #997, opened 2019-02-27
Create MapCounterInt for counting integers

milleruntime:
Created MapCounterInt as possible fix to [this](https://lgtm.com/projects/g/apache/accumulo/snapshot/7a9a5f6b26c77a59a9a195b9ea29355f7f6f2a1e/files/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java?sort=name&dir=ASC&mode=heatmap#x2f5b87141746ce7c:1) at line 3209 and [this](https://lgtm.com/projects/g/apache/accumulo/snapshot/7a9a5f6b26c77a59a9a195b9ea29355f7f6f2a1e/files/server/base/src/main/java/org/apache/accumulo/server/master/balancer/GroupBalancer.java?sort=name&dir=ASC&mode=heatmap#xc3a19e02e617cd65:1)

This seemed like the conservative option, instead of changing the integers to longs in the thrift class, [Compacting](https://github.com/apache/accumulo/blob/692fec6494a8348517aea469fb95563700ab7557/core/src/main/java/org/apache/accumulo/core/master/thrift/Compacting.java#L36) but that is another option.


core/src/main/java/org/apache/accumulo/core/util/MapCounterInt.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private HashMap<KT,MutableInt> map;
+
+  public MapCounterInt() {
+    map = new HashMap<>();
+  }
+
+  public int increment(KT key, int i) {
+    MutableInt mutableInt = map.get(key);
```
keith-turner: Could update this to use compute if absent.

core/src/main/java/org/apache/accumulo/core/util/MapCounterInt.java
```diff
@@ -0,0 +1,85 @@
...
+
+    return mutableInt.i;
+  }
+
+  public Set<KT> keySet() {
+    return map.keySet();
+  }
+
+  public Collection<Integer> values() {
+    Collection<MutableInt> vals = map.values();
```
keith-turner: This method could be update to return `Collections2.transform(map.values(), mi->mi.i)`

milleruntime: Looking at the javadoc for that Guava transform I noticed this...
"When a live view is <i>not</i> needed, it may be faster to copy the transformed collection and use the copy."

Which makes me wonder if what we have is faster.  Or is that just saying to make a copy of what is returned?

keith-turner:
Another possible solution to this problem is to add a `int getInt(KT key)` to the existing MapCounter class.  This method could safely convert a long to int throwing an exception if long does not fit.

core/src/main/java/org/apache/accumulo/core/util/MapCounterInt.java
keith-turner: I would stick with what is there after reading that comment.

keith-turner:
Could possibly return a LongStream instead of Collection<Long>.  Then could user could use `LongStream.mapToInt()` with `Math.toIntExact()` to safely convert a long stream to an int stream.

PR #999, opened 2019-02-28
fixes #998 handle splitting tablet with no data and files

### LINKED ISSUE #998: Attempting to split tablet that has files with no data fails
Under some circumstances a  tablet can end up with files that do not contain data for the tablet.   If this happens and the tablet ever tries to split then it will fail.

The failure happens at the following point.

https://github.com/apache/accumulo/blob/rel/1.9.2/server/base/src/main/java/org/apache/accumulo/server/util/FileUtil.java#L335

keith-turner:


server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
```diff
@@ -1635,6 +1635,17 @@ private SplitRowSpec findSplitRow(Collection<FileRef> files) {
       return null;
     }
 
+    if (keys.isEmpty()) {
+      log.info("Cannot split tablet " + extent + ", files contain no data for tablet.");
+
+      // set the following to keep tablet from attempting to split until the tablets set of files
+      // changes.
+      sawBigRow = true;
+      timeOfLastMinCWhenBigFreakinRowWasSeen = lastMinorCompactionFinishTime;
```
mikewalch: This could be another PR but might want to come up with better naming for these variables given they are being used for this case.

keith-turner: I will do that as a follow on change.  Do not want them in the same commit.

keith-turner: #1016

PR #1019, opened 2019-03-08
 #1013 Fix ConfigurableMacBase

mikewalch:
* Return client properties from MiniAccumuloCluster
* SSL configuration was not being returned by MiniAccumuloCluster
  causing SSL ITs to fail


mikewalch:
This should fix integration tests taking 6 vs 4 hours

PR #1026, opened 2019-03-12
Fixes #930: Change Default Compaction

### LINKED ISSUE #930: Default compaction settings are sub-optimal
The defaults for max files per tablet and max files per compaction are 15 and 10.  While testing 2.0.0-alpha-2 I noticed that this lead to sub-optimal compaction behavior.  I think it would be better if both were 15.

Manno15:
Changes default max files per compaction to 15 to match table.file.max


mjwall:
@keith-turner are you working on this?  Should @Manno15 close this PR?

keith-turner:
After working on #1036 I think it may be best to leave the settings as is.  The changes I suggested in #930 were a band-aid for the real problem that I was seeing which is fixed by #1036. It helps to have a buffer between max files per compaction and max files inorder to prevent worst case performance for merging minor compactions.

keith-turner:
I opened #1071 to expand on the issue I mentioned about merging minor compactions.  I think it would be best to close this issue w/o merging.  Sorry for the delay and thanks for patch @Manno15

PR #1036, opened 2019-03-15
fixes #1033 optimize default compaction strategy

### LINKED ISSUE #1033: Default Compaction strategy is sub-optimal
Consider a tablet with the following files.  If the compaction ratio is 3 then all files would meet the criteria for compaction.  However if the max files to compact is 10, then the files C4 and F[5-d] will be selected for compaction.  This is very suboptimal over time.  It would be much better if a subset of files that met the compaction ratio criteria were returned.  For example C[2-4] and F[5-b] could be selected, which is 10 files that meet the ratio criteria.  Another possibility is only selecting only the F files, which meet the criteria and is less than max files. 

| File | Size |
|------|-------|
| C1 | 100M |
| C2 | 100M |
| C3 | 100M |
| C4 | 100M |
| F5 | 1M |
| F6 | 1M |
| F7 | 1M |
| F8 | 1M |
| F9 | 1M |
| Fa | 1M |
| Fb | 1M |
| Fc | 1M |
| Fd | 1M |

keith-turner:


keith-turner:
Below are the results of running a simulation of the old and new compaction strategy.  For the simulation N 1K files were added in a loop 1000 times.  After the files were added the tablet was compacted.

Files Added between compactions | Total data added | Old code rewrite | New code rewrite  | Relative work (new rewrite/old rewrite)
----------------|-----|----------------|----------------|------
1 | 1,000,000 | 4,412,000 | 4,416,000 | 1.00
2 | 2,000,000 | 9,184,000 | 9,184,000 | 1.00
3 | 3,000,000 | 15,240,000 | 15,240,000 | 1.00
4 | 4,000,000 | 19,336,000 | 21,304,000 | 1.10
5 | 5,000,000 | 27,040,000 | 25,495,000 | 0.94
6 | 6,000,000 | 29,250,000 | 33,984,000 | 1.16
7 | 7,000,000 | 37,499,000 | 41,874,000 | 1.12
8 | 8,000,000 | 58,576,000 | 31,507,000| 0.54
9 | 9,000,000 | 1,129,500,000 | 34,936,000 | 0.03

The old strategy performed slightly better for adding 4,6, and 7 files between compactions.  I don't have a solid understanding of this and need to dig into this some more.   The case of adding 9 files between compaction is significantly better, the updated strategy only does 3% of the work done by the current strategy.  This is the case that motivated this change.

keith-turner:
I tweaked the similuation a bit.  The simuation does a single compaction after adding files. I changed it to run compactions while needed at the end.  I also tracked the number of files per tablet during the simulation and found that the new code has less files on average.  Having less files on average offsets  the increased amount of work.

 Files added between compactions | Old rewrite | Old final file count | Old avg files 
----------|----|---|----
1 | 4,412,000 |  8 |  6.30
2 |  9,184,000 |  5  | 6.38
3 |  15,240,000 |  6 |  6.35
4 |  19,336,000 |  7  | 6.32
5 |  27,040,000 |  6  | 6.31
6 |  29,250,000 |  9  | 6.37
7 |  37,499,000 |  9  | 6.34
8 |  58,576,000 |  6  | 6.43
9 |  1,138,500,000 |  1  | 6.16

 Files added between compactions | New rewrite | New final file count | New avg files
----------|----|---|----
1 | 4,416,000 | 6 | 5.97
2 | 9,184,000 | 5 | 6.21
3 | 15,240,000 | 6 | 6.13
4 | 21,304,000 | 5 | 6.09
5 | 25,495,000 | 10 | 6.07
6 | 33,984,000 | 8 | 6.04
7 | 41,874,000 | 8 | 6.03
8 | 39,684,000 | 1 | 6.99
9 | 46,295,000 | 1 | 7.86

PR #1046, opened 2019-03-19
Updating master metrics to provide FATE operational metrics.

EdColeman:
This pull request if for comments to approach - the code should be functionally complete, but additional work on unit and IT test is required before final merge.  

The metrics included are:

- number of current fate transactions when metric is sampled.
the number of changes to the FATE parent node in Zookeeper - this provides a measure of transaction progress between samples.
number of zookeeper errors occurring when sampling to help diagnose zookeeper issues if metrics are not responding.

Also some other IT tests can be refactored to reduce code duplication creating tests that require a long running transaction.


milleruntime:
@EdColeman Looks good to me.  I don't know if you need the IT and Test sink since I am not convinced the metrics work properly with mini.  Also FYI Travis caught some javadoc formatting.

server/master/src/main/java/org/apache/accumulo/master/metrics/fate/FateMetricValues.java
```diff
@@ -0,0 +1,200 @@
...
+   * Get the current FATE metrics values, initial values are set to 0. In the event of errors, the
+   * values may not be fully populated and the default values (0) will be returned - this form
+   * should be avoided for metric counters that expect values will be monotonically increasing - use
+   * with gauges instead.
+   *
+   * @param instance
+   *          Accumulo client instance.
+   * @return populated metrics values
+   */
+  static FateMetricValues updateFromZookeeper(final Instance instance) {
```
keith-turner: Could (meaning feel free to ignore) move this to the builder. Like :

```java
fmv = FateMetricValues.builder().fromZookeeper(instance).copying(otherFMV).build();
```

server/master/src/main/java/org/apache/accumulo/master/metrics/fate/FateMetricValues.java
```diff
@@ -0,0 +1,200 @@
...
+
+      if (log.isTraceEnabled()) {
+        log.trace(
+            "ZkNodeStat: {czxid: {}, mzxid: {}, pzxid: {}, ctime: {}, mtime: {}, "
+                + "version: {}, cversion: {}, num children: {}",
+            node.getCzxid(), node.getMzxid(), node.getPzxid(), node.getCtime(), node.getMtime(),
+            node.getVersion(), node.getCversion(), node.getNumChildren());
+      }
+    } catch (KeeperException ex) {
+      builder.incrZkConnectionErrors();
```
keith-turner: I like how this increment a metric.  Could possibly log, maybe at debug.

server/master/src/main/java/org/apache/accumulo/master/metrics/fate/FateMetrics.java
```diff
@@ -0,0 +1,182 @@
...
+
+  private volatile boolean enabled = false;
+
+  public FateMetrics(final Instance instance, final long minimumRefreshDelay) {
+
+    this.instance = instance;
+
+    this.minimumRefreshDelay = Math.max(DEFAULT_MIN_REFRESH_DELAY, minimumRefreshDelay);
+
+    // instance = master.getInstance();
```
keith-turner: delete

server/master/src/main/java/org/apache/accumulo/master/metrics/fate/FateMetricValues.java
```diff
@@ -0,0 +1,200 @@
...
+   * preserves previous values in case an error or exception prevents the values from being
+   * completely populated, this form may be more suitable for metric counters.
+   *
+   * @param instance
+   *          Accumulo client instance
+   * @param currentValues
+   *          the current fate metrics used as default
+   * @return populated metrics values
+   */
+  static FateMetricValues updateFromZookeeper(final Instance instance,
```
mikewalch: Instance is deprecated in 2.0. It's OK here but when you merge this into master you'll probably need to change this.  You might want to pass the Master instead.

PR #1048, opened 2019-03-21
Possible fix and unit test. Fixes #961

### LINKED ISSUE #961: Log recovery can not succeed after failed marker written in recovery directory
Accumulo version 1.8.1.  When tserver log recovery fails in LogSorter.sort() and a FailedMarkerPath (failed) file is written to the recovery directory, subsequent recoveries fail due to the presence of this file.  In subsequent recovery attempts, when the subdirectories of the recovery root directory are iteratated over in the RecoveryLogReader constructor, the failed marker is interpreted as any other partial results directory and an exception is generated when the underlying data can not be read.  This happens even when a finished marker exists (both were present with the same timestamp - another issue?) because failed comes before finished alphabetically.

Copied from 

https://issues.apache.org/jira/browse/ACCUMULO-4861

milleruntime:
One possible fix for this issue.  This just ignores the failed file.  AFAIK we aren't actually doing anything with this failed file so it could just be removed.  But maybe that be done for 2.0

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogReader.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogReader.java
@@ -127,7 +127,7 @@ public class RecoveryLogReader implements CloseableIterator<Entry<LogFileKey,Log
         foundFinish = true;
         continue;
       }
-      if (SortedLogState.isFailed(child.getPath().getName())) {
+      if (SortedLogState.FAILED.getMarker().equals(child.getPath().getName())) {
         continue;
       }
       FileSystem ns = fs.getVolumeByPath(child.getPath()).getFileSystem();
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogReader.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogReader.java
@@ -127,6 +127,9 @@ public class RecoveryLogReader implements CloseableIterator<Entry<LogFileKey,Log
         foundFinish = true;
         continue;
       }
+      if (SortedLogState.isFailed(child.getPath().getName())) {
+        continue;
+      }
       FileSystem ns = fs.getVolumeByPath(child.getPath()).getFileSystem();
       heap.add(new Index(new Reader(ns.makeQualified(child.getPath()), ns.getConf())));
     }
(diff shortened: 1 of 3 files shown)


milleruntime:
> Should we at least log a debug message?

It already logs the original error.  I am not sure how much value another log message adds...

server/tserver/src/test/java/org/apache/accumulo/tserver/log/RecoveryLogsReaderTest.java
```diff
@@ -177,4 +178,26 @@ public void testSortCheck() {
...
   }
 
+  /**
+   * Test a failed marker doesn't cause issues. See Github issue
+   * https://github.com/apache/accumulo/issues/961
+   */
+  @Test
+  public void testFailed() throws Exception {
+    Path manyMaps = new Path("file://" + root.getRoot().getAbsolutePath() + "/manyMaps");
+    fs.create(new Path(manyMaps, SortedLogState.FAILED.getMarker())).close();
```
keith-turner: Will other test see this file?

server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogReader.java
```diff
@@ -127,6 +127,9 @@ public RecoveryLogReader(VolumeManager fs, Path directory, LogFileKey start, Log
         foundFinish = true;
         continue;
       }
+      if (SortedLogState.FAILED.getMarker().equals(child.getPath().getName())) {
```
keith-turner: Could add this condition as an OR to the first if stmt in the for loop.

server/tserver/src/test/java/org/apache/accumulo/tserver/log/RecoveryLogsReaderTest.java
milleruntime: That is a good question.  I will check

server/tserver/src/main/java/org/apache/accumulo/tserver/log/RecoveryLogReader.java
milleruntime: I was thinking this is easier to remove, if that's what we plan on for 2.0.

server/tserver/src/test/java/org/apache/accumulo/tserver/log/RecoveryLogsReaderTest.java
milleruntime: I added a delete at the end of the Test

PR #1063, opened 2019-03-28
Move metrics registration to master run.

EdColeman:
Addresses issue 1058 - the registration of metrics should be independent of replication.


milleruntime:
@EdColeman would you like me to merge this?

PR #1066, opened 2019-03-29
Speed up containKey calls by not always recomputing hashcodes

jzgithub1:
By only recomputing hash codes when the underlying object data changes may speed-up some operations like containsKey that Map.get call many times.  I implemented on ByteSequence and its subclasses and Key.  Maybe we get a performance boost.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/data/Key.java
+++ b/core/src/main/java/org/apache/accumulo/core/data/Key.java
@@ -48,6 +48,8 @@ public class Key implements WritableComparable<Key>, Cloneable {
   protected byte[] colVisibility;
   protected long timestamp;
   protected boolean deleted;
+  private boolean staleHashCode = true;
+  private int hashCode = 0;
 
   /**
    * Create a {@link Key} builder.
@@ -863,6 +865,7 @@ public class Key implements WritableComparable<Key>, Cloneable {
     colVisibility = k.colVisibility;
     timestamp = k.timestamp;
     deleted = k.deleted;
+    staleHashCode = true;
 
   }
 
@@ -888,6 +891,7 @@ public class Key implements WritableComparable<Key>, Cloneable {
 
     timestamp = WritableUtils.readVLong(in);
     deleted = in.readBoolean();
+    staleHashCode = true;
   }
 
   @Override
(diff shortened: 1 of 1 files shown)


mjwall:
> In general, this is a good idea, but the apparently immutable ArrayByteSequence isn't truly immutable. The underlying array can still be changed outside of the class.
> 
> It's probably more worth spending time on getting https://github.com/apache/fluo-bytes reviewed, polished, and released, so we can have a truly immutable byte sequence representation in a highly stable API.

Recommend closing and looking at https://github.com/apache/fluo-bytes as @ctubbsii suggests.

mjwall:
May want to look at https://github.com/OpenHFT/Chronicle-Bytes as well.

jzgithub1:
Thanks for information about other projects that already address this problem.  I addressed the issue of the underlying data being modified and mark the hashcode as 'stale' so  the hashcode gets updated by a subsequent call to the hashcode function in this pull request. 'mvn test' passes and I think 'mvn verify' will too.  I assume that you reviewed the code and found valid deficiencies in the solution so I will accept your conclusions.

I will consider this issue closed and work on something else.

keith-turner:
A situation like the following could be problematic because the `toArray()` function may or may not copy.  If it does not copy, then the following code changes the byte sequence but the hashcode does not change.

```java
ByteSequence byteSeq = new ArrayByteSequence(new byte[]{8,9});
System.out.println(byteSeq.hashCode());
byteSeq.toArray()[0]=3;
System.out.println(byteSeq.hashCode());
```

jzgithub1:
Thanks for taking your time to show me that example Keith. I have not found that use case in my search of the Accumulo code but the possibility of someone doing that can't be ignored.  If I revert the modifications I made on the ArrayByteSequence, ByteSequence,  and MutableByteSequence then will the modifications to the Key class be helpful.  I can see in the RFileClientTest that recomputing of hashcodes will be prevented with the changes I made to the Key class.  MemKey extends from Key so the speed gain in key search can be gained in MemKey lookups.  This is a small issue and I know we are trying to get a release out, so I don't want to take up too much of anybody's cycles on this.   I have sped up some comparison code in the past doing something like this so I am just trying to do the same thing here.

ivakegg:
The basic problem here is that the underlying byte array can be modified outside of the  Key.  Unfortunately I think this is too dangerous to implement and will bite somebody down the road.

PR #1073, opened 2019-04-03
Move similar tests into PermissionsIT

milleruntime:
* Move similar tests into PermissionsIT, dropping ArbitraryTablePropertiesIT


test/src/main/java/org/apache/accumulo/test/functional/PermissionsIT.java
```diff
@@ -735,6 +738,7 @@ private void testGrantedTablePermission(AccumuloClient test_user_client, TablePe
       case ALTER_TABLE:
         Map<String,Set<Text>> groups = new HashMap<>();
         groups.put("tgroup", new HashSet<>(Arrays.asList(new Text("t1"), new Text("t2"))));
+        testArbitraryProperty(test_user_client, tableName, true);
```
keith-turner: Unrelated to your change, but it seems nothing is done with `groups`

test/src/main/java/org/apache/accumulo/test/functional/PermissionsIT.java
```diff
@@ -780,4 +784,54 @@ private void verifyHasNoTablePermissions(AccumuloClient root_client, String user
...
+      // Loop through properties to make sure the new property is added to the list
+      count = 0;
+      for (Entry<String,String> property : c.tableOperations().getProperties(tableName)) {
+        if (property.getKey().equals(propertyName) && property.getValue().equals(description2))
+          count++;
+      }
+      assertEquals(count, 1);
+
+      // Remove the property and make sure there is no longer a value associated with it
+      c.tableOperations().removeProperty(tableName, propertyName);
```
keith-turner: Ideally removeProperty would be tested when the user does not have permission.  If the call to setProp fails because the user does not have permission, the exception will prevent this code from executing.  I don't think the exiting test checked this either.  So this should not hold up progress, this change still seems like an overall improvement.

milleruntime: Yeah I see what you are saying.  The testMissingSystemPermission() method in this class does test that removeProperty throws a AccumuloSecurityException.  [Here](https://github.com/milleruntime/accumulo/blob/9292f715adaa69748696e115aa2b6246c32218aa/test/src/main/java/org/apache/accumulo/test/functional/PermissionsIT.java#L204)

PR #1082, opened 2019-04-09
Improve FateConcurrencyIT to reduce transient test errors (Issue #1079).

EdColeman:
Test previously looked for a compaction and then tried to match that with a table id. This
improves the check so that it also considers the table when checking for compaction has started.
This is to reduce transient errors reported in #1079.

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
@@ -90,7 +90,9 @@ public class FateConcurrencyIT extends AccumuloClusterHarness {
 
   private String secret;
 
-  // when true, multiple tables, multiple compactions will be used during run.
+  // Test development only. When true, multiple tables, multiple compactions will be
+  // used during the test run which simulates transient condition that was causing
+  // the test to fail..
   private boolean runMultipleCompactions = false;
 
   @Before
@@ -207,7 +209,8 @@ public class FateConcurrencyIT extends AccumuloClusterHarness {
     Instance instance = connector.getInstance();
     String tableId;
 
-    // development testing - force multiple compactions to see that it's handled.
+    // for development testing - force transient condition that was failing this test so that
+    // we know if multiple compactions are running, they are properly handled by the test code.
     if (runMultipleCompactions) {
       runMultipleCompactions();
     }
@@ -309,10 +312,10 @@ public class FateConcurrencyIT extends AccumuloClusterHarness {
   }
 
   /**
-   * Helper method for testing findFate refactor - creates multiple tables and launches compactions
-   * so that more that 1 FATE compaction operation is running during the test. This forces the
-   * condition where multiple compactions could be running - but it is not necessary for the test
-   * itself.
+   * This method was helpful for debugging a condition that was causing transient test failures.
+   * This forces a condition that the test should be able to handle. This method is not needed
+   * during normal testing, it was kept to aid future test development / troubleshooting if
+   * other transient failures occur.
    */
   private void runMultipleCompactions() {
 
--- a/test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
@@ -90,6 +90,9 @@ public class FateConcurrencyIT extends AccumuloClusterHarness {
 
   private String secret;
 
+  // when true, multiple tables, multiple compactions will be used during run.
+  private boolean runMultipleCompactions = false;
+
   @Before
   public void setup() {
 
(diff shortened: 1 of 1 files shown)


test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
```diff
@@ -204,6 +207,8 @@ public void getFateStatus() {
     Instance instance = connector.getInstance();
     String tableId;
 
+    runMultipleCompactions();
```
ivakegg: why are we running the multiple compactions here.  I think this should replace the starting of a compaction on line 226 or should not be called at all.

test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
```diff
@@ -220,7 +225,15 @@ public void getFateStatus() {
 
     Future<?> compactTask = startCompactTask();
 
-    assertTrue("compaction fate transaction exits", findFate(tableName));
+    try {
+
+      assertTrue("compaction fate transaction exits", findFate(tableName));
```
ivakegg: The blockUntilCompactionsRunning() essentially does this test.  Perhaps that should return a boolean as to whether a compaction was found and that should be tested instead of depending on the timing of the compaction start on 226.

test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
ivakegg: Ah, I see that this was disabled by the boolean, and this was only to force the error.  At least add a comment to that effect here.

test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
EdColeman: I'll incorporate both suggestions and push changes in near future.

test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
```diff
@@ -204,6 +207,11 @@ public void getFateStatus() {
     Instance instance = connector.getInstance();
     String tableId;
 
+    // development testing - force multiple compactions to see that it's handled.
+    if (runMultipleCompactions) {
+      runMultipleCompactions();
+    }
+
```
milleruntime: If it is a more reliable test when running more than one compaction, I say just have it always do that.

milleruntime: Nevermind, I see what you are doing... maybe just update the comment to make it more clear.  Maybe use the word "debug" somewhere so its clear this is just to help debug a test failure.

test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
```diff
@@ -302,6 +308,30 @@ public void getFateStatus() {
     }
   }
 
+  /**
+   * Helper method for testing findFate refactor - creates multiple tables and launches compactions
+   * so that more that 1 FATE compaction operation is running during the test. This forces the
+   * condition where multiple compactions could be running - but it is not necessary for the test
+   * itself.
+   */
+  private void runMultipleCompactions() {
```
milleruntime: Could just say "Helper method for debugging possible test failure"

test/src/main/java/org/apache/accumulo/test/functional/FateConcurrencyIT.java
```diff
@@ -322,25 +352,38 @@ public void getFateStatus() {
...
+          // Validate that there is a compaction fate transaction - otherwise test is invalid.
+          if (findFate(tableName)) {
+            return true;
+          }
+        }
+
       } catch (AccumuloSecurityException | AccumuloException ex) {
         throw new IllegalStateException("failed to get active compactions, test fails.", ex);
+      } catch (KeeperException ex) {
+        log.trace("Saw possible transient zookeeper error");
```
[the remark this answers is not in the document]
milleruntime: Or maybe info... just saying that this is OK but I did see this exception.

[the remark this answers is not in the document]
EdColeman: This is not an abnormal condition - it's just the way that zookeeper works and that it throws an exception to signal that a zk node no longer exists - sometimes, that could be an error if code was depending on something that should be there, other times, it's just that the node was removed by other threads or processes as a normal operation.  The logging at trace was so that if anyone was specifically interested logging could be enabled, but normally it's just zookeeper being zookeeper - the major change was to stop failing the test if this "normal", but transient  condition occurred, during the test.

PR #1089, opened 2019-04-15
Remove Major compaction from Bulk Import

milleruntime:
I pulled this from [ACCUMULO-3944](https://issues.apache.org/jira/browse/ACCUMULO-3944) but not sure if it should be done or not.


milleruntime:
After looking more into this, I don't think this really does anything for improving ingest performance.  And there is a chance that it may hurt scan performance, even stopping scans if too many files are open.

PR #1095, opened 2019-04-16
Update native map makefile to work with Java 11.

keith-turner:
The native map makefile checks to see if javah exists inorder to
determine if java is a JDK or JRE.  In Java 11, javah no longer exists.
The makefile was modfied to check if javac exists, which is a JDK
command that exists in Java 8 and 11.


keith-turner:
> Can you merge this to 1.9 also?

Sure

keith-turner:
Merged in 75a87b573f6ab6ec23511338d9abe193953bdbfa

PR #1100, opened 2019-04-17
Avoid locking in tserver when reading online tablets

keith-turner:
Profiling Accumulo during running a performance test showed lots of lock
contention for the map of online tablets.  Most operations need to read
this map.  This change creates a read only snapshot of the online
tablets that is used to avoid lock contention.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
@@ -30,7 +30,6 @@ import java.security.PrivilegedExceptionAction;
 import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


keith-turner:
> This is cool. Did you see performance improvements avoiding all these locks?

A little bit.  What I find is that when I get rid of one lock that is causing contention it allows another one to shine.  There are still other cases of lock contention that I saw in the profiling data.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -367,15 +365,13 @@ public TabletServer(ServerContext context) {
...
-      synchronized (onlineTablets) {
-        long now = System.currentTimeMillis();
-        for (Tablet tablet : onlineTablets.values())
-          try {
-            tablet.updateRates(now);
-          } catch (Exception ex) {
-            log.error("Error updating rates for {}", tablet.getExtent(), ex);
-          }
-      }
+      long now = System.currentTimeMillis();
```
milleruntime: Will dropping the sychronized block in the simple timer cause problems?  I am not sure how much of the SimpleTimer is shared and what we should worry about.

keith-turner: I don't think it will cause any problems.  It was sync before to avoid concurrent mod exceptions.  Now it will iterate over the immutable snapshot.

PR #1107, opened 2019-04-18
Fix ConcurrentModificationException in HostRegexTableLoadBalancer

adamjshook:


server/base/src/main/java/org/apache/accumulo/server/master/balancer/HostRegexTableLoadBalancer.java
```diff
@@ -226,10 +227,10 @@ protected void parseConfiguration(ServerConfiguration conf) {
     if (null == t) {
       throw new RuntimeException("Table Operations cannot be null");
     }
-    tableIdToTableName = new HashMap<>();
-    poolNameToRegexPattern = new HashMap<>();
+    Map<String,String> tableIdToTableNameBuilder = new HashMap<>();
```
keith-turner: Could use an ImmutableMap.Builder

server/base/src/main/java/org/apache/accumulo/server/master/balancer/HostRegexTableLoadBalancer.java
```diff
@@ -244,11 +245,15 @@ protected void parseConfiguration(ServerConfiguration conf) {
...
             String tableName = customProp.getKey().substring(HOST_BALANCER_PREFIX.length());
             String regex = customProp.getValue();
-            poolNameToRegexPattern.put(tableName, Pattern.compile(regex));
+            poolNameToRegexPatternBuilder.put(tableName, Pattern.compile(regex));
           }
         }
       }
     }
+
+    tableIdToTableName = ImmutableMap.copyOf(tableIdToTableNameBuilder);
```
keith-turner: Not completely sure, but I am wondering if tableIdToTableName and poolNameToRegexPattern should be made volatile.  Before this change the reference to the map never changed.  Now the ref is being changed to a new map.  Making it volatile would ensure other threads see the new map.

server/base/src/main/java/org/apache/accumulo/server/master/balancer/HostRegexTableLoadBalancer.java
adamjshook: I tried that originally, but it contains nulls?  Or maybe it was duplicate keys.  I patched this a while ago but failed to contribute it back at the time.

server/base/src/main/java/org/apache/accumulo/server/master/balancer/HostRegexTableLoadBalancer.java
adamjshook: Good catch; I will update it.

server/base/src/main/java/org/apache/accumulo/server/master/balancer/HostRegexTableLoadBalancer.java
keith-turner: I have run into those issues with Guava's immutable code before.  Your change seems like the best approach with those issues.

adamjshook:
Thank you for the review @keith-turner!

adamjshook:
@keith-turner Remind me what protocol here is in moving it upstream to `master`? It doesn't directly apply -- requires a minor modification to key by `TableId` instead of `String`.  Should I open another PR? Or cherry pick, update, and push straight to `master` with this PR number in the commit?

keith-turner:
@adamjshook need to merge 1.9 into master and resolve conflicts.  I was actually working through that.  Would you like to do it?  Your commit and another commit needed to be merged and both had conflicts.

adamjshook:
@keith-turner Ah, since you're already doing it, then please continue.  Thanks for the help.

keith-turner:
I do not do PR for the merge commit.  Adding the PR issue number(s) to the merge commit would be nice.  I sometimes do this.

keith-turner:
These are the changes before and after my merge.  Do these look good? 

https://github.com/apache/accumulo/compare/18320ad..1a373f5

adamjshook:
Looks good to me!

PR #1135, opened 2019-04-28
Issue 1129 fix grep pn option

DonResnik:
Here are the commands I ran to test this:
root@uno>  createtable test
root@uno>  insert row1 colf colq value
root@uno>  insert row1 colf1 colq value
root@uno>  insert row2 colf colq value
root@uno>  insert row2 colf1 colq value
root@uno test> scan
row1 colf:colq1 []    value
row1 colf1:colq1 []    value
row2 colf:colq1 []    value
row2 colf1:colq1 []    value
root@uno test> setshelliter -class org.apache.accumulo.core.iterators.ColumnFamilyCounter -p 100 -pn testIter
[skipped the name and name-value setting options]
root@uno test> scan -pn testIter
row1 colf: []    1
row1 colf1: []    1
row2 colf: []    1
row2 colf1: []    1
root@uno test> grep row1 -pn testIter
row1 colf: []    1
row1 colf1: []    1
root@uno test> egrep .*row1.* -pn testIter
row1 colf: []    1
row1 colf1: []  1


DonResnik:
I rebased and squashed to update the bad commit comments.  They are still in the history here, but they are not on the branch.  Please let me know if I need to do anything else.

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
```diff
@@ -114,8 +144,11 @@ public Options getOptions() {
     final Options opts = super.getOptions();
     numThreadsOpt = new Option("nt", "num-threads", true, "number of threads to use");
     negateOpt = new Option("v", "negate", false, "only include rows without search term");
+    profileNameOpt = new Option("pn", "profile-name", true,
```
keith-turner: Not completely sure, but I think the earlier call to `super.getOptions();` may already add a profile option.

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
```diff
@@ -93,10 +102,31 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
   }
 
   protected void setUpIterator(final int prio, final String name, final String term,
```
keith-turner: This method is called for each grep term.  I don't think it needs to resetup the profile iters for each term.

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
```diff
@@ -63,6 +66,12 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
       if (cl.hasOption(negateOpt.getOpt())) {
         negate = true;
       }
+
```
keith-turner: The super class `ScanCommand` has a protected method called `addScanIterators()`.  I am not completely sure, but I think you may be able to call it to setup the profile iters.  If so, very little changes would be needed for this class.

DonResnik: Thanks for the suggestion.  I had to make a small change to addScanIterators signature to accept ScannerBase instead of Scanner, but it removed most of my code update, so I hope you are ok with the change.

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
DonResnik: yes, code updated to remove this

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
DonResnik: agreed, code updated

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
DonResnik: Also, I re-ran the test I documented in the ticket and I got the same results with this new update.

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
```diff
@@ -71,10 +71,11 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
       scanner.setTimeout(getTimeout(cl), TimeUnit.MILLISECONDS);
 
       setupSampling(tableName, cl, shellState, scanner);
+      addScanIterators(shellState, cl, scanner, "");
 
       for (int i = 0; i < cl.getArgs().length; i++) {
         setUpIterator(Integer.MAX_VALUE - cl.getArgs().length + i, "grep" + i, cl.getArgs()[i],
-            scanner, cl, negate);
+            scanner, cl, negate, shellState);
```
keith-turner: `shellState` is newly passed, but it seems like its not used.

shell/src/main/java/org/apache/accumulo/shell/commands/EGrepCommand.java
```diff
@@ -31,7 +30,7 @@
 
   @Override
   protected void setUpIterator(final int prio, final String name, final String term,
-      final BatchScanner scanner, CommandLine cl, boolean negate) throws IOException {
+      final BatchScanner scanner, CommandLine cl, boolean negate, final Shell shellState) throws Exception {
```
keith-turner: Seems like `shellState` is not used.

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
```diff
@@ -71,10 +71,11 @@ public int execute(final String fullCommand, final CommandLine cl, final Shell s
       scanner.setTimeout(getTimeout(cl), TimeUnit.MILLISECONDS);
 
       setupSampling(tableName, cl, shellState, scanner);
+      addScanIterators(shellState, cl, scanner, "");
```
keith-turner: Does this change also benefit the EgrepCommand?

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
DonResnik: correct, removed shellState from method signature

shell/src/main/java/org/apache/accumulo/shell/commands/EGrepCommand.java
DonResnik: correct again, removed shellState from method signature

shell/src/main/java/org/apache/accumulo/shell/commands/GrepCommand.java
DonResnik: Yes, ergrep uses grep execute() method

keith-turner:
Looks good, thanks @DonResnik

PR #1137, opened 2019-04-29
Remove Simulation Mode from MR API

milleruntime:
Taken from #1132


hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoopImpl/mapreduce/OutputFormatBuilderImpl.java
```diff
@@ -67,12 +67,6 @@ public OutputFormatBuilderImpl(Class<?> callingClass) {
     return this;
   }
 
-  @Override
-  public OutputFormatBuilder.OutputOptions<T> simulationMode(boolean value) {
-    this.simulationMode = value;
```
mikewalch: Is this read anywhere?

milleruntime: Nope!

PR #1149, opened 2019-05-07
Drop Server Activity page from Monitor

milleruntime:


milleruntime:
> I was just telling @keith-turner yesterday that I thought we should drop this. +1
> Was there any REST endpoints that were used only by this that could also be dropped?

I didn't find any but I feel like there should be?  I will look around some more.

milleruntime:
I found one in WebViews 64aa2c4dca0fcf0a6c99016b49df9a12c2ce6fc0.  I think that is all.

PR #1155, opened 2019-05-09
fix #405 - Utilize newTableConfig in ITs

### LINKED ISSUE #405: Utilize NewTableConfiguration in ITs
There are places in our ITs that I think could use NewTableConfiguration when tables are created.  This should help make the tests more stable. 

- MiniAccumuloClusterTest
- LargeRowIT
- TabletIT
- SpitIT
- RegexGroupBalanceIT
- SessionDurabilityIT
- ConfigurableCompactionIT
- DurabilityIT

Not sure about the Replication tests... If you have to create both tables first, then set the table replication property?
- ReplicationIT
- UnorderedWorkAssignerReplicationIT
- CyclicReplicationIT 
- KerberosReplicationIT
- MultiInstanceReplicationIT

hkeebler:
Updated suggested IT tests with NewTableConfiguration and consolidated most table property setting on the table initialization where applicable.  Verified all tests except for the KerberosReplicationIT.  MiniAccumuloClusterTest and CyclicReplicationIT were already utilizing the NewTableConfiguration so changes there.

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/functional/ConfigurableCompactionIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/functional/ConfigurableCompactionIT.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.security.SecureRandom;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Random;
 import java.util.TreeSet;
@@ -32,6 +33,7 @@ import org.apache.accumulo.core.client.Accumulo;
 import org.apache.accumulo.core.client.AccumuloClient;
 import org.apache.accumulo.core.client.BatchWriter;
 import org.apache.accumulo.core.client.Scanner;
+import org.apache.accumulo.core.client.admin.NewTableConfiguration;
 import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.core.data.Mutation;
 import org.apache.accumulo.core.metadata.MetadataTable;
@@ -93,10 +95,13 @@ public class ConfigurableCompactionIT extends ConfigurableMacBase {
   public void test() throws Exception {
     try (AccumuloClient c = Accumulo.newClient().from(getClientProperties()).build()) {
       final String tableName = getUniqueNames(1)[0];
-      c.tableOperations().create(tableName);
-      c.tableOperations().setProperty(tableName, Property.TABLE_COMPACTION_STRATEGY.getKey(),
-          SimpleCompactionStrategy.class.getName());
+
+      c.tableOperations().create(tableName,
+          new NewTableConfiguration()
+              .setProperties(Collections.singletonMap(Property.TABLE_COMPACTION_STRATEGY.getKey(),
+                  SimpleCompactionStrategy.class.getName())));
       runTest(c, tableName, 3);
+
       c.tableOperations().setProperty(tableName,
           Property.TABLE_COMPACTION_STRATEGY_PREFIX.getKey() + "count", "" + 5);
       runTest(c, tableName, 5);
@@ -109,17 +114,17 @@ public class ConfigurableCompactionIT extends ConfigurableMacBase {
       final String tableName = getUniqueNames(1)[0];
       File destFile =
           installJar(getCluster().getConfig().getAccumuloDir(), "/TestCompactionStrat.jar");
-      c.tableOperations().create(tableName);
       c.instanceOperations().setProperty(
           Property.VFS_CONTEXT_CLASSPATH_PROPERTY.getKey() + "context1", destFile.toString());
-      c.tableOperations().setProperty(tableName, Property.TABLE_MAJC_RATIO.getKey(), "10");
-      c.tableOperations().setProperty(tableName, Property.TABLE_CLASSPATH.getKey(), "context1");
+      Map props = new HashMap<>();
+      props.put(Property.TABLE_MAJC_RATIO.getKey(), "10");
+      props.put(Property.TABLE_CLASSPATH.getKey(), "context1");
       // EfgCompactionStrat will only compact a tablet w/ end row of 'efg'. No other tablets are
       // compacted.
-      c.tableOperations().setProperty(tableName, Property.TABLE_COMPACTION_STRATEGY.getKey(),
+      props.put(Property.TABLE_COMPACTION_STRATEGY.getKey(),
           "org.apache.accumulo.test.EfgCompactionStrat");
-
-      c.tableOperations().addSplits(tableName, new TreeSet<>(Arrays.asList(new Text("efg"))));
+      c.tableOperations().create(tableName, new NewTableConfiguration().setProperties(props)
+          .withSplits(new TreeSet<>(Arrays.asList(new Text("efg")))));
 
       for (char ch = 'a'; ch < 'l'; ch++)
         writeFlush(c, tableName, ch + "");
--- a/test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
@@ -28,6 +28,7 @@ import java.util.Map.Entry;
 import org.apache.accumulo.core.client.Accumulo;
 import org.apache.accumulo.core.client.AccumuloClient;
 import org.apache.accumulo.core.client.BatchWriter;
+import org.apache.accumulo.core.client.admin.NewTableConfiguration;
 import org.apache.accumulo.core.client.admin.TableOperations;
 import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.core.data.Mutation;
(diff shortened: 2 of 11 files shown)


hkeebler:
oops, saw 2 commented code lines left in.  I will remove and wait for other comments.

test/src/main/java/org/apache/accumulo/test/functional/ConfigurableCompactionIT.java
```diff
@@ -109,17 +114,18 @@ public void testPerTableClasspath() throws Exception {
       final String tableName = getUniqueNames(1)[0];
       File destFile =
           installJar(getCluster().getConfig().getAccumuloDir(), "/TestCompactionStrat.jar");
-      c.tableOperations().create(tableName);
+      // c.tableOperations().create(tableName);
```
milleruntime: Should drop commented out code.

test/src/main/java/org/apache/accumulo/test/functional/SessionDurabilityIT.java
```diff
@@ -124,8 +127,9 @@ public void testConditionDurability2() throws Exception {
     try (AccumuloClient c = Accumulo.newClient().from(getClientProperties()).build()) {
       String tableName = getUniqueNames(1)[0];
       // table default is durable writes
-      c.tableOperations().create(tableName);
-      c.tableOperations().setProperty(tableName, Property.TABLE_DURABILITY.getKey(), "none");
+      c.tableOperations().create(tableName, new NewTableConfiguration()
+          .setProperties(Collections.singletonMap(Property.TABLE_DURABILITY.getKey(), "none")));
+      // c.tableOperations().setProperty(tableName, Property.TABLE_DURABILITY.getKey(), "none");
```
milleruntime: Another commented code.

milleruntime:
You can use this command to run your ITs:
mvn clean verify -Dit.test=SplitIT,TabletIT -Dtest=foo -Dspotbugs.skip

test/src/main/java/org/apache/accumulo/test/functional/SessionDurabilityIT.java
```diff
@@ -55,8 +58,8 @@ public void nondurableTableHasDurableWrites() throws Exception {
     try (AccumuloClient c = Accumulo.newClient().from(getClientProperties()).build()) {
       String tableName = getUniqueNames(1)[0];
       // table default has no durability
-      c.tableOperations().create(tableName);
-      c.tableOperations().setProperty(tableName, Property.TABLE_DURABILITY.getKey(), "none");
+      c.tableOperations().create(tableName, new NewTableConfiguration()
+          .setProperties(Collections.singletonMap(Property.TABLE_DURABILITY.getKey(), "none")));
```
keith-turner: Could statically import the singletonMap method to make this a little shorter.

test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
```diff
@@ -88,7 +89,7 @@ private void cleanup(AccumuloClient c, String[] tableNames) throws Exception {
   }
 
   private void createTable(AccumuloClient c, String tableName) throws Exception {
-    c.tableOperations().create(tableName);
+    c.tableOperations().create(tableName, new NewTableConfiguration());
```
keith-turner: Why add the empty `new NewTableConfiguration()`?

hkeebler:
Updated, ran the verify and push to remote branch.  Ready.

test/src/main/java/org/apache/accumulo/test/replication/MultiInstanceReplicationIT.java
```diff
@@ -534,24 +534,18 @@ public void dataWasReplicatedToThePeerWithoutDrain() throws Exception {
...
-      String masterTableId = clientMaster.tableOperations().tableIdMap().get(masterTable);
-      assertNotNull(masterTableId);
-
-      clientPeer.tableOperations().create(peerTable);
+      clientPeer.tableOperations().create(peerTable, new NewTableConfiguration());
       String peerTableId = clientPeer.tableOperations().tableIdMap().get(peerTable);
       assertNotNull(peerTableId);
 
-      // Give our replication user the ability to write to the table
-      clientPeer.securityOperations().grantTablePermission(peerUserName, peerTable,
```
keith-turner: Is this present in the new changes?  If not, was it intentionally deleted?  I got kinda lost in the diff for this file.

milleruntime:
Are all the ITs you changed passing?

test/src/main/java/org/apache/accumulo/test/functional/SessionDurabilityIT.java
hkeebler: You mean
`static final Map<k,v> DURABILITY_NONE = Collections.singletonMap(Property.TABLE_DURABILITY.getKey(), "none");`
...etc.
and then refer to 
```
c.tableOperations().create(tableName, new NewTableConfiguration()
          .setProperties(DURABILITY_NONE);
```
Should I review the others for something similiar where possible?

test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
hkeebler: I believe that was the intent of this ticket.  The NewTableConfiguration does some sort of checking and validating and down the road maybe more default behavior.  BUT I'm game for whatever is desired.

test/src/main/java/org/apache/accumulo/test/replication/MultiInstanceReplicationIT.java
hkeebler: good catch.  I got lost in a few of these too.  The test seemed to run anyway.  I will double check that.

milleruntime:
I ran some of your changes locally and saw 
[ERROR] testWriteSpeed(org.apache.accumulo.test.functional.DurabilityIT)  Time elapsed: 125.991 s  <<< ERROR!
org.junit.runners.model.TestTimedOutException: test timed out after 120000 milliseconds

This probably not be related to your change as we have a few flaky tests and this is one of them.

test/src/main/java/org/apache/accumulo/test/functional/SessionDurabilityIT.java
keith-turner: What you suggested makes sense in some places where the same map is repeated.  I was thinking of the following.

Add a static method import as follows. 

```java
import static java.util.Collections.singletonMap;
```
Then the code could be written as follows, dropping `Collections`

```java
c.tableOperations().create(tableName, new NewTableConfiguration()
   .setProperties(singletonMap(Property.TABLE_DURABILITY.getKey(), "none")));
```

hkeebler:
@milleruntime I believe that this test is not run by default.  Jeff M. found that this test can fail on the 2nd or 3rd test - comparing sync time and flush time.  Depending on the hardware the times can be different than expected. He may be putting a ticket in.

test/src/main/java/org/apache/accumulo/test/functional/SessionDurabilityIT.java
hkeebler: Will do.

test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
keith-turner: `tableOperations().create(tableName)` and `c.tableOperations().create(tableName, new NewTableConfiguration())` should be functionally equivalent.  I suspect the intent of the ticket was to add `NewTableConfiguration` when there were things to set on it.

test/src/main/java/org/apache/accumulo/test/replication/MultiInstanceReplicationIT.java
keith-turner: I am not sure if its needed, I did not look that closely at the test.  I was just looking at the diff for correctness of table names and props.

Manno15:
> I ran some of your changes locally and saw
> [ERROR] testWriteSpeed(org.apache.accumulo.test.functional.DurabilityIT) Time elapsed: 125.991 s <<< ERROR!
> org.junit.runners.model.TestTimedOutException: test timed out after 120000 milliseconds
> 
> This probably not be related to your change as we have a few flaky tests and this is one of them.

@milleruntime Yeah, I have had several issues regarding this test in particular. Generally not with the timeouts though but with the assertTrue statements being false. Sometimes sync was faster than flush, sometimes it was flush faster than log. No real consistent results from my testing but it was either of these two failing. 

I did notice that the older hardware we have here did run it successfully multiple times but our newer machines were the ones being inconsistent. Should this be a separate issue?  I have tried to find  workarounds for it but no real luck. Only thing that seemed to get more consistent results  was returning the first attempt (attempts[0]) rather than the median (attempts[2]) one in the writeSome function. [https://github.com/apache/accumulo/blob/master/test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java#L241](url) That's probably not a viable solution for this test though.

milleruntime:
Feel free to try different things. 

This test has been a thorn for some time so it would be good to fix it up.   I opened a separate ticket for DurabilityIT: https://github.com/apache/accumulo/issues/1159

hkeebler:
I think this is a duplicate -  a later pull request was made #1160 (my goof) and accepted and merged

PR #1162, opened 2019-05-15
Ref #1084 Fixes File not Found Error

Manno15:
The main issue was importDirectory, when running a standalone instance, looked inside the users local filesystem instead of in hdfs which is what was causing this error. My fix for it was using the filesystem's default name key (mine being hdfs://localhost:8020 for example) and adding that to the front of the base path the test was using. This allowed the importDirectory function to have access to the full path so it would look inside hdfs instead of the users local filesystem. 

I added the delete on exit command since this change altered the test when run with MiniAccumuloCluster. The path wouldn't be just the tmp directory like it was previously so it wouldn't properly be deleted. This caused license issues when building. 

In reference to issue #1084

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/ImportExportIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/ImportExportIT.java
@@ -92,9 +92,7 @@ public class ImportExportIT extends AccumuloClusterHarness {
       FileSystem fs = cluster.getFileSystem();
       Path tmp = cluster.getTemporaryPath();
       log.info("Using FileSystem: " + fs);
-      // Add Fs.defaultFS to baseDir so importDirectory has the full path for Standalone Instance
-      String defaultFS = fs.getConf().get(FileSystem.FS_DEFAULT_NAME_KEY);
-      Path baseDir = new Path(defaultFS + tmp, getClass().getName());
+      Path baseDir = new Path(fs.getUri().toString() + tmp, getClass().getName());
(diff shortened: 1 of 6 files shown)


test/src/main/java/org/apache/accumulo/test/ImportExportIT.java
```diff
@@ -92,15 +92,20 @@ public void testExportImportThenScan() throws Exception {
       FileSystem fs = cluster.getFileSystem();
       Path tmp = cluster.getTemporaryPath();
       log.info("Using FileSystem: " + fs);
-      Path baseDir = new Path(tmp, getClass().getName());
+      // Add Fs.defaultFS to baseDir so importDirectory has the full path for Standalone Instance
+      String defaultFS = fs.getConf().get(FileSystem.FS_DEFAULT_NAME_KEY);
```
keith-turner: I am not completely sure, but I think this could possibly be replaced with `fs.getUri().toString()`.

keith-turner: Actually, if that works then could do `Path baseDir = new Path(fs.getUri() + tmp, getClass().getName());`

Manno15: Yeah, you're right. That does work. I could have sworn I tried that already but I guess I was mistaken. Thanks for the help. I will make the changes to the rest of them.

PR #1165, opened 2019-05-21
Fixes #1159. Refactor DurabilityIT

### LINKED ISSUE #1159: Fix DurabilityIT
This IT is a performance IT which means its measuring performance but can fail quite frequently, depending on the situation.  The test testWriteSpeed is particularly flaky and should be rewritten. There are a few things that can be done to fix this IT.

1. Split out the Durability Unit Tests into a separate Test or IT
2. Drop the testWriteSpeed test and drop the performance aspect.

Manno15:
The performance aspect of this IT is being moved to accumulo-testing.

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
@@ -91,32 +91,6 @@ public class DurabilityIT extends ConfigurableMacBase {
     c.tableOperations().create(tableName);
   }
 
-  @Test(timeout = 2 * 60 * 1000)
-  public void testWriteSpeed() throws Exception {
-    try (AccumuloClient client = Accumulo.newClient().from(getClientProperties()).build()) {
-      TableOperations tableOps = client.tableOperations();
-      String[] tableNames = init(client);
-      // write some gunk, delete the table to keep that table from messing with the performance
-      // numbers of successive calls
-      // sync
-      long t0 = writeSome(client, tableNames[0], N);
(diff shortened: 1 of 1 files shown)


milleruntime:
Sorry I didn't know the new test was only a performance Test.

Manno15:
Sorry for the confusion guys. I should have asked for clarification on the issue ticket. I will make the changes when I get in tomorrow.

Manno15:
I removed testWriteSpeed and I also change the writeSome function to void since no other function used the return value.

keith-turner:
I am not sure @Manno15 .  I restarted the build, lets see what happens.

Manno15:
@keith-turner Thanks for restarting it. It looks like it built this time. Not sure what happened the first time.

test/src/main/java/org/apache/accumulo/test/functional/DurabilityIT.java
```diff
@@ -218,7 +192,7 @@ private void restartTServer() throws Exception {
     cluster.start();
   }
 
-  private long writeSome(AccumuloClient c, String table, long count) throws Exception {
+  private void writeSome(AccumuloClient c, String table, long count) throws Exception {
     int iterations = 5;
```
keith-turner: @Manno15 could `iterations`, `attempts`, and the for loop be removed?  Seems like this loop only existed to write multiple times and get the median time.

Manno15: Yeah, that appears to be the case. I'll remove it and then test it. Thanks

Issue #1121, opened 2019-04-23
Use conditional mutations to update metadata table


keith-turner:
https://issues.apache.org/jira/browse/ACCUMULO-2268

keith-turner:
Abstracting writes to the metadata table #816 is related to this.

keith-turner:
Work on this has been completed as part of the elasticity project https://github.com/orgs/apache/projects/164

Issue #936, opened 2019-02-04
Consider storing root tablets list of files in zookeeper


keith-turner:
User tablets store their list of files as absolute URIs in Accumulo's metadata table.  Metadata tablets store their list of files in the root tablet.  The root tablet does not store its list of files anywhere and relies on HDFS.  This creates specialized code for the root tablet.  This also make some changes like #642 more difficult and makes [running in S3 more complicated](https://gist.github.com/keith-turner/f6dcbd33342732e42695d66509239983).

Storing the root tablets list of files in Zookeeper may clean up a lot of this.  Would make sense to consider Accumulo metadata abstractions #816 and #798 if this is done.

Issue #817, opened 2018-12-10
Refactor tablet loading code to use TabletMetadata


keith-turner:
The code that loads tablets could be simplified if it used the new tablet metadata abstraction layer.  

The goal would be to replace the `SortedMap<Key,Value>` at [TabletServer line 2464](https://github.com/apache/accumulo/blob/5484fa80022c52784d4000cdd422941fddec1441/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java#L2464) with a [TabletMetadata](https://github.com/apache/accumulo/blob/5484fa80022c52784d4000cdd422941fddec1441/core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletMetadata.java) object.

This would be an involved change because the load process attempts to handle tablets that were partially split.  This code could all be cleaned up to use TabletMetadata.  

It would probably be best to do this change after #798 is done and tablet metadata can be read for the root tablet.  Also since the code to handle partial splits may modify the metadata table, it may be best to do this change after #816.

keith-turner:
This was done by e357d8ab5a29f28794593d31ca2981619266499c and #1302

PR #1174, opened 2019-05-29
fix #816 abstract metadata and change root metadata schema

### LINKED ISSUE #816: Create a metadata mutation builder

keith-turner:
There are three major changes in this commit :

 * An abstraction layer for interacting with Accumulo's persisted
   metadata called Ample was introduced.  The goal is to eventually make
   all metadata read and write operations use Ample.
 * How the root tablet's metadata is stored in zookeeper was changed. It
   was changed to use a single zookeeper node (which is good for making
   atomic updates to multiple fields). In this single zookeeper node a
   json value is stored.  This json value has the same schema as all
   other metadata tablets, it uses the same column families and
   qualifiers.  This makes updating the json using Accumulo mutations
   easy and reading the json as Accumulo Key values easy.
 * Alot of the root tablet code that used to interact directly with
   Zookeeper was updated to use Ample.  In follow on changes, a lot of
   specialized root tablet code can be completely removed.  Those
   changes were not made in this commit inorder to keep it from
   becoming too large, making it hard to review.

This change is starting point to support many other changes like #816, #817, #936, #1121


keith-turner:
I will probably merge this at the end of the day because I want to make changes that build on this.

core/src/main/java/org/apache/accumulo/core/metadata/schema/RootTabletMetadata.java
```diff
@@ -0,0 +1,178 @@
...
+      String qual = bs2Str(entry.getKey().getColumnQualifierData().toArray());
+      String val = bs2Str(entry.getValue().get());
+
+      gd.columnValues.computeIfAbsent(fam, k -> new TreeMap<>()).put(qual, val);
+    }
+
+    return GSON.toJson(gd);
+  }
+
+  public static RootTabletMetadata fromJson(String json) {
```
milleruntime: Short javadoc would be nice.

core/src/main/java/org/apache/accumulo/core/summary/Gatherer.java
```diff
@@ -527,8 +530,8 @@ public SummaryCollection get(long timeout, TimeUnit unit)
 
   private int countFiles() {
     // TODO use a batch scanner + iterator to parallelize counting files
```
milleruntime: Open a ticket?

PR #1183, opened 2019-06-06
Update TESTING.md

milleruntime:
Made less wordy and added more useful commands.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -19,13 +19,10 @@ limitations under the License.
 
 This document is meant to serve as a quick reference to the automated test suites of Accumulo.
 
-The automated testing suite can be categorized as two sets of tests: unit tests and integration tests. These are the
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -17,96 +17,88 @@ limitations under the License.
...
+```
 
-The integration tests are medium length tests (order minutes for each test class and order hours for the complete suite)
-but are checking for regressions that were previously seen in the codebase. These tests do require a noticable amount of
-resources, at least another gigabyte of memory over what Maven itself requires. As such, it's recommended to have at
-least 3-4GB of free memory and 10GB of free disk space.
+# Integration Tests
 
-## Test Categories
+The integration tests are medium length tests but are checking for regressions that were previously seen in the codebase.
```
jmark99: Would it read better to replace 'but' with 'that'?

jmark99: Otherwise, looks good!

PR #1186, opened 2019-06-06
Add logging of top tablets to enable better hot spot detection

alerman:
In order to aid in detection of read and write hot spots, the goal of this PR is to log the already calculated stats, only when a config property is set.

As our production systems are on the 1.9 line, and this PR will not have any effect unless the user explicitly updates the property, I would like to see this get into the 1.9 baseline.


alerman:
I'm going to close this and reopen tomorrow after adding a configurable timeout for this that defaults to a longer time.

PR #1187, opened 2019-06-07
Log busy tablets by ingest and query at a configurable time duration

alerman:
In order to aid in detection of read and write hot spots, the goal of this PR is to log the already calculated stats, only when a config property is set.

As our production systems are on the 1.9 line, and this PR will not have any effect unless the user explicitly updates the property, I would like to see this get into the 1.9 baseline.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
@@ -359,51 +359,58 @@ public class TabletServer extends AccumuloServerContext implements Runnable {
     this.logSorter = new LogSorter(instance, fs, aconf);
     this.replWorker = new ReplicationWorker(this, fs);
     this.statsKeeper = new TabletStatsKeeper();
-    final int numTopTabletsToLog = aconf.getCount(Property.TSERV_LOG_TOP_TABLETS_COUNT);
-    final long logTopTabletsDelay = aconf.getTimeInMillis(Property.TSERV_LOG_TOP_TABLETS_INTERVAL);
-    if (numTopTabletsToLog > 0) {
+    final int numBusiestTabletsToLog = aconf.getCount(Property.TSERV_LOG_BUSIEST_TABLETS_COUNT);
+    final long logBusiestTabletsDelay =
+        aconf.getTimeInMillis(Property.TSERV_LOG_BUSIEST_TABLETS_INTERVAL);
+
+    //This thread will calculate and log out the busiest tablets based on ingest count and
+    // query count every #{logBusiestTabletsDelay}
+    if (numBusiestTabletsToLog > 0) {
       SimpleTimer.getInstance(aconf).schedule(new Runnable() {
         @Override
         public void run() {
-          Comparator<Pair<String,Long>> topTabletComparator = new Comparator<Pair<String,Long>>() {
-            @Override
-            public int compare(Pair<String,Long> first, Pair<String,Long> second) {
-              return second.getSecond().compareTo(first.getSecond());
-            }
-          };
-          PriorityQueue<Pair<String,Long>> topTabletsByIngestCount =
-              new PriorityQueue<>(numTopTabletsToLog, topTabletComparator);
-          PriorityQueue<Pair<String,Long>> topTabletsByQueryCount =
-              new PriorityQueue<>(numTopTabletsToLog, topTabletComparator);
+          Comparator<Pair<String,Long>> busiestTabletComparator =
+              new Comparator<Pair<String,Long>>() {
+                @Override
+                public int compare(Pair<String,Long> first, Pair<String,Long> second) {
+                  return second.getSecond().compareTo(first.getSecond());
+                }
+              };
+          PriorityQueue<Pair<String,Long>> busiestTabletsByIngestCount =
+              new PriorityQueue<>(numBusiestTabletsToLog, busiestTabletComparator);
+          PriorityQueue<Pair<String,Long>> busiestTabletsByQueryCount =
+              new PriorityQueue<>(numBusiestTabletsToLog, busiestTabletComparator);
           synchronized (onlineTablets) {
             for (Tablet tablet : onlineTablets.values()) {
-              addToTopTablets(tablet.totalIngest(), topTabletsByIngestCount, numTopTabletsToLog);
-              addToTopTablets(tablet.totalQueries(), topTabletsByQueryCount, numTopTabletsToLog);
+              addToBusiestTablets(tablet.totalIngest(), busiestTabletsByIngestCount,
+                  numBusiestTabletsToLog);
+              addToBusiestTablets(tablet.totalQueries(), busiestTabletsByQueryCount,
+                  numBusiestTabletsToLog);
             }
-            logTopTablets(topTabletsByIngestCount, "QUERY", numTopTabletsToLog);
-            logTopTablets(topTabletsByQueryCount, "INGEST", numTopTabletsToLog);
+            logBusiestTablets(busiestTabletsByIngestCount, "QUERY", numBusiestTabletsToLog);
+            logBusiestTablets(busiestTabletsByQueryCount, "INGEST", numBusiestTabletsToLog);
           }
         }
 
-        private void addToTopTablets(long count,
-            PriorityQueue<Pair<String,Long>> topTabletsQueue, int numTopTabletsToLog) {
-          if (topTabletsQueue.size() < numTopTabletsToLog
-              || topTabletsQueue.peek().getSecond() < count) {
-            if (topTabletsQueue.size() == numTopTabletsToLog) {
-              topTabletsQueue.remove();
+        private void addToBusiestTablets(long count,
+            PriorityQueue<Pair<String,Long>> busiestTabletsQueue, int numBusiestTabletsToLog) {
+          if (busiestTabletsQueue.size() < numBusiestTabletsToLog
+              || busiestTabletsQueue.peek().getSecond() < count) {
+            if (busiestTabletsQueue.size() == numBusiestTabletsToLog) {
+              busiestTabletsQueue.remove();
             }
           }
         }
 
-        private void logTopTablets(PriorityQueue<Pair<String,Long>> topTabletsQueue,
-            String label, int numTopTabletsToLog) {
-          for (int i = 0; i < numTopTabletsToLog; i++) {
-            Pair<String,Long> pair = topTabletsQueue.poll();
-            log.debug("Top {} tablet by {} count -- extent: {} count: {}", i, label,
+        private void logBusiestTablets(PriorityQueue<Pair<String,Long>> busiestTabletsQueue,
+            String label, int numBusiestTabletsToLog) {
+          for (int i = 0; i < numBusiestTabletsToLog; i++) {
+            Pair<String,Long> pair = busiestTabletsQueue.poll();
+            log.debug("{} busiest tablet by {} count -- extent: {} count: {}", i, label,
                 pair.getFirst(), pair.getSecond());
           }
         }
-      }, logTopTabletsDelay, logTopTabletsDelay);
+      }, logBusiestTabletsDelay, logBusiestTabletsDelay);
     }
 
     SimpleTimer.getInstance(aconf).schedule(new Runnable() {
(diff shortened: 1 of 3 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +359,53 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+            public int compare(Pair<String,Long> first, Pair<String,Long> second) {
+              return second.getSecond().compareTo(first.getSecond());
+            }
+          };
+          PriorityQueue<Pair<String,Long>> topTabletsByIngestCount =
+              new PriorityQueue<>(numTopTabletsToLog, topTabletComparator);
+          PriorityQueue<Pair<String,Long>> topTabletsByQueryCount =
+              new PriorityQueue<>(numTopTabletsToLog, topTabletComparator);
+          synchronized (onlineTablets) {
+            for (Tablet tablet : onlineTablets.values()) {
```
milleruntime: Looping on every online Tablet could hinder performance on a larger cluster.  There could be a lot of tablets.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +359,53 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+              || topTabletsQueue.peek().getSecond() < count) {
+            if (topTabletsQueue.size() == numTopTabletsToLog) {
+              topTabletsQueue.remove();
+            }
+          }
+        }
+
+        private void logTopTablets(PriorityQueue<Pair<String,Long>> topTabletsQueue,
+            String label, int numTopTabletsToLog) {
+          for (int i = 0; i < numTopTabletsToLog; i++) {
```
milleruntime: I am not sure what you are logging... what do you mean by "top"?

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +359,53 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
     this.logSorter = new LogSorter(instance, fs, aconf);
     this.replWorker = new ReplicationWorker(this, fs);
     this.statsKeeper = new TabletStatsKeeper();
+    final int numTopTabletsToLog = aconf.getCount(Property.TSERV_LOG_TOP_TABLETS_COUNT);
+    final long logTopTabletsDelay = aconf.getTimeInMillis(Property.TSERV_LOG_TOP_TABLETS_INTERVAL);
+    if (numTopTabletsToLog > 0) {
+      SimpleTimer.getInstance(aconf).schedule(new Runnable() {
```
milleruntime: Would be nice to have some comments explaining this thread.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
alerman: There is a version of this deployed on all of our production clusters today. (This version actually has some improvements). The code to collect stats loops over the tablets every 5 seconds, and the default for this is once an hour

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
alerman: ill refactor to make it more clear

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
alerman: will add

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
alerman: Additionally, as long as the general approach is acceptable I will patch one of our test clusters with this prior to merge to ensure we dont see negative impacts

milleruntime:
Cool, thanks.  "Busy" is more clear to me than "top".  You could just use "busy" and make the name something like TSERV_LOG_BUSY_TABLETS_COUNT

alerman:
Yea, I was thinking that. I'll update it to use busy

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +359,60 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+            }
+            logBusyTablets(busiestTabletsByIngestCount, "QUERY", numBusyTabletsToLog);
+            logBusyTablets(busiestTabletsByQueryCount, "INGEST", numBusyTabletsToLog);
+          }
+        }
+
+        private void addToBusiestTablets(long count,
+            PriorityQueue<Pair<String,Long>> busiestTabletsQueue, int numBusiestTabletsToLog) {
+          if (busiestTabletsQueue.size() < numBusiestTabletsToLog
+              || busiestTabletsQueue.peek().getSecond() < count) {
```
keith-turner: I think this priority queue has the minimum count at the top, it that correct?

When is something ever added to busiestTabletsQueue?

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +359,60 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+          PriorityQueue<Pair<String,Long>> busiestTabletsByQueryCount =
+              new PriorityQueue<>(numBusyTabletsToLog, busiestTabletComparator);
+          synchronized (onlineTablets) {
+            for (Tablet tablet : onlineTablets.values()) {
+              addToBusiestTablets(tablet.totalIngest(), busiestTabletsByIngestCount,
+                  numBusyTabletsToLog);
+              addToBusiestTablets(tablet.totalQueries(), busiestTabletsByQueryCount,
+                  numBusyTabletsToLog);
+            }
+            logBusyTablets(busiestTabletsByIngestCount, "QUERY", numBusyTabletsToLog);
```
keith-turner: Every read and write to Accumuo needs to sync on `onlineTablets`.  This can cause contention between unrelated threads.  I fixed this for 2.0 with #1100.  However in 1.9 it would be best to minimize the time spent with the lock held on onlineTablets.  There are two possible ways to do this here, could just move these logging statements out since logging could do IO.  Could take it a step further and do something like the following, which avoids the O(n log m) operation of adding the prio queue while the lock is held.

```java
List< Tablet> tablets;
synchronized (onlineTablets) {
  tablets = new ArrayList<>(onlineTablets.values);
}

// do all computation that was done in sync block here using tablets list

```

milleruntime: Good catch.  I had a feeling touching `onlineTablets` would be troublesome.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
alerman: doh!

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +359,60 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+              };
+          PriorityQueue<Pair<String,Long>> busiestTabletsByIngestCount =
+              new PriorityQueue<>(numBusyTabletsToLog, busiestTabletComparator);
+          PriorityQueue<Pair<String,Long>> busiestTabletsByQueryCount =
+              new PriorityQueue<>(numBusyTabletsToLog, busiestTabletComparator);
+          synchronized (onlineTablets) {
+            for (Tablet tablet : onlineTablets.values()) {
+              addToBusiestTablets(tablet.totalIngest(), busiestTabletsByIngestCount,
+                  numBusyTabletsToLog);
+              addToBusiestTablets(tablet.totalQueries(), busiestTabletsByQueryCount,
```
keith-turner: The counts the priority queue is tracking just increase forever.  So if one tablet has been on a tablet server a year and its count is high, it may show up as busy even though its not currently busy.  Need to track the delta in the counts between sampling.

keith-turner: I see there are ingest and query rate objects that use exponential decay.  These are declared in proximity to the counts.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +364,77 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+              busiestTabletsQueue.remove();
+            }
+            busiestTabletsQueue.add(new Pair<String,Double>(extent, count));
+          }
+        }
+
+        private void logBusyTablets(Map<String,PriorityQueue<Pair<String,Double>>> busyTabletsMap,
+            String label, int numBusiestTabletsToLog) {
+          PriorityQueue<Pair<String,Double>> busyTabletsQueue = busyTabletsMap.get(label);
+          for (int i = 0; i < numBusiestTabletsToLog; i++) {
```
keith-turner: I think the expectation is that this code empties the queue.  If it did not empty the queue for some reason it could lead to odd situations like the same tablet being on the queue multiple times.  To make this expectation more clear and to decouple this code from how the queue is created, could do the following instead of the for loop.

```java
int i = 0;
while(!busyTabletsQueue.isEmpty()) {
  Pair<String,Double> pair = busyTabletsQueue.poll();
  log.debug("{} busiest tablet by {} -- extent: {} count: {}", i++, label, pair.getFirst(),
￼                 pair.getSecond());
}
```

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +364,77 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+            Pair<String,Double> pair = busyTabletsQueue.poll();
+            log.debug("{} busiest tablet by {} -- extent: {} count: {}", i, label, pair.getFirst(),
```
keith-turner: Since the key extent could be really long, it may sense to put it last in the log message.  Maybe swap count and extent.  Also, would it make sense to replace count with the label?  Something like :

```java
log.debug("{} busiest tablet by {}: {} -- extent: {} ", i, label.toLowerCase(), pair.getSecond(), pair.getFirst());
```

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -277,6 +278,11 @@
   private static final long TIME_BETWEEN_GC_CHECKS = 5000;
   private static final long TIME_BETWEEN_LOCATOR_CACHE_CLEARS = 60 * 60 * 1000;
 
+  private static final String INGEST_COUNT = "INGEST COUNT";
```
ivakegg: We are wondering if the counts will actually mean anything without knowing when the tablet was initially created on this server.  If the goal is to know the busiest tablets, then the rates are probably sufficient.  However if you want to know the total rate then we will need to log the count with along with the time that the count started.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -358,6 +364,79 @@ public TabletServer(ServerConfigurationFactory confFactory, VolumeManager fs) th
...
+
+          logBusyTablets(busyTabletMap, QUERY_COUNT);
+          logBusyTablets(busyTabletMap, QUERY_RATE);
+          logBusyTablets(busyTabletMap, INGEST_COUNT);
+          logBusyTablets(busyTabletMap, INGEST_RATE);
+        }
+
+        private void addToBusiestTablets(String extent, double count,
+            PriorityQueue<Pair<String,Double>> busiestTabletsQueue, int numBusiestTabletsToLog) {
+          if (busiestTabletsQueue.size() < numBusiestTabletsToLog
```
keith-turner: @alerman while I was writing unit test for #1291 I found that this algorithm does not work computing the top k.  After the queue is full it only adds items that are greater than the max in the queue.  However, something could be less than the max in the queue and still in the top K.  I just wanted to let you know about this in case you were using this patch.

Before writing unit test, I thought the algorithm was fine.

keith-turner:
These changes were merged in 83429b63e83682618a4679104331b48eed6d2172

PR #1204, opened 2019-06-13
Allow master start to block waiting for a number of tservers

EdColeman:
Adds two parameters:

 - MASTER_STARTUP_TSERVER_AVAIL_MIN_COUNT - sets desried number of tservers
 - MASTER_STARTUP_TSERVER_AVAIL_MAX_WAIT - sets maximum time to wait.

This is the same changes submitted as pull request #1158 for 1.9.x, merged to
2.0.  Request #1158 will be available as a patch, but is not expected to be merged
becuase of semver requirements.

### DIFF
--- a/server/master/src/main/java/org/apache/accumulo/master/Master.java
+++ b/server/master/src/main/java/org/apache/accumulo/master/Master.java
@@ -77,6 +77,7 @@ import org.apache.accumulo.core.trace.TraceUtil;
 import org.apache.accumulo.core.util.Daemon;
 import org.apache.accumulo.fate.AgeOffStore;
 import org.apache.accumulo.fate.Fate;
+import org.apache.accumulo.fate.util.Retry;
 import org.apache.accumulo.fate.zookeeper.ZooLock;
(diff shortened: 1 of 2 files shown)


server/master/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -1207,6 +1214,91 @@ boolean canSuspendTablets() {
...
+
+    if (minTserverCount <= 0) {
+      log.info(
+          "tserver availability check disabled, contining with-{} servers." + "To enable, set {}",
+          tserverSet.size(), Property.MASTER_STARTUP_TSERVER_AVAIL_MIN_COUNT.getKey());
+      return;
+    }
+
+    long maxWait =
+        getConfiguration().getTimeInMillis(Property.MASTER_STARTUP_TSERVER_AVAIL_MAX_WAIT);
```
milleruntime: This is the new way to get the Accumulo config in 2.0+ so this looks good.

PR #1243, opened 2019-07-02
Fix #1242 - Update MasterMetadataUtil.getUpdateForTabletDataFile

### LINKED ISSUE #1242: Update MasterMetadataUtil.getUpdateForTabletDataFile to use ample api
The getUpdateForTableDataFile is the last piece in MasterMetadataUtil NOT using the new ample api.

hkeebler:
Updated to use the ample api.  Also refactored to use same variable name "tablet" as mutator for consistency.

Observation:  why is it possible for some things like 'unusedwalogs' to be null but not others?

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
@@ -34,7 +34,6 @@ import org.apache.accumulo.core.client.AccumuloException;
 import org.apache.accumulo.core.client.Scanner;
 import org.apache.accumulo.core.clientImpl.ScannerImpl;
 import org.apache.accumulo.core.data.Key;
-import org.apache.accumulo.core.data.Mutation;
 import org.apache.accumulo.core.data.PartialKey;
 import org.apache.accumulo.core.data.Range;
 import org.apache.accumulo.core.data.TableId;
@@ -42,13 +41,10 @@ import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.dataImpl.KeyExtent;
 import org.apache.accumulo.core.metadata.MetadataTable;
 import org.apache.accumulo.core.metadata.RootTable;
-import org.apache.accumulo.core.metadata.schema.Ample;
 import org.apache.accumulo.core.metadata.schema.Ample.TabletMutator;
 import org.apache.accumulo.core.metadata.schema.DataFileValue;
 import org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSection;
 import org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSection.DataFileColumnFamily;
-import org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSection.LogColumnFamily;
-import org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSection.ScanFileColumnFamily;
 import org.apache.accumulo.core.metadata.schema.TabletMetadata.LocationType;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.accumulo.core.util.ColumnFQ;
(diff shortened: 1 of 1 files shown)


server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
```diff
@@ -247,61 +243,58 @@ public static void updateTabletDataFile(ServerContext context, KeyExtent extent,
       String address, ZooLock zooLock, Set<String> unusedWalLogs, TServerInstance lastLocation,
       long flushId) {
     if (extent.isRootTablet()) {
-      if (unusedWalLogs != null) {
-        updateRootTabletDataFile(context, unusedWalLogs);
-      }
-      return;
+      updateRootTabletDataFile(context, unusedWalLogs);
+    } else {
```
keith-turner: I like this restructuring of the code to use if/else dropping the return.

server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
```diff
@@ -247,61 +243,58 @@ public static void updateTabletDataFile(ServerContext context, KeyExtent extent,
...
-      TabletsSection.ServerColumnFamily.TIME_COLUMN.put(m, new Value(time.getBytes(UTF_8)));
-      // stuff in this location
+      tablet.putFile(path, dfv);
+      tablet.putTime(time);
+
       TServerInstance self = getTServerInstance(address, zooLock);
-      self.putLastLocation(m);
-      // erase the old location
-      if (lastLocation != null && !lastLocation.equals(self))
-        lastLocation.clearLastLocation(m);
```
keith-turner: Eventually this clearLastLocation method will become dead code and can be removed.

server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
```diff
@@ -247,61 +243,58 @@ public static void updateTabletDataFile(ServerContext context, KeyExtent extent,
...
+    if (mergeFile != null) {
+      tablet.deleteFile(mergeFile);
+    }
+    if (unusedWalLogs != null) {
+      unusedWalLogs.forEach(tablet::deleteWal);
+    }
+    filesInUseByScans.forEach(tablet::putScan);
 
-    return m;
+    tablet.mutate();
```
keith-turner: This method should call `tablet.putZooLock(zooLock);`  to preserve existing behavior.

keith-turner:
> Observation: why is it possible for some things like 'unusedwalogs' to be null but not others?

I don't know.  I think it would be cleaner if null were never expected and callers always passed something.  Callers could pass an empty collection.  Would need to analyze the code that calls this to make that change.

hkeebler:
I agree and sort of started that analysis.  We can do that as part of this ticket.

server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
hkeebler: should it be deprecated?

server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
hkeebler: yes.  I missed that it was being done in the MetadataTableUtil.update()

server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
keith-turner: No, its not public API.  So when nothing is using it, it can be deleted.

hkeebler: So how are those types of things cleaned up.  Is there something run periodically that identifies unused code?

keith-turner: I think there are tools that will find unused public methods, but I am not sure.  I think in the past someone has run a tool like that and removed the code that made sense.  But I am not sure what the tool was.

milleruntime: IntelliJ has a great built in Analysis tool for unused code.  That is what I use.

hkeebler:
Confirmed that collections passed into the updateTabletDataFile are not null.  They have data or are empty.    
For unusedWalLogs followed a single thread ending with `Set<String> unusedWalLogs = tablet.beginClearingUnusedLogs();`
Other collections cycle back to MetadataTableUtil methods which initialize the collections.  So I think we are good here.

server/base/src/main/java/org/apache/accumulo/server/util/MasterMetadataUtil.java
hkeebler: So funny... I lterally just a few moments ago saw one faded out and remembered .."oh yeah, intellj does this automatically"... so next question is if something is noticed here not being used should we put a ticket in to remove it?  Like ```
MetadataTableUtil.getBulkFilesLoaded(ServerContext context, AccumuloClient client,
      KeyExtent extent, long tid)
```
This signature with 'tid' is not used.

milleruntime: I usually do a separate refactor PR like:  Remove unused code.  It is easier to review if its all just unused code being removed.

keith-turner:
@hkeebler if it would not cause you problems, could you please hold of on squishing everything into one commit in the future?  I am doing a lot of context switching and its really nice to see all of the commits in the code review and only look at what changed since the last time I looked.  I was circling back to this to see if its ready to merge and I have no idea what changes were made since the last time I looked at it.

hkeebler:
@keith-turner yes I can.  I usally do a commit -a --amend to reduce the messages but will not do that in the future on re-pushes.

PR #1248, opened 2019-07-03
fix #1247 store fate txid in ~blip marker value

### LINKED ISSUE #1247: Annotate blip markers when exception happens during bulk import
When an unexpected exception happens during bulk import it may be that not all files are imported.  This happened in #1231.  Bulk import are executed by FATE.  When FATE see's an unexpected exception it marks the transaction as failed and starts executing the Undo operations.  For bulk import, one of the undo steps could set a value of `FAILED` on the blip marker.  If this were done then a user could grep the `~blip` range for `FAILED` to find all bulk import directories that may have been partially imported.

Possible causes of unexpected exceptions are bugs and improperly handled zookeeper and hdfs exceptions.

keith-turner:
Before this change the bulk import code was serializing and logging fate
transaction ids differently than other parts of the code.  When adding
FATE transaction ids to blip markers, I decided it would be best if they
were consistently persisted (in metadata table ) and logged. That
decision made this change much larger than simply adding something to
the blip value, but I feel its worth it.

The following are the main changes :

 * Introduce new class FateTxId with methods for consistenly formatting
   fate transaction ids.
 * Make bulk fate logging use FateTxId
 * Update ~blip and loaded family to use FateTxId.  Made loaded family
   backwards compat.
 * Update existing fate logging code to use FateTxId.  Instead of
   String.format

Follow on work to update all Repos to use FateTxId would be nice.  Chose
not do to that in the change because it would make it harder to review.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/fate/AgeOffStore.java
+++ b/core/src/main/java/org/apache/accumulo/fate/AgeOffStore.java
@@ -95,7 +95,7 @@ public class AgeOffStore<T> implements TStore<T> {
             case FAILED:
             case SUCCESSFUL:
               store.delete(txid);
-              log.debug("Aged off FATE tx {}", String.format("%016x", txid));
+              log.debug("Aged off FATE tx {}", FateTxId.formatTid(txid));
               break;
             default:
               break;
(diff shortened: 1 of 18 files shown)


keith-turner:
This is what this change looks like

```
root@muchos> scan -t accumulo.metadata -b ~b -e ~c
~blip/2/b-00009a8 : []    FATE[77ca280813b02273]
```

PR #1253, opened 2019-07-08
Created SequenceLexicoder to resolve #1252

### LINKED ISSUE #1252: ListLexicoder does not support empty lists
Attempting to use a `ListLexicoder` to encode an empty list yields an `ArrayIndexOutOfBoundsException` due to the following line in `ByteUtils.concat`:

```java
byte[] ret = new byte[len + fields.length - 1];
```

Unfortunately, any attempt to resolve this is not likely to be backwards compatible with existing data encoded using the current implementation.

I suggest adding a separate class (perhaps called something like `SequenceLexicoder` to distinguish it from the existing `ListLexicoder` name) the is able to handle possibly-empty lists and deprecating the existing class (or at least clearly documenting that it is not appropriate to use in cases where the list may be empty). Probably the best solution in terms of implementation is to append an empty element to the end of the list that gets passed to `ByteUtils.concat`. In this case, an empty list will be encoded as an empty value, and a list containing exactly one empty element (e.g., a list containing only an empty string) will be encoded as a single delimiter byte.

reggert:
This provides a replacement for `ListLexicoder` that supports empty lists, resolving #1252 .

The branch is based on the `master` branch, but I can rebase it to a different branch if necessary to include it in the 2.0 release.

I'm also open to suggestions for a different name for the class if anyone can think of a better one.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/lexicoder/ListLexicoder.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/lexicoder/ListLexicoder.java
@@ -31,6 +31,7 @@ import org.apache.accumulo.core.clientImpl.lexicoder.AbstractLexicoder;
  * encoded element sorts lexicographically.
  *
  * Note: Empty lists are not supported.
+ * @see SequenceLexicoder for an encoding that supports empty lists.
  *
  * @since 1.6.0
  */
@@ -49,6 +50,9 @@ public class ListLexicoder<LT> extends AbstractLexicoder<List<LT>> {
    */
   @Override
   public byte[] encode(List<LT> v) {
+    if (v.isEmpty()) {
+      throw new IllegalArgumentException("ListLexicoder does not support empty lists");
+    }
     byte[][] encElements = new byte[v.size()][];
 
     int index = 0;
--- a/core/src/main/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoder.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoder.java
@@ -16,6 +16,7 @@
  */
 package org.apache.accumulo.core.client.lexicoder;
 
+import static java.util.Objects.requireNonNull;
 import static org.apache.accumulo.core.clientImpl.lexicoder.ByteUtils.concat;
 import static org.apache.accumulo.core.clientImpl.lexicoder.ByteUtils.escape;
 import static org.apache.accumulo.core.clientImpl.lexicoder.ByteUtils.split;
@@ -26,6 +27,8 @@ import java.util.List;
 
 import org.apache.accumulo.core.clientImpl.lexicoder.AbstractLexicoder;
 
+import javax.annotation.Nonnull;
+
 /**
  * A lexicoder to encode/decode a Java List to/from a byte array where the concatenation of each
  * encoded element sorts lexicographically.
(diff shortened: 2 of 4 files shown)


core/src/test/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoderTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+  private final List<Long> data0 = new ArrayList<>();
+  private final List<Long> data1 = new ArrayList<>();
+  private final List<Long> data2 = new ArrayList<>();
+  private final List<Long> data3 = new ArrayList<>();
+  private final List<Long> data4 = new ArrayList<>();
+  private final List<Long> data5 = new ArrayList<>();
+
+  @Before
+  public void setUp() {
```
keith-turner: Would be nice if the data had empty elements in order to test the sort order of list with empty elements vs empty list.

core/src/main/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoder.java
```diff
@@ -0,0 +1,78 @@
...
+
+    byte[][] escapedElements = split(b, offset, len);
+    assert escapedElements.length > 0;
+    ArrayList<LT> ret = new ArrayList<>(escapedElements.length);
+
+    for (int i = 0; i < escapedElements.length - 1; i++) {
+      byte[] escapedElement = escapedElements[i];
+      ret.add(lexicoder.decode(unescape(escapedElement)));
+    }
+
```
keith-turner: Could throw an exception if the last ignored element in `escapedElements` is not empty byte array.

core/src/main/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoder.java
```diff
@@ -0,0 +1,78 @@
...
+    encElements[v.size()] = new byte[0];
+
+    return concat(encElements);
+  }
+
+  @Override
+  protected List<LT> decodeUnchecked(byte[] b, int offset, int len) {
+
+    byte[][] escapedElements = split(b, offset, len);
+    assert escapedElements.length > 0;
```
keith-turner: Could throw an exception here.

core/src/main/java/org/apache/accumulo/core/client/lexicoder/ListLexicoder.java
```diff
@@ -30,6 +30,8 @@
  * A lexicoder to encode/decode a Java List to/from a byte array where the concatenation of each
  * encoded element sorts lexicographically.
  *
+ * Note: Empty lists are not supported.
+ *
```
keith-turner: This could use the `@see` javadoc tag to point to SequenceLexicoder

Also, could make the encode method throw an exception when encoding an empty list.  The exception could have a message saying not support and mention the new encoded.

core/src/test/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoderTest.java
reggert: I just copied `ListLexicoder` test and added an empty list to the data sets. I can add a case for empty elements, but it will require changing the element type from `Long` to something with variable since like `String`.

core/src/main/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoder.java
reggert: True. It might make a useful sanity check to catch cases where the data is not formatted as expected, though we otherwise wouldn't really care about the last element.

core/src/main/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoder.java
reggert: I thought about this, but then realized that `split` will always return at least one element, even if the input is empty. I added the assertion to validate this invariant. I'll add a comment to it to explain the reasoning.

core/src/main/java/org/apache/accumulo/core/client/lexicoder/ListLexicoder.java
reggert: Agreed.

reggert:
I've addressed the issues and cleaned up the code a bit. I've also re-based onto the 2.0 branch to avoid being dependent upon Java 11. I'll open a new PR with the updated merge target.

core/src/test/java/org/apache/accumulo/core/client/lexicoder/SequenceLexicoderTest.java
keith-turner: Thanks for doing this in the new PR

PR #1256, opened 2019-07-08
Fix oldest age on bulk import page. Closes #1251

### LINKED ISSUE #1251: Fix Monitor Bulk Import "Per TabletServer" table
I am not sure what this table on the Monitor is supposed to track but I don't think its working.  We could probably just remove the table from the page.  It is under Activity -> Bulk Imports

milleruntime:


EdColeman:
Did you look to see if this can be back-ported to 1.9.x - or has the code sufficiently changed that it is not an issue?

milleruntime:
> Did you look to see if this can be back-ported to 1.9.x - or has the code sufficiently changed that it is not an issue?

This only exists in the new 2.0 Monitor.

PR #1275, opened 2019-07-15
Fix #1210 Create type for time in metadata

### LINKED ISSUE #1210: Create type for time to use when reading and write Accumulo metadata
Currently when using Ample to read and write time to the Accumulo medata a String is used.  It would be better to use a type that encodes the time type and long.

hkeebler:
Created a MetadataTime object  and utilized the TimeType instead of a single
char M and L; Also removed the MetadataTableUtil.getDataFileSizes which
was only used by tests.  It was replaced with using
getFileAndLogEntries(..).getSecond();

Quite a few files were effected by this change but most of them are small.  More work can be done to clean the time data passed around.  I would like to remove the getMetadataValue methods from the TabletTime  (replaced with the getMetadataTime methods) but that can be a second pass.  Additionally I will write a small unit test for the MetadataTime class.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/metadata/schema/MetadataTime.java
+++ b/core/src/main/java/org/apache/accumulo/core/metadata/schema/MetadataTime.java
@@ -1,0 +1,114 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.accumulo.core.metadata.schema;
+
+import java.util.Objects;
+
+import org.apache.accumulo.core.client.admin.TimeType;
+
+/**
+ * Immutable metadata time object
+ */
+public class MetadataTime {
+  private final long time;
+  private final TimeType type; // default type is M for milliseconds
+
+  public MetadataTime(long time, TimeType type) {
+    this.time = time;
+    this.type = type;
+  }
+
+  /**
+   * Creates a MetadataTime object from a string
+   *
+   * @param timestr
+   *          string representation of a metatdata time, ex. "M12345678"
+   * @return a MetadataTime object represented by string
+   */
+
+  public static MetadataTime parse(String timestr) throws IllegalArgumentException {
+
+    if (timestr != null && timestr.length() > 1) {
+      return new MetadataTime(Long.parseLong(timestr.substring(1)), valueOf(timestr.charAt(0)));
+    } else
+      throw new IllegalArgumentException("Unknown metadata time value " + timestr);
+  }
+
+  /**
+   * Converts timetypes to data codes used in the table data implementation
+   *
+   * @param code
+   *          character M or L otherwise exception thrown
+   * @return a TimeType {@link TimeType} represented by code.
+   */
+  public static TimeType valueOf(char code) {
+    switch (code) {
+      case 'M':
+        return TimeType.MILLIS;
+      case 'L':
+        return TimeType.LOGICAL;
+      default:
+        throw new IllegalArgumentException("Unknown time type code : " + code);
+    }
+  }
+
+  /**
+   * @return the single char code of this objects timeType
+   */
+  public static char getCode(TimeType type) {
+    switch (type) {
+      case MILLIS:
+        return 'M';
+      case LOGICAL:
+        return 'L';
+      default: // this should never happen
+        throw new IllegalArgumentException("Unknown time type: " + type);
+    }
+  }
+
+  public char getCode() {
+    return getCode(this.type);
+  }
+
+  public TimeType getType() {
+    return type;
+  }
+
+  public long getTime() {
+    return time;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o instanceof MetadataTime) {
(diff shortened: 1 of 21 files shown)


server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
```diff
@@ -669,7 +670,7 @@ private void deleteTablets(MergeInfo info) throws AccumuloException {
             + Path.SEPARATOR + extent.getTableId() + Constants.DEFAULT_TABLET_LOCATION;
         MetadataTableUtil.addTablet(
             new KeyExtent(extent.getTableId(), null, extent.getPrevEndRow()), tdir,
-            master.getContext(), timeType, this.master.masterLock);
+            master.getContext(), metadataTime.getType(), this.master.masterLock);
```
keith-turner: There is a difference in the code here in that metadataTime could possibly be null.  However I think its ok, because I am thinking it should not be null because tablets should always have time type.  Also the old code was doing something really strange with `\0`

server/base/src/main/java/org/apache/accumulo/server/tablets/TabletTime.java
```diff
@@ -21,22 +21,20 @@
...
 import org.apache.accumulo.core.client.admin.TimeType;
 import org.apache.accumulo.core.data.Mutation;
+import org.apache.accumulo.core.metadata.schema.MetadataTime;
 import org.apache.accumulo.server.data.ServerMutation;
 import org.apache.accumulo.server.util.time.RelativeTime;
 
 public abstract class TabletTime {
-  public static final char LOGICAL_TIME_ID = 'L';
-  public static final char MILLIS_TIME_ID = 'M';
+  public static final char LOGICAL_TIME_ID = MetadataTime.getCode(TimeType.LOGICAL);
```
keith-turner: Not completely sure, but I was thinking inlining this may be good.

hkeebler:
Thanks for the quick review.  I'll address all your responses.

core/src/main/java/org/apache/accumulo/core/metadata/schema/MetadataTime.java
```diff
@@ -0,0 +1,114 @@
...
+    return false;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(time, type);
+  }
+
+  @Override
+  public String toString() {
```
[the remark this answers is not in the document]
ctubbsii: Of the two, I prefer the explicit method.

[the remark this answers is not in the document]
hkeebler: Agree.  encode() it is.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
```diff
@@ -2255,7 +2257,7 @@ public boolean isMajorCompactionQueued() {
       log.debug("Files for low split {} {}", low, lowDatafileSizes.keySet());
       log.debug("Files for high split {} {}", high, highDatafileSizes.keySet());
 
-      String time = tabletTime.getMetadataValue();
+      MetadataTime time = tabletTime.getMetadataTime(); /** hik **/
```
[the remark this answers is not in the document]
hkeebler: oops.

server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
hkeebler: Yes.  I was not happy with assigning null but that is what the previous code was doing although it was obviously always assigned something because there are no null char checks down the line.  It could be assigned the default value of MILLIS.

server/base/src/main/java/org/apache/accumulo/server/tablets/TabletTime.java
hkeebler: So wherever LOGICAL_TIME_ID is used instead use MetadataTime.getCode(TimeType.LOGICAL).  I started that way but then thought it might be cleaner to keep the constants. BUT...I also think it may be better to remove this Timetype from TabletTime.  It is really just a time calculator.  There is no need for the code to ask TabletTime for the Timetype.  The more I think about it the more I definitely think it should be removed.

server/base/src/main/java/org/apache/accumulo/server/tablets/TabletTime.java
```diff
@@ -21,22 +21,20 @@
...
-    switch (timeType) {
-      case LOGICAL:
-        return LOGICAL_TIME_ID;
-      case MILLIS:
-        return MILLIS_TIME_ID;
-    }
-
-    throw new IllegalArgumentException("Unknown time type " + timeType);
+    /*
+     * switch (timeType) { case LOGICAL: return LOGICAL_TIME_ID; case MILLIS: return MILLIS_TIME_ID;
```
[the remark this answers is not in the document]
hkeebler: As with one of my comments below.  I think this code surrounding the timeType can be removed from this class because it is all provided from the Metadata class now.

server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
keith-turner: Thinking about it a bit more, I think your change is good because it would fail with an NPE if not set.  I think the expectation is that every tablet should have this column.

hkeebler:
In TabletTime `public static String maxMetadataTime(String mv1, String mv2)` should probably be updated to use MetadataTime objects instead of string but that is for another ticket I think.

keith-turner:
> should probably be updated to use MetadataTime objects instead of string but that is for another ticket 

That sounds good.  Definitely good to avoid putting too much into a single PR.  I like to make sure I open issues for all follow on work identified in a PR before closing it so I don't forget about.

keith-turner:
> a MetadataTimeTest was added and the TabletTimeTest updated to remove duplicate tests.

@hkeebler  I am not seeing the MetadataTimeTest

hkeebler:
Sorry, test should be there now.

hkeebler:
@lbschanno noticed that I need to make the MetadataTime class final to truly make it immutable. So I will make that change but hold back a new push for any additional comments.  Thanks @lbschanno

keith-turner:
@hkeebler I can merge this after you add `final` to metadata time.

hkeebler:
This is ready.  Note I did not squash the commit messages due to your previous request.

PR #1276, opened 2019-07-17
Refactor how constraint violations are handled

### LINKED ISSUE #824: Refactor how constraint violations are handled
[Tablet server code](https://github.com/apache/accumulo/blob/c8d19d45f361f49a4d42b07310393b7bdf68b3b4/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java#L1059) that handles contraint violation uses an exception to return information about constraint violations.  I am not sure about this, but I think it may clean up the code if an exception was not used to return information in this way.  The code could instead always return an object contains info about 0 or more constraint violations.  I think this could cut down on duplicate code, how that may not be the case.

lbschanno:
Refactor out the usage of exceptions as a means of returning information
about constraint violations.

Tablet.prepareMutationsForCommit threw an exception if constraint
violations were found in any mutations and the tablet server was
available for write. Information about the violations were pulled from
the exception for logging and any additional commit preparation. Using
the exception as a means of control flow is an anti-pattern.

Wrap the CommitSession in TabletMutationPrepAttempt that contains all
identified violations and non-violators and explicitly check and handle
any violations in the main body of pertinent functions in TabletServer.

TConstraintViolationException.java is no longer used. Delete it.

Fixes #824


server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
```diff
@@ -1434,7 +1434,25 @@ private void writeConditionalMutations(Map<KeyExtent,List<ServerConditionalMutat
...
+      return attempt.hasViolations() && (!attempt.hasNonViolators() || attempt.hasCommitSession());
+    }
+
+    private void addMutationsAsTCMResults(final List<TCMResult> list,
+        final Collection<? extends Mutation> mutations, final TCMStatus status) {
+      mutations.stream().map(mutation -> toTCMResult(mutation, status)).forEach(list::add);
+    }
 
+    private TCMResult toTCMResult(final Mutation mutation, final TCMStatus status) {
+      return new TCMResult(((ServerConditionalMutation) mutation).getID(), status);
```
ctubbsii: Could inline this, since it's only used once. If doing so would make the lambda less readable in `addMutationsAsTCMResults`, you could use the curly brace form so the lambda used in the map function wraps to the next line better.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/TabletMutationPrepAttempt.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Return true if at least one constraint violation was found, or false otherwise.
+   */
+  public boolean hasViolations() {
+    return !violations.isEmpty();
+  }
+
+  /**
+   * Return the list of mutations that violated a constraint. Either null or a non-empty list.
```
ctubbsii: Nice javadocs. Some of the null-ness of these methods are inconsistent. Some are never null but could be empty, and some are null or non-empty lists. I understand this is residual from some of the previous implementation being refactored into here, but I wonder if there's room for improvement now that this encapsulating class exists.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/TabletMutationPrepAttempt.java
```diff
@@ -0,0 +1,124 @@
...
+    attemptedTabletPrep = true;
+  }
+
+  /**
+   * Add a mutation that violates a constraint, along with information about the constraints that it
+   * violated.
+   */
+  void addViolator(final Mutation mutation, final Violations mutationViolations) {
+    if (violators == null) {
+      violators = new ArrayList<>();
```
ctubbsii: Not thread-safe in a way that could destroy state. Have you made sure that this is only ever used in a single thread? This could be related to my earlier comment about null-ness consistency. You could make sure that this member variable is initialized when this object is constructed, as long as "null" vs. "empty" isn't a meaningful distinction.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/TabletMutationPrepAttempt.java
```diff
@@ -0,0 +1,124 @@
...
+    violators.add(mutation);
+    violations.add(mutationViolations);
+  }
+
+  /**
+   * Add a mutation that did not violate any constraints.
+   */
+  void addNonViolator(final Mutation mutation) {
+    if (nonViolators == null) {
+      nonViolators = new ArrayList<>();
```
ctubbsii: Same comment as above re: thread-safety and null-ness.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/TabletMutationPrepAttempt.java
lbschanno: I'll modify it so that the lists are initialized when the object is constructed. From my understanding of the original code, null vs. not is only a meaningful distinction for the `CommitSession` variable.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/TabletMutationPrepAttempt.java
lbschanno: See comment below.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/TabletMutationPrepAttempt.java
lbschanno: See comment above.

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java
lbschanno: It's not a problem to inline it.

lbschanno:
I added some unit tests for `TabletMutationPrepAttempt`. If you like, I will also tackle adding some unit tests for the modifications to `Tablet.prepareMutationsForCommit` if they are not already covered by another test.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
```diff
@@ -1115,53 +1114,29 @@ private synchronized CommitSession finishPreparingMutations(long time) {
...
-
-      if (nonViolators.size() > 0) {
-        // if everything is a violation, then it is expected that
-        // code calling this will not log or commit
-        commitSession = finishPreparingMutations(time);
-        if (commitSession == null) {
-          return null;
-        }
+    final ConstraintChecker constraints = constraintChecker.derive();
+    final TabletMutationPrepAttempt attempt = new TabletMutationPrepAttempt();
```
keith-turner: The existing code attempts to minimize the work done when no mutations violate a constraint.  It would be nice to carry this behavior forward, but I am not sure of the best way to do this.

Below is one possibility I thought of...

```java
Iterator<Mutation> miter = mutations.iterator(); //TODO reverse iterator would be more efficient when everything is removed
while(miter.hasNext()) {
  Violations violations = constraints.check(cenv, mutation);
  if(violations != null) {
      attempt.addViolator(mutation, violations);  // TODO this could allocate the internal list the first time called
     miter.remove();  // TODO is it ok to modify the list passed in??
  }
}

// if nothing violates, then the passed in list is not copied or modified...
attempt.setMutations(mutations);
```

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
```diff
@@ -1115,53 +1114,29 @@ private synchronized CommitSession finishPreparingMutations(long time) {
...
+        attempt.addViolator(mutation, violations);
+      } else {
+        attempt.addNonViolator(mutation);
       }
+    });
 
-      throw new TConstraintViolationException(violations, violators, nonViolators, commitSession);
+    // If there are no violations, or at least some mutations that do not violate the constraints,
+    // attempt to prepare the tablet and retrieve the commit session.
+    if (!attempt.hasViolations() || attempt.hasNonViolators()) {
```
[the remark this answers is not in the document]
lbschanno: Yes, if we don't need to handle the case where the list of supplied mutations is empty.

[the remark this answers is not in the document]
lbschanno: I checked each call to this function and it looks like there's never an empty list passed in, so I will change it to `if(attempt.hasNonViolators())`.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
lbschanno: I checked each call to this function and there are cases where the original mutations list is still subsequently used, so we should avoid modifying it.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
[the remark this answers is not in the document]
keith-turner: If an empty list would cause problems and it unexpected, then it may make sense to throw an exception if one is passed in.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
keith-turner: > I checked each call to this function and there are cases where the original mutations list is still subsequently used, so we should avoid modifying it.

Ok, thanks for checking that.  I suppose that is why the existing code does something with a hashset when there are mutations.  Could possibly do something like the following.

```java
List<Mutation> vl = null;

for(Mutation mutation : mutations) {
  // TODO check for violations and if when one is found add to vl... creating vl if needed
}

if(vl.isEmpty()){
   // just use mutations list
} else {
   // copy mutations list and subtract violating mutations
}

```

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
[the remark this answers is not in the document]
lbschanno: I don't think an empty list will cause problems since non-violators are only a possibility for a non-empty list.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java
lbschanno: My only caution about this approach is that it makes it possible for the original mutation list to be modified if there are no violations and the list of non-violators is retrieved and modified. I don't think it's currently an issue, and if you're not concerned about it becoming an issue in the future, I'll make the change.

keith-turner: > My only caution about this approach is that it makes it possible for the original mutation list to be modified if there are no violations 

Yeah that is a problem.  One way to address the issue is for the originator of the list to make it immutable.  Accumulo could benefit from using more immutable collections.

lbschanno:
@keith-turner thanks for the changes, they've been merged.

keith-turner:
Merged b9505a6a36a97dc9c029f6621648e83e8f867507

PR #1290, opened 2019-07-24
Rename Bulk ITs to be less confusing

milleruntime:


hkeebler:
👍

________________________________

PR #1304, opened 2019-08-01
Fix #1281 Update TabletTime.maxMetatdataTime

### LINKED ISSUE #1281: Update TabletTime.maxMetadataTime(..) to use MetadataTime objects instead of string
In TabletTime public static String maxMetadataTime(String mv1, String mv2) should  be updated to use MetadataTime objects instead of string.  This was identified in #1210.

hkeebler:
Replaced string processing with MetadataTime objects. 

Renamed MetadataTime class method valueOf(char) to getType(char) .  The valueOf was confusing and didn't really make sense once it was pulled from the ENUM TimeType class.

### DIFF
--- a/server/base/src/test/java/org/apache/accumulo/server/tablets/TabletTimeTest.java
+++ b/server/base/src/test/java/org/apache/accumulo/server/tablets/TabletTimeTest.java
@@ -33,6 +33,10 @@ import org.junit.Test;
 public class TabletTimeTest {
   private static final long TIME = 1234L;
   private MillisTime mtime;
+  private static final MetadataTime m1234 = new MetadataTime(1234, TimeType.MILLIS);
+  private static final MetadataTime m5678 = new MetadataTime(5678, TimeType.MILLIS);
+  private static final MetadataTime l1234 = new MetadataTime(1234, TimeType.LOGICAL);
+  private static final MetadataTime l5678 = new MetadataTime(5678, TimeType.LOGICAL);
 
   @Before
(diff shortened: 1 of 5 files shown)


server/base/src/test/java/org/apache/accumulo/server/tablets/TabletTimeTest.java
```diff
@@ -33,6 +33,10 @@
 public class TabletTimeTest {
   private static final long TIME = 1234L;
   private MillisTime mtime;
+  private static final MetadataTime m1234 = new MetadataTime(1234, TimeType.MILLIS);
```
keith-turner: Seems like object like these could be created in MetadataTimeTest and that would make the test shorter.

hkeebler: agree, I was being lazy.

hkeebler: Last commit includes MetadataTimeTest cleanup. :)

keith-turner: > agree, I was being lazy.

I have written many hard to read test in the past.  Doing lots of code reviews has motivated me to spend more time making my test more readable.  My least favorite part of doing a code review is looking at the test, however I think a thorough review of the test pays future dividends so I do it.  When I saw how you made one test more readable I thought that would be nice in the other test.

PR #1343, opened 2019-08-29
Various improvements to InstanceOperationsImpl

milleruntime:
Mainly java 11 readability improvements


milleruntime:
These are fairly insignificant since they are just for readability so I will merge.

PR #1360, opened 2019-09-12
Improve concurrent access to gc statistics.

EdColeman:
This pull request is intended to allow safe, concurrent access to the gc metrics so they can be accessed by the metrics system.  The ultimate goal is to get the gcstats published to the metrics system to ease monitoring.  The changes are being made in stages to simply reviews.

Originally, some access was synchronized where needed, but this didn't allow for safe access from outside of the main gc thread - this corrects that deficiency.

### DIFF
--- a/server/gc/src/main/java/org/apache/accumulo/gc/SimpleGarbageCollector.java
+++ b/server/gc/src/main/java/org/apache/accumulo/gc/SimpleGarbageCollector.java
@@ -546,7 +546,6 @@ public class SimpleGarbageCollector extends AccumuloServerContext implements Ifa
   }
 
   private void run() {
-    long tStart, tStop;
 
     // Sleep for an initial period, giving the master time to start up and
     // old data files to be unused
@@ -592,7 +591,7 @@ public class SimpleGarbageCollector extends AccumuloServerContext implements Ifa
       Trace.on("gc", sampler);
 
       Span gcSpan = Trace.start("loop");
-      tStart = System.currentTimeMillis();
+      final long tStart = System.nanoTime();
       try {
         System.gc(); // make room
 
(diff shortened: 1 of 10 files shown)


server/gc/src/main/java/org/apache/accumulo/gc/SimpleGarbageCollector.java
```diff
@@ -865,6 +897,12 @@ static boolean isDir(String delete) {
 
   @Override
   public GCStatus getStatus(TInfo info, TCredentials credentials) {
-    return status;
+    statusLock.lock();
+    try {
+      GCStatus copy = new GCStatus(status);
```
milleruntime: I think these two lines are all you need to do for this change... make a copy and return. 

It seems like all you need is an immutable copy of the GCStatus object, which can be done without any locking.  The locking is not needed and just adds complexity.  No matter what you do, you will only ever be able to get a snapshot of the GC at a moment in time and we want to avoid making too many changes to 1.9.

You could just add a separate method like ```getSnapshot()``` which makes a copy of status.  This would make a merge forward easier as well.

keith-turner:
@EdColeman are you only interested in the last stats or the last and current stats.  Just getting the last stats could simple because those could be stored in an AtomicReference whenever they are swapped. 

The current stats are trickier because another thread is changing them.  Your locking approach will make sure the current stats are visible to another thread.  I was going to propose a way to simplify the locking code using lambdas, but then I realized this was java 1.7.

Another possible approach other than the locking is to not use the thrift objects inside the GC code.  Create our own simple objects with volatile counters.  Whenever the thrift object is requested, could materialize it from the internal status object.  Not proposing to do this, its just something I thought of while thinking through the solutions space.

Another option is to do nothing.  It seems like the monitor has been reading this information without any volatile or sync.  I suppose the monitor has probably reported stale information in some circumstances.  I have seen multithreaded code that did not use volatile for a boolean read stale information for tens of minutes.

server/gc/src/main/java/org/apache/accumulo/gc/SimpleGarbageCollector.java
EdColeman: The code that I think that also needs to lock is when the "last run" values are set to the current values and the zero'ed for the next run.
`   status.lastLog = status.currentLog;
     status.currentLog = new GcCycleStats();`

The rest of the locking was to maintain current behavior - and to make it consistent and correct for places where the current code is not locking.  It seemed odd to have portions that are thread safe and other portions that require single-thread access for correct behavior. This seemed a better practice, especially when having access to the snapshot is done from other thread(s)

EdColeman:
@keith-turner I'll post up an approach that uses local, thread safe stats and then converts to the thrift values when requested and we can discuss if that is a preferred path.

EdColeman:
Superseded by https://github.com/apache/accumulo/pull/1376

Issue #1351, opened 2019-09-06
Add test to GarbageCollectorIT for continue point or remove continue point


keith-turner:
Accumulo garbage collection will read a batch of deletes into memory and process them.  When reading the next batch it uses the last path it saw as a continue point.  Need to ensure the continue point code covers all data.

Alternatively, instead of testing the continue point code it could be removed by refactoring the GC code as follows so that continue is not needed.  The code below keeps the iterator for the entire cycle instead of calling continue.

```java
Iterator<String> candidates = getContext().getAmple().getGcCandidates(level);

while(candidates.hasNext()){
     List<String> candidatesBatch =  readCandidatesThatFitInMemory(candidates);
    
    collectBatch(candidatesBatch);
}
```

keith-turner:
https://github.com/apache/accumulo/pull/1344#discussion_r321812590

cradal:
Looking at this issue now.

PR #1366, opened 2019-09-18
Fix #1365 2.1 Upgrade processing for #1043 ~del

### LINKED ISSUE #1365: Develop upgrade/transition for #1043 updates
Update Upgrader9to10.upgradeMetadata() to convert any existing ~del entries to the new ~del entries.  Something like the following could be done:

for each key with ~del prefix and value.isEmpty(){
  create new del key with value("Converted on initialization")
  delete old delete entry
}

Needs to be done for both USER and METADATA metadata.

hkeebler:
This is fairly simple code to modify the ~del entries on the master upgrade startup.  The IT test is just for verification of the upgrade processsing and should probably not be included if this is actually merged.  Attempts to test this in a "real time" upgrade process were not successful due to errors unrelated to this code. (so she says ;))

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
@@ -35,13 +35,12 @@ import org.apache.accumulo.core.data.Mutation;
 import org.apache.accumulo.core.data.PartialKey;
 import org.apache.accumulo.core.data.Range;
 import org.apache.accumulo.core.data.TableId;
-import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.dataImpl.KeyExtent;
 import org.apache.accumulo.core.metadata.MetadataTable;
 import org.apache.accumulo.core.metadata.RootTable;
 import org.apache.accumulo.core.metadata.schema.Ample;
 import org.apache.accumulo.core.metadata.schema.AmpleImpl;
-import org.apache.accumulo.core.metadata.schema.MetadataSchema;
+import org.apache.accumulo.core.metadata.schema.MetadataSchema.DeletesSection;
 import org.apache.accumulo.core.security.Authorizations;
 import org.apache.accumulo.fate.zookeeper.ZooUtil;
 import org.apache.accumulo.server.ServerContext;
@@ -134,7 +133,7 @@ public class ServerAmpleImpl extends AmpleImpl implements Ample {
 
     try (BatchWriter writer = context.createBatchWriter(level.metaTable())) {
       for (String path : paths) {
-        Mutation m = new Mutation(MetadataSchema.DeletesSection.encodeRow(path));
+        Mutation m = new Mutation(DeletesSection.encodeRow(path));
         m.putDelete(EMPTY_TEXT, EMPTY_TEXT);
         writer.addMutation(m);
       }
@@ -155,9 +154,9 @@ public class ServerAmpleImpl extends AmpleImpl implements Ample {
 
       return candidates.iterator();
     } else if (level == DataLevel.METADATA || level == DataLevel.USER) {
-      Range range = MetadataSchema.DeletesSection.getRange();
+      Range range = DeletesSection.getRange();
       if (continuePoint != null && !continuePoint.isEmpty()) {
-        String continueRow = MetadataSchema.DeletesSection.encodeRow(continuePoint);
+        String continueRow = DeletesSection.encodeRow(continuePoint);
         range = new Range(new Key(continueRow).followingKey(PartialKey.ROW), true,
             range.getEndKey(), range.isEndKeyInclusive());
       }
@@ -170,8 +169,10 @@ public class ServerAmpleImpl extends AmpleImpl implements Ample {
       }
       scanner.setRange(range);
 
-      return Iterators.transform(scanner.iterator(),
-          entry -> MetadataSchema.DeletesSection.decodeRow(entry.getKey().getRow().toString()));
+      return Iterators.transform(
+          Iterators.filter(scanner.iterator(),
+              entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME)),
+          entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString()));
 
     } else {
       throw new IllegalArgumentException();
@@ -196,9 +197,8 @@ public class ServerAmpleImpl extends AmpleImpl implements Ample {
   public static Mutation createDeleteMutation(ServerContext context, TableId tableId,
       String pathToRemove) {
     Path path = context.getVolumeManager().getFullPath(tableId, pathToRemove);
-    Mutation delFlag =
-        new Mutation(new Text(MetadataSchema.DeletesSection.encodeRow(path.toString())));
-    delFlag.put(EMPTY_TEXT, EMPTY_TEXT, new Value(new byte[] {}));
+    Mutation delFlag = new Mutation(new Text(DeletesSection.encodeRow(path.toString())));
+    delFlag.put(EMPTY_TEXT, EMPTY_TEXT, DeletesSection.SkewedKeyValue.NAME);
     return delFlag;
   }
 
--- a/server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
@@ -171,11 +171,6 @@ public class ServerAmpleImpl extends AmpleImpl implements Ample {
       return StreamSupport.stream(scanner.spliterator(), false)
           .filter(entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME))
           .map(entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString())).iterator();
-      /*
-       * return Iterators.transform( Iterators.filter(scanner.iterator(), entry ->
-       * entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME)), entry ->
-       * DeletesSection.decodeRow(entry.getKey().getRow().toString()));
-       */
     } else {
       throw new IllegalArgumentException();
     }
--- a/server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
@@ -25,6 +25,7 @@ import java.util.Collection;
 import java.util.Iterator;
 import java.util.function.Consumer;
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import org.apache.accumulo.core.client.BatchWriter;
 import org.apache.accumulo.core.client.MutationsRejectedException;
@@ -50,7 +51,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.base.Preconditions;
-import com.google.common.collect.Iterators;
 
 public class ServerAmpleImpl extends AmpleImpl implements Ample {
 
@@ -168,12 +168,14 @@ public class ServerAmpleImpl extends AmpleImpl implements Ample {
         throw new RuntimeException(e);
       }
       scanner.setRange(range);
-
-      return Iterators.transform(
-          Iterators.filter(scanner.iterator(),
-              entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME)),
-          entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString()));
-
+      return StreamSupport.stream(scanner.spliterator(), false)
+          .filter(entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME))
+          .map(entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString())).iterator();
+      /*
+       * return Iterators.transform( Iterators.filter(scanner.iterator(), entry ->
+       * entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME)), entry ->
+       * DeletesSection.decodeRow(entry.getKey().getRow().toString()));
+       */
     } else {
       throw new IllegalArgumentException();
     }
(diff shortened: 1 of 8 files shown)


hkeebler:
Also wanted to mention that the code was pretty much taken from the SimpleGarbageCollector and Ample processing of gc delete candidates.

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
```diff
@@ -352,4 +372,87 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
...
+      range = new Range(new Key(continueRow).followingKey(PartialKey.ROW), true, range.getEndKey(),
+          range.isEndKeyInclusive());
+    }
+
+    Scanner scanner = ctx.createScanner(tableName, Authorizations.EMPTY);
+    scanner.setRange(range);
+
+    // find old candidates for deletion; chop off the prefix
+    for (Map.Entry<Key,Value> entry : scanner) {
+      if (!entry.getValue().toString().equals(UPGRADED)) {
```
keith-turner: I really like this approach of setting something in the value.  I had hoped everything related to this change could be isolated to the upgrade code, but not that I am looking at the code I see that it can't.  While the upgrade code is running two other processes could also happen concurrently.   The Accumulo GC may be running and Metadata tablets could be compacting.  The Accumulo GC would see the old and new del markers.  Metadata table compactions could write out new delete markers which this code could corrupt.

One possible solution to this would be to always put something in the value for new style delete markers, which you mentioned doing in #1344.  Then AMPLE code could always ignore old style and new delete markers written out during upgrade would not be modified.

hkeebler: I guess tservers are started first and gc could be started before the master.  Blast.  This is a simple enough fix but it would be nice if  setting the value had some meaningful value ... ignore the pun.  Maybe just "skewed" to indicate the key was altered.  Then down the road if the desire is to change things back these are easily identifiable.

keith-turner: Always putting `skewed` in the values sounds good to me.  Then Ample could ignore anything that does not have that in the value.

hkeebler:
Includes requested updates.  Ran -Psunny tests and related IT tests.

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
```diff
@@ -352,4 +372,87 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
...
+
+        // if nothing thrown then we're good so mark all deleted
+        for (String olddelete : deletes) {
+          writer.addMutation(deleteOldDeleteMutation(olddelete));
+        }
+        writer.flush();
+
+        // give it some time for memory to clean itself up if needed
+        sleepUninterruptibly(5, TimeUnit.SECONDS);
+        continuePoint = deletes.get(deletes.size() - 1);
```
keith-turner: There may be situations where  `deletes.size()` is empty and this will throw an exception.  Not sure about this, but it may make the code simpler to refactor this code as described in #1351 to remove the continue point.

test/src/main/java/org/apache/accumulo/test/upgrade/GCUpgradeTestIT.java
```diff
@@ -0,0 +1,207 @@
...
+      Range range = MetadataSchema.DeletesSection.getRange();
+
+      Scanner scanner;
+      try {
+        scanner = c.createScanner(level.metaTable(), Authorizations.EMPTY);
+      } catch (TableNotFoundException e) {
+        throw new RuntimeException(e);
+      }
+      scanner.setRange(range);
+      assertEquals(150000, Iterators.size(scanner.iterator()));
```
keith-turner: Could do a stronger check here than just the count

test/src/main/java/org/apache/accumulo/test/upgrade/GCUpgradeTestIT.java
```diff
@@ -0,0 +1,207 @@
...
+        throw new RuntimeException(e);
+      }
+      scanner.setRange(range);
+      scanner.iterator().forEachRemaining(entry -> {
+        actual.put(entry.getKey().getRow().toString(), entry.getValue().toString());
+      });
+
+      assertEquals(expected, actual);
+
+      // ENSURE IDEMPOTENCE - run upgrade again to ensure nothing is changed because there is
```
keith-turner: nice

server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
```diff
@@ -170,8 +169,10 @@ public void deleteGcCandidates(DataLevel level, Collection<String> paths) {
       }
       scanner.setRange(range);
 
-      return Iterators.transform(scanner.iterator(),
-          entry -> MetadataSchema.DeletesSection.decodeRow(entry.getKey().getRow().toString()));
+      return Iterators.transform(
```
keith-turner: Instead of using more Guava code here, this could use Java streams. I wan't sure how to get the stream for a scanner so I experimented locally.   I ended up with the following that at least compiles.

```java
  return StreamSupport.stream(scanner.spliterator(), false)
    .filter(entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME))
    .map(entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString())).iterator();
```

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
```diff
@@ -352,4 +372,87 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
...
+      ;
+    }
+  }
+
+  static boolean getOldCandidates(ServerContext ctx, String tableName, String continuePoint,
+      List<String> result) throws TableNotFoundException {
+
+    Range range = MetadataSchema.DeletesSection.getRange();
+    if (continuePoint != null && !continuePoint.isEmpty()) {
+      String continueRow = OLD_DELETE_PREFIX + continuePoint;
```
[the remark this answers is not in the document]
hkeebler: The range is still based on just "~del" ,  the section prefix before and after the upgrade. I previously took out:
```java
public static String getRowPrefix() {
      return section.getRowPrefix();
```
from MetadataSchema.Deletesection because it was only used in the upgrade.  I can put it back and use it instead of the constant.   The BlipSection has one o it was the norm.  
Did I address your concern here?

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
```diff
@@ -352,4 +372,87 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
...
+        }
+        writer.flush();
+
+        // give it some time for memory to clean itself up if needed
+        sleepUninterruptibly(5, TimeUnit.SECONDS);
+        continuePoint = deletes.get(deletes.size() - 1);
+        log.debug("continuing from {}", continuePoint);
+      }
+    } catch (Exception e) {
+      ;
```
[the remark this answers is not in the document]
hkeebler: Apologies, forgot go back and address this.

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
```diff
@@ -352,4 +372,87 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
...
+  static private Mutation deleteOldDeleteMutation(final String delete) {
+    Mutation m = new Mutation(OLD_DELETE_PREFIX + delete);
+    m.putDelete(EMPTY_TEXT, EMPTY_TEXT);
+    return m;
+  }
+
+  static private Mutation upgradeDeleteMutation(final String delete) {
+    // cannot use the Ample implementation because we want to modify the value field
+    Mutation m = new Mutation(MetadataSchema.DeletesSection.encodeRow(delete));
+    m.put(EMPTY_TEXT, EMPTY_TEXT, new Value(UPGRADED)); // new Value(new byte[] {}));
```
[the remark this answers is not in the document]
hkeebler: How do  you feel about updating Ample.DataLevel to include/return TableId?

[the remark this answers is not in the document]
keith-turner: > How do you feel about updating Ample.DataLevel to include/return TableId?

I am in favor.

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
hkeebler: @keith-turner the solution in 1351 works well.  If an Iterator reflects a point in time (immutable) then this could never loop unless memory never gets cleaned up.

server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
hkeebler: @keith-turner this also appears to work.  Why streams over iterators?  Curious.

keith-turner: > Why streams over iterators?

I just think its nice to use functionality built into Java vs a third party library.  The Guava iterator utilities were created before Streams existed in Java.

hkeebler:
Latest commit addresses recent comments.  Big items:
- refactor of continuepoint  usage
- added test for no deletes to process on upgrade

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
```diff
@@ -372,87 +374,77 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
...
-
-      while (stillDeletes) {
-        List<String> deletes = new ArrayList<>();
-        log.info("looking for candidates");
-        stillDeletes = getOldCandidates(ctx, tableName, continuePoint, deletes);
+      log.info("looking for candidates in table {}", tableName);
+      Iterator<String> oldCandidates = getOldCandidates(ctx, tableName);
+      while (oldCandidates.hasNext()) {
+        // give it some time for memory to clean itself up if needed
+        sleepUninterruptibly(5, TimeUnit.SECONDS);
```
keith-turner: This will sleep for the first loop iteration, which seems uneeded.   Not sure its worth adding complexity to code to avoid it.

server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
```diff
@@ -168,12 +168,14 @@ public void deleteGcCandidates(DataLevel level, Collection<String> paths) {
...
-      return Iterators.transform(
-          Iterators.filter(scanner.iterator(),
-              entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME)),
-          entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString()));
-
+      return StreamSupport.stream(scanner.spliterator(), false)
+          .filter(entry -> entry.getValue().equals(DeletesSection.SkewedKeyValue.NAME))
+          .map(entry -> DeletesSection.decodeRow(entry.getKey().getRow().toString())).iterator();
+      /*
+       * return Iterators.transform( Iterators.filter(scanner.iterator(), entry ->
```
keith-turner: commented code

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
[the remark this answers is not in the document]
keith-turner: I think I misunderstood the purpose of `OLD_DELETE_PREFIX`.  I thought it was for isolating the upgrade code from future schema changes.  However, I think it was created because the constant no longer exists in `DeletesSection`.

server/base/src/main/java/org/apache/accumulo/server/metadata/ServerAmpleImpl.java
hkeebler: blast.

server/master/src/main/java/org/apache/accumulo/master/upgrade/Upgrader9to10.java
hkeebler: yes.  I found that putting it at the beginning made better sense to free up memory.  Because the loop is holding onto the deletes structure, at the end caused it to continually process a single entry until something else caused some memory to free up.  Bottom line - it seemed to run more efficiently.  If it was permanent code I would worry more about it.

hkeebler: could do: 
```java
int t = 0;
while (oldCandidates.hasNext()) {
        // give it some time for memory to clean itself up if needed
        sleepUninterruptibly(t, TimeUnit.SECONDS);
.....
        t =5 (or even t += 1);
} //eof while
```

keith-turner: I like doing `t = 0` and `t=5`.  If doing `t+=1` would probably want to constraint it via a max.  Since its for Java GC I am not sure what a good time range is though.

hkeebler: I think I took it from a test or the GC code.  5 may be alot but it is "in theory" it is only upgrading once.

hkeebler: I guess this whole "avoid out of memory" error is a hit or miss - at least for the upgrader which is required to process all its data before finishing.  This last run took 9 minutes before memory was once again available (according to our code calculations).  In this case we assume full at 50% but the java GC is looking at the heap utilization.  Which in theory may not be effected if we stay below 50%, maybe.  Note that I need to set this number to 25% to make it happen.

hkeebler: Compromised on time to sleep at 3 sec.  Honestly it just depends on how many log messages you willing to look at.

hkeebler:
Removed comment and adjusted timeout.

hkeebler:
Since this has one approval , if no other comments received by Friday I will merge.  Thanks.

Issue #1316, opened 2019-08-08
Implement lifecycle logging


keith-turner:
Following the lifecycle of certain events in Accumulo via documented logical loggers would be really nice.  For example it would be nice to be able to follow the lifecycle of a write ahead log across the master, tserver, and GC using a logger name like `org.apache.accumulo.log.wal` that does not correspond to any class.  These logical loggers could be documented in the Accumulo user manual, like [Fluo does](https://fluo.apache.org/docs/fluo/1.2/administration/manage-applications#debugging-applications).

Could have life cycle logging for the following events to start with

 * Write ahead logs
 * Tablet events (like assignment, loading, compaction, scan, update, unload, recovery, etc).  All tablet events could be logged under `org.apache.accumulo.log.tablet`.  Something like scan could be logged under `org.apache.accumulo.log.tablet.scan` at trace allowing a user to turn on trace scanning for scans only.
 * FATE operations like bulk import could log everything related to bulk import in client, master, and tserver under `org.apache.accumulo.log.bulk`.

One way to implement this would be to create class  with static methods for each life-cycle we want to log.  For example all logging of tablet lifecycle events would be done in a single class via static methods.  The advantage of this approach is that it brings all of the logging together in one place in the code which will lead to more consistent log messages.

Most existing log messages should not follow this approach.

keith-turner:
Could use much shorter logger names like `accumulo.tablet` and `accumulo.wal` like Fluo does.

keith-turner:
In addition to documenting the loggers in the user manual, it would be nice to create a schema for these log events  that could be used by log stash apache/fluo-muchos#254

EdColeman:
I'm interested in taking a run at this as well as #1249 - not assigning myself at time incase other events delay me in starting to work the issue(s)

PR #1372, opened 2019-09-23
WIP #1316 Lifecycle logging

keith-turner:
I did some experimentation of what the structure for #1316 might look like.  Sharing this experimentation for discussion only.  For this PR I put the logging code in a single package intentionally.  The reason for this is it would make it easy for user to look at the source code for this single package and see the important log messages Accumulo generates.   Having it all in one place may also make it easier to generate a grok file for logstash, making a task like apache/fluo-muchos#254 easier.

I gave a lot of thought to the logical logger hierarchy, but its very incomplete.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/logging/BulkImportLogger.java
+++ b/core/src/main/java/org/apache/accumulo/core/logging/BulkImportLogger.java
@@ -1,0 +1,27 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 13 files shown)


EdColeman:
From a quick review, I like the approach.

Do you have any thoughts concerning a ZooKeeper logger - used in ZooReader, ZooStore,... that could log zookeeper operations - or would that be too low level and best just done with current debug logs?

hkeebler:
I like the tidyness and uniformity.  Logging can get messy but there is a loss of whether something is logged to trace, or debug, or error.  Maybe it could be reflected in the method call like logger.traceTabletMutated(...) .  And, I have to admit since @Manno15 and I are trying to understand the migration process for a ticket he is working being able to follow the "exact" log message back into the code is very helpful.

keith-turner:
> Do you have any thoughts concerning a ZooKeeper logger 

I might be in favor of that.  The only concern I would have is I would not want to duplicate any functionality the zookeeper client offers in terms of logging.  Also, would it be better to change zookeeper to have more useful logging rather than Accumulo?

> follow the "exact" log message back into the code is very helpful.

It is interesting that this code could become an accidental guide book into the most important parts of the Accumulo code.  One that is always kept up to date.

keith-turner:
This PR was created for feedback.  Took a first step in implementing this in #1480

PR #1400, opened 2019-10-29
Move AbstractLexicoder to API

milleruntime:
Curious as to why this is the only lexicoder class not in the public API... this change will assist in #1390

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/lexicoder/AbstractLexicoder.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/lexicoder/AbstractLexicoder.java
@@ -14,9 +14,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.accumulo.core.clientImpl.lexicoder;
-
-import org.apache.accumulo.core.client.lexicoder.AbstractEncoder;
-import org.apache.accumulo.core.client.lexicoder.Lexicoder;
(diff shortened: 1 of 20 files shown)


milleruntime:
Another option would be to drop this class altogether and have the lexicoders implement the Lexicoder interface directly

milleruntime:
> I had some reservations initially, because there's already two type hierarchies in the lexicoder package for public API, and I wasn't sure if this would complicate things, but I now think it's probably fine.

I am leaning towards just removing the class since it is in the impl package and we don't need both.

milleruntime:
> > I had some reservations initially, because there's already two type hierarchies in the lexicoder package for public API, and I wasn't sure if this would complicate things, but I now think it's probably fine.
> 
> I am leaning towards just removing the class since it is in the impl package and we don't need both.

After looking into this, I discovered that all the lexicoder tests use this class so I think moving it is fine.

milleruntime:
@ctubbsii @keith-turner before this change was made, there were many classes in the `org.apache.accumulo.core.client.lexicoder` package that extend `org.apache.accumulo.core.clientImpl.lexicoder.AbstractLexicoder`. I can't figure out why the apilyzer plugin didn't hit on the super of these classes being the AbstractLexicoder in the `clientImpl.lexicoder` package. Since the parent class is in a `clientImpl` package, I would think this would violate the API by exposing an internal type and get dinged by the plugin. Could this be a bug with the plugin? Any ideas?

PR #1404, opened 2019-10-31
Add example of client to javadoc

milleruntime:


core/src/main/java/org/apache/accumulo/core/client/Accumulo.java
```diff
@@ -51,11 +51,18 @@ private Accumulo() {}
...
    * <code>
+   * // Create client directly from connection information
    * try (AccumuloClient client = Accumulo.newClient()
    *        .to(instanceName, zookeepers)
    *        .as(user, password).build())
    * {
-   *   // use the client
+   *    // use the client
+   * }
+   * // Create client from java properties or properties file
```
keith-turner:
```suggestion
   * // Create client using the instance name, zookeeper, and credentials  from java properties or properties file 
```

core/src/main/java/org/apache/accumulo/core/client/Accumulo.java
```diff
@@ -51,11 +51,18 @@ private Accumulo() {}
...
+   * try (AccumuloClient client = Accumulo.newClient()
+   *        .from(properties).build())
+   * {
+   *    // connect as the user configured in properties
```
keith-turner:
```suggestion
   *    // use the client
```

core/src/main/java/org/apache/accumulo/core/client/Accumulo.java
```diff
@@ -51,11 +51,18 @@ private Accumulo() {}
...
    * <pre>
    * <code>
+   * // Create client directly from connection information
    * try (AccumuloClient client = Accumulo.newClient()
    *        .to(instanceName, zookeepers)
    *        .as(user, password).build())
    * {
-   *   // use the client
+   *    // use the client
+   * }
```
keith-turner: Personally, I would add an empty line after this block.

PR #1406, opened 2019-10-31
Backport gc metrics to 1.9.x to use hadoop metrics2.

EdColeman:
This backport bring the gc metrics from master / 2.0 into the 1.9.x branch. The metrics
can be enabled / disabled with GC_METRICS_ENABLED property (default false for 1.9.x)  The
metrics changes include:

- Publish current gc metrics to hadoop 2 metrics reporting system.
- Add gc run counter that increments on each gc cycle run
- Add metric to track time required for gc post op (compact, flush, none)

Also Includes:

- Move IT resource file for PR #1381 to correct directory so test passes
in CI and on the command-line. Previously, it may have only worked in
some IDEs, because of the way those IDEs put test resources on the class
path.


test/src/main/java/org/apache/accumulo/test/metrics/MetricsFileTailer.java
```diff
@@ -0,0 +1,261 @@
...
+      Configuration config = new PropertiesConfiguration(filename);
+
+      // Configurations configs = new Configurations();
+      //
+      // Configuration config = configs.properties(new File("config.properties"));
+      // FileBasedConfigurationBuilder<FileBasedConfiguration> builder =
+      // new FileBasedConfigurationBuilder<FileBasedConfiguration>(PropertiesConfiguration.class)
+      // .configure(params.fileBased().setFile(propertiesFile));
+      //
+      // Configuration config = builder.getConfiguration();
```
milleruntime: Should drop this commented out code.  Can be done in a separate commit.

test/src/main/java/org/apache/accumulo/test/functional/GcMetricsIT.java
```diff
@@ -0,0 +1,242 @@
...
+    MetricsFileTailer gcTail = new MetricsFileTailer("accumulo.sink.file-gc");
+    Thread t1 = new Thread(gcTail);
+    t1.start();
+
+    // used for manual jmx / jconsole validation - not for automated testing
+    // try {
+    // Thread.sleep(320_000);
+    // } catch (InterruptedException ex) {
+    // Thread.currentThread().interrupt();
+    // }
```
milleruntime: Should drop the commented code.  Can be done in a separate commit.

PR #1436, opened 2019-11-17
Fixes 1418: Remove DefaultMap, use computeIfAbsent

### LINKED ISSUE #1418: Remove DefaultMap and use computeIfAbsent instead.
I think [DefaultMap](https://github.com/apache/accumulo/blob/bbee47b8b95d230ba9def235917013dcfeaf7c38/server/base/src/main/java/org/apache/accumulo/server/util/DefaultMap.java) could be removed from the code.  There are currently two places in the code that create a DefaultMap.  I think these could instead create a HashMap and when `get()` is called on that hasmap use `computeIfAbsent()` creating whatever DefaultMap would have created.

Its only used in two places.  For the usage in `Master.getMasterMonitorInfo()` I think only the `get()` in that method needs to be replaced, nowhere else.  For the usage at `Master.badServers` I think only usage in `Master.gatherTableInformation()` needs to be updated.

ata18:
Fixes #1418 Remove DefaultMap and use computeIfAbsent instead


keith-turner:
Thanks for the contribution @ata18 and welcome to the Apache Accumulo community.  If you would like to be listed as a contributor on Accumulo's people page please make a website PR for [people.md](https://github.com/apache/accumulo-website/edit/master/pages/people.md).

If you intend to make more contributions, please consider subscribing to the [dev list](https://accumulo.apache.org/contact-us/) and introducing yourself.  Also, you can use the dev list or slack to reach out if you have questions about anything.

PR #1439, opened 2019-11-19
Fix #1417: Remove usage of Text from code creating mutations

### LINKED ISSUE #1417: Remove usage of Text from code creating mutations
While working on #1412 I noticed some unnecessary usage of Text in the creations of mutations, like at [InMemoryMapIT.java line 193](https://github.com/apache/accumulo/blob/6b54a5d89aa10103105db1a6c6c88903fcaa90e6/test/src/main/java/org/apache/accumulo/test/InMemoryMapIT.java#L193).  These usages are historical from a time long ago when mutation only supported text, now it supports charsequence.  

Find and sed w/ a regex could be used like in #1412 to replace calls matching `put(new Text(X), new Text(Y), new Value(Z))` with `put(X,Y,Z)`.  If after doing the replacement things do not compile, one possible reason could be that X,Y,Z are different types like `byte[]` and `String`.  These cases should be caught by the compiler. Maybe these could be manually refactored to use the new `at()` method added to mutation in 2.0 which easily supports mixing of types.

One thing to watch out for in this change is test for mutation that are explicitly testing the mutation methods with text types.  For example may want to checkout MutationTest after making the changes.

srajtiwari:


milleruntime:
Thanks for the contribution @srajtiwari and welcome to the Apache Accumulo community. If you would like to be listed as a contributor on Accumulo's people page please make a website PR for [people.md](https://github.com/apache/accumulo-website/edit/master/pages/people.md).

If you intend to make more contributions, please consider subscribing to the dev list and introducing yourself. Also, you can use the [dev list](https://accumulo.apache.org/contact-us/) or slack to reach out if you have questions about anything.

Issue #1350, opened 2019-09-06
Update ListVolumesUsed to support root tablet metadata.


keith-turner:
After the changes in #1313 the root tablets file metadata is no different from any other tablet.  Now that it is the same ListVolumesUsed could be updated to remove the specialized code it has for zookeeper and just call `listTable(Ample.DataLevel.ROOT, context);`.  

https://github.com/apache/accumulo/pull/1344#discussion_r321815325

PR #1472, opened 2020-01-10
Update ListVolumesUsed to support root table metadata (#1350)

jmark99:
With changes in #1313 the root tablet file metadata is no different from
any other tablet. Given that fact, ListVolumesUsed has been updated to
remove the specialized code that it had for zookeeper and instead make
use of the Ample API for interfacing with zookeeper.

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
@@ -116,7 +116,7 @@ public class ListVolumesUsed {
     try {
       listTable(Ample.DataLevel.ROOT, context);
     } catch (UnsupportedOperationException ex) {
-      // print nothing if no table name exists
+      System.out.println("\tNo volumes present");
     }
     System.out.println();
     listTable(Ample.DataLevel.METADATA, context);
--- a/server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
@@ -19,16 +19,12 @@
 package org.apache.accumulo.server.util;
 
 import java.util.Iterator;
-import java.util.Map.Entry;
 import java.util.TreeSet;
 
-import org.apache.accumulo.core.client.Scanner;
 import org.apache.accumulo.core.conf.SiteConfiguration;
-import org.apache.accumulo.core.data.Key;
-import org.apache.accumulo.core.data.Value;
 import org.apache.accumulo.core.metadata.schema.Ample;
-import org.apache.accumulo.core.metadata.schema.MetadataSchema;
-import org.apache.accumulo.core.security.Authorizations;
+import org.apache.accumulo.core.metadata.schema.TabletMetadata;
+import org.apache.accumulo.core.metadata.schema.TabletsMetadata;
 import org.apache.accumulo.core.tabletserver.log.LogEntry;
 import org.apache.accumulo.server.ServerContext;
 import org.apache.accumulo.server.fs.VolumeManager.FileType;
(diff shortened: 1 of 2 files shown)


server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
```diff
@@ -131,7 +113,11 @@ private static void listTable(Ample.DataLevel level, ServerContext context) thro
   }
 
   public static void listVolumes(ServerContext context) throws Exception {
-    listZookeeper(context);
+    try {
+      listTable(Ample.DataLevel.ROOT, context);
+    } catch (UnsupportedOperationException ex) {
+      // print nothing if no table name exists
+    }
```
milleruntime: Why are you suppressing the error here?  You could just let the error be thrown, like the other calls to ```listTable```

jmark99: If it's not caught, the command errors out without completing. The previous method just iterated over a collection of 0 elements and continued processing.

jmark99: The other option would be to print out that there are no volumes in ZK to display. But the previous method did not do so and I was attempting to mimic its behavior.

milleruntime: I see.  Yeah I would at least print there are no volumes, that would be helpful.

keith-turner:
@jmark99 the method listTable needs to be updated to not user a scanner.  Instead of using the scanner, it should do something like the following.  This compiles, but I have not tested.

```java
    TabletsMetadata tablets = TabletsMetadata.builder().forLevel(level)
        .fetch(ColumnType.FILES, ColumnType.LOGS).build(context);

    TreeSet<String> volumes = new TreeSet<>();

    for (TabletMetadata tabletMeta : tablets) {
      tabletMeta.getFiles().forEach(file -> volumes.add(getTableURI(file)));
      tabletMeta.getLogs().forEach(le -> getLogURIs(volumes, le));
    }
```

To write the above I had to add a method to the TabletsMetadata builder.  Below is the diff for that change.

```diff
diff --git a/core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java b/core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
index febc430..175fbb9 100644
--- a/core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
+++ b/core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
@@ -195,6 +195,13 @@ public class TabletsMetadata implements Iterable<TabletMetadata>, AutoCloseable
     }
 
     @Override
+    public Options forLevel(DataLevel level) {
+      this.level = level;
+      this.range = TabletsSection.getRange();
+      return this;
+    }
+
+    @Override
     public TableRangeOptions forTable(TableId tableId) {
       if (tableId.equals(RootTable.ID)) {
         this.level = DataLevel.ROOT;
@@ -267,6 +274,11 @@ public class TabletsMetadata implements Iterable<TabletMetadata>, AutoCloseable
   public interface TableOptions {
 
     /**
+     * Read all of the tablet metadata for this level.
+     */
+    Options forLevel(DataLevel level);
+
+    /**
      * Get the tablet metadata for this extents end row. This should only ever return a single
      * tablet. No checking is done for prev row, so it could differ.
      */
```

Looking at this I realized that `TabletsMetadata.builder()` should probably hang off of Ample for discover-ability.

server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
```diff
@@ -131,7 +113,11 @@ private static void listTable(Ample.DataLevel level, ServerContext context) thro
   }
 
   public static void listVolumes(ServerContext context) throws Exception {
-    listZookeeper(context);
+    try {
+      listTable(Ample.DataLevel.ROOT, context);
+    } catch (UnsupportedOperationException ex) {
```
keith-turner: After the listTable method is updated to not use a scanner, I do not think this exception needs to be caught anymore.

keith-turner:
I opened #1473

server/base/src/main/java/org/apache/accumulo/server/util/ListVolumesUsed.java
jmark99: Thanks, @keith-turner. I will take a look at the listTable method and continue working on this.

keith-turner:
@jmark99 did you try running the command?

jmark99:
@keith-turner I successfully ran the 'sunny' test suite and using an Uno instance ran the 'accumulo admin volumes -l' command as well. 

<pre>
 accumulo admin volumes -l
2020-01-13 13:54:43,241 [conf.SiteConfiguration] INFO : Found Accumulo configuration on classpath at /home/mark/dev/fluo-uno/install/accumulo-2.1.0-SNAPSHOT/conf/accumulo.properties
Listing volumes referenced in ROOT tablets section
        Volume : hdfs://localhost:8020/accumulo
Listing volumes referenced in ROOT deletes section (volume replacement occurs at deletion time)
        Volume : hdfs://localhost:8020/accumulo
Listing volumes referenced in ROOT current logs
        Volume : hdfs://localhost:8020/accumulo

Listing volumes referenced in METADATA tablets section
        Volume : hdfs://localhost:8020/accumulo
Listing volumes referenced in METADATA deletes section (volume replacement occurs at deletion time)
        Volume : hdfs://localhost:8020/accumulo
Listing volumes referenced in METADATA current logs
        Volume : hdfs://localhost:8020/accumulo

Listing volumes referenced in USER tablets section
        Volume : hdfs://localhost:8020/accumulo
Listing volumes referenced in USER deletes section (volume replacement occurs at deletion time)
Listing volumes referenced in USER current logs
        Volume : hdfs://localhost:8020/accumulo
</pre>

milleruntime:
This was merged in d6ac765e91442c4e1f4953f882b9daf2aa243ee4. Closing

PR #1498, opened 2020-02-03
Reduce Path obj creation with VolumeUtil

milleruntime:
* Make VolumeUtil.switchVolume() return the Path object it created
instead of returning a string. This allows callers to use the Path
object or call toString if one is needed. There are a few cases where
the string returned was used to create another Path object


milleruntime:
Thanks @EdColeman !

PR #1501, opened 2020-02-06
Use TabletFile for Tablet, Scans and compaction

milleruntime:
* Made use of TabletFile from Ample through Tablet code, compactions
and scans where object represents a file in a Tablet
* Moved TabletFile to metadata package
* Added TabletFile constructor for the cases when switching volumes and
bulk import where we aren't reading a string from the metadata but have
to create a Path object to resolve the absolute path. That Path object
gets passed to TabletFile so multiple Paths don't need to be created
* Garbage collector will still use strings for delete candidates since
they can be a tablet directory


server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
```diff
@@ -159,14 +160,14 @@ public static TabletFiles updateTabletVolumes(ServerContext context, ZooLock zoo
...
-      String metaPath = entry.getKey().meta().toString();
+    for (Entry<TabletFile,DataFileValue> entry : tabletFiles.datafiles.entrySet()) {
+      String metaPath = entry.getKey().getNormalizedPath();
       Path switchedPath = switchVolume(metaPath, FileType.TABLE, replacements);
       if (switchedPath != null) {
         filesToRemove.add(entry.getKey());
-        FileRef switchedRef = new FileRef(switchedPath.toString(), switchedPath);
-        filesToAdd.put(switchedRef, entry.getValue());
-        ret.datafiles.put(switchedRef, entry.getValue());
+        TabletFile switchedFile = new TabletFile(switchedPath, switchedPath.toString());
```
keith-turner: When I looked at this code I found myself wondering if any follow on code relies on the metadata entry for the TabletFile created here.  If so that could result in a silent bug (where we try to delete something in the metadata table that does not exists).  This makes me think it would be nice to be able to construct a TabletFile w/o a metadata entry and anything trying to get the metadata path would get an exception.  This would cause things to fail fast.   

Could possibly have a TabletFile constructor that only takes a path and when this constructor is used to create a the object, attempts to get the metadata entry result in an exception.

server/base/src/main/java/org/apache/accumulo/server/util/MetadataTableUtil.java
```diff
@@ -247,44 +247,48 @@ public static void splitTablet(KeyExtent extent, Text oldPrevEndRow, double spli
...
+  public static void finishSplit(KeyExtent extent, Map<TabletFile,DataFileValue> datafileSizes,
+      List<TabletFile> highDatafilesToRemove, ServerContext context, ZooLock zooLock) {
     finishSplit(extent.getMetadataEntry(), datafileSizes, highDatafilesToRemove, context, zooLock);
   }
 
-  public static void addDeleteEntries(KeyExtent extent, Set<FileRef> datafilesToDelete,
+  /**
+   * datafilesToDelete are strings because they can be a TabletFile or directory
+   */
+  public static void addDeleteEntries(KeyExtent extent, Set<String> datafilesToDelete,
```
keith-turner: Why did this end up `Set<String>`?

server/base/src/main/java/org/apache/accumulo/server/util/MetadataTableUtil.java
```diff
@@ -414,7 +420,7 @@ public static void deleteTable(TableId tableId, boolean insertDeletes, ServerCon
 
     result.addAll(tablet.getLogs());
 
-    tablet.getFilesMap().forEach((tf, v) -> sizes.put(new FileRef(tf.getMetadataEntry()), v));
+    tablet.getFilesMap().forEach(sizes::put);
```
keith-turner: That is kinda of a neat that `sizes::put` can be done now.

server/base/src/test/java/org/apache/accumulo/server/util/FileUtilTest.java
```diff
@@ -66,13 +66,13 @@ public void createTmpDir() throws IOException {
...
-    FileRef r1 = createMock(FileRef.class);
-    expect(r1.path()).andReturn(new Path("/foo"));
+    Collection<TabletFile> c = new java.util.ArrayList<>();
+    TabletFile r1 = createMock(TabletFile.class);
+    expect(r1.getNormalizedPath()).andReturn("/foo");
     replay(r1);
     c.add(r1);
-    FileRef r2 = createMock(FileRef.class);
-    expect(r2.path()).andReturn(new Path("/bar"));
+    TabletFile r2 = createMock(TabletFile.class);
```
keith-turner: This test seems suspicious.  I am wondering if its still needed.  I will look into this later.

server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
```diff
@@ -611,12 +609,11 @@ private void deleteTablets(MergeInfo info) throws AccumuloException {
...
       scanner.fetchColumnFamily(DataFileColumnFamily.NAME);
       scanner.fetchColumnFamily(TabletsSection.CurrentLocationColumnFamily.NAME);
-      Set<FileRef> datafiles = new TreeSet<>();
+      Set<String> datafiles = new TreeSet<>();
       for (Entry<Key,Value> entry : scanner) {
         Key key = entry.getKey();
         if (key.compareColumnFamily(DataFileColumnFamily.NAME) == 0) {
-          datafiles
-              .add(new FileRef(TabletFileUtil.validate(key.getColumnQualifierData().toString())));
+          datafiles.add(key.getColumnQualifierData().toString());
```
keith-turner: Why drop validation here?

server/tserver/src/main/java/org/apache/accumulo/tserver/compaction/MajorCompactionRequest.java
```diff
@@ -105,7 +105,7 @@ public MajorCompactionReason getReason() {
     return reason;
   }
 
-  public Map<FileRef,DataFileValue> getFiles() {
+  public Map<TabletFile,DataFileValue> getFiles() {
```
keith-turner: This class is used by CompactionStrategy which is a pluggable component.  So this changes could break existing compaction strategy plugins.  

I think this change is ok because with these types being outside of SPI, its hard to analyze in a rigorous way.   It would not surprise me if there were already other breaking changes.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
milleruntime: That is a good point.  The volume replacement code (which is where this is used) will delete the original file but I don't see where it will write out the new path.  Is this done during compaction?  If so, then we will have problems if we throw an exception since the string won't ever exist in the metadata.

server/base/src/main/java/org/apache/accumulo/server/util/MetadataTableUtil.java
milleruntime: This method was bumping into the GC code that uses tabletDirs.  One possible code path for a file to be added to those datafilesToDelete comes from ```GcVolumeUtil.getDeleteTabletOnAllVolumesUri()```  [Here](https://github.com/apache/accumulo/blob/8ea08f8ef59cd199ec23d414d8dbc89be7eb1667/server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java#L627)

server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
milleruntime: Hmm this might have been a mistake.  I had this code using TabletFile at one point but then had to revert to Strings because of the GC code, per my other comment.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
keith-turner: > but I don't see where it will write out the new path. Is this done during compaction?

I am not sure, was there a specific class you were looking at?  I can take a look.


> If so, then we will have problems if we throw an exception since the string won't ever exist in the metadata.

It would be annoying if making TabletFile more strict causes code that is working for the wrong reasons to break, but it would be nice to cause code that is not working but failing silently to break.

milleruntime:
> I finished reviewing, the github feature to mark files as viewed was invaluable for this review. Is the only think using FileRef now write ahead log code?

Thanks for the review.  

I don't see where WAL code is using FileRef, maybe you are thinking of LogEntry?  LogEntry takes a String for filename, which I think may be another good case for TableFile, if the WAL paths are the same.

FileRef is still used by the two parts I found that use directories - the GC and the old bulkImport CopyFailed FATE operation.   I thought of creating an object for TabletDir for these cases but I don't think its worth it (I couldn't come up with a graceful reuse of TabletFile without dropping ```final``` on its members) since I think we can just use Path for these.

server/base/src/test/java/org/apache/accumulo/server/util/FileUtilTest.java
keith-turner: I think this test is ok, its just making sure that a method in FileUtil calls a very specific method on TabletFile.  I think the method it is testing could be replaced with a one line stream operation.

keith-turner:
> I don't see where WAL code is using FileRef, maybe you are thinking of LogEntry?

I was just guessing at what was using it.  Mainly I was curious FileRef was not deleted, I should have just asked that.  You already answered that question.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
milleruntime: > > but I don't see where it will write out the new path. Is this done during compaction?
> 
> I am not sure, was there a specific class you were looking at? I can take a look.

Just this method, ```VolumeUtil.updateTabletVolumes()``` and the only place its called, [Tablet](https://github.com/apache/accumulo/blob/8ea08f8ef59cd199ec23d414d8dbc89be7eb1667/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java#L342).

server/base/src/main/java/org/apache/accumulo/server/util/MetadataTableUtil.java
keith-turner: Thanks for the pointer.

server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
milleruntime: Fixed in 3aacfe603ad9c5d432df1da6411297783adba004

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeUtil.java
keith-turner: I was looking at the code and it seems like there may be an existing bug.  I opened #1501

PR #1536, opened 2020-02-27
Improve master metrics reported via hadoop metrics

EdColeman:
Adds FATE transaction state and op types
Improves metrics tests
Add IT to verify that enabling legacy metrics does not crash master


milleruntime:
I am taking a look at this pull request but wanted to clarify that these changes are new to both 1.9 and master?

test/src/main/java/org/apache/accumulo/test/functional/LegacyMetricsIT.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  /**
+   * Pause for manual testing - allow time to connect with jconsole after launch.
+   *
+   * @param sleep_ms
+   *          sleep duration in milliseconds
+   */
+  private void manual_test_pause(int sleep_ms) {
```
milleruntime: Method is unused.  You could always call it in the test since you are checking the boolean.  You could also make the one in MastMetricsIT static to reuse it between the ITs and pass in the boolean.

test/src/main/java/org/apache/accumulo/test/functional/LegacyMetricsIT.java
```diff
@@ -0,0 +1,138 @@
...
+    assertTrue(legacyMetricsEnabled);
+
+    List<SlowOps> tables = new ArrayList<>();
+
+    for (int i = 0; i < tableCount; i++) {
+      String uniqueName = getUniqueNames(1)[0] + "_" + i;
+      SlowOps gen = new SlowOps(getConnector(), uniqueName, maxWait, tableCount);
+      tables.add(gen);
+      gen.startCompactTask();
+    }
```
milleruntime: Do you need the SlowOps to just verify Master is up and running?  Seems like this would just make the IT run longer.

test/src/main/java/org/apache/accumulo/test/functional/MasterMetricsIT.java
```diff
@@ -0,0 +1,372 @@
...
+  }
+
+  @Before
+  public void setup() {
+
+    if (testDisabled()) {
+      return;
+    }
+
+    // accumuloClient = Accumulo.newClient().from(getClientProps()).build();
```
milleruntime: A glimpse into the future?? Haha

test/src/main/java/org/apache/accumulo/test/functional/MasterMetricsIT.java
```diff
@@ -0,0 +1,372 @@
...
+    boolean fateMetricsEnabled =
+        cluster.getSiteConfiguration().getBoolean(Property.MASTER_FATE_METRICS_ENABLED);
+
+    boolean useLegacyMetrics =
+        cluster.getSiteConfiguration().getBoolean(Property.GENERAL_LEGACY_METRICS);
+
+    if (!fateMetricsEnabled || useLegacyMetrics) {
+
+      log.info("master fate metrics are disabled - MASTER_FATE_METRICS_ENABLED={}, "
+          + "GENERAL_LEGACY_METRICS={}", fateMetricsEnabled, useLegacyMetrics);
```
milleruntime: Shouldn't this configuration validation be done in MasterMetricsFactory when the properties are read by the Master?  I didn't see any actual check for this.

server/master/src/main/java/org/apache/accumulo/master/metrics/fate/FateLegacyJMXMetricsMBean.java
```diff
@@ -16,12 +16,12 @@
  */
 package org.apache.accumulo.master.metrics.fate;
 
-public interface FateMetricsMBean {
+public interface FateLegacyJMXMetricsMBean {
```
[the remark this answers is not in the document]
EdColeman: This was a rename so that on the merge forward, these JMX specific classes are drooped and hadoop2 metrics classes the can be named FateMetrics.  The interface name needed to change to match the JMX required naming convention

test/src/main/java/org/apache/accumulo/test/functional/MasterMetricsIT.java
EdColeman: The tests are only valid with hadoop2 metrics - this prevents test failures if legancy metrics were enabled during the testing.  Added a warning if legacy metrics are enabled that additional FATE metrics added by this PR will not be available.

test/src/main/java/org/apache/accumulo/test/functional/LegacyMetricsIT.java
EdColeman: Being that I must be the only one that finds these markers helpful for recalling a good place to put a pause for manual testing with jconsole, I've removed the code.

test/src/main/java/org/apache/accumulo/test/functional/LegacyMetricsIT.java
EdColeman: Changed to use table create / delete so master / tablets did have some work required.  The compaction only ran for a short while, but create/delete should be quicker.

PR #1538, opened 2020-02-28
Fix DatafileManager MajC for empty files. Closes #1535

### LINKED ISSUE #1535: Investigate empty files in Major Compactions
I noticed while working on #1519 that if a Major compaction produces an empty file that it will still be added to the list of majorCompactingFiles.  This happens [here](https://github.com/apache/accumulo/blob/1.9/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java#L596). This is different from the rest of the code which checks if ```DataFileValue.getNumEntries() > 0```.  This file will be deleted towards the beginning of ```bringMajorCompactionOnline()``` [here](https://github.com/apache/accumulo/blob/1.9/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java#L541).  The list files is cleared at the end of the MajC but there parts of the code which check the size of the list that this could effect.

If there is no reason why we care about an empty file that is already deleted, then this is a bug and the file shouldn't be added to the list.

The place were this file is deleted for MajC also seems sub-optimal since we are renaming the file immediately before removing it.  Minor compactions will do one or the other.

milleruntime:
* Don't rename the file if it is empty
* Don't add the file to the list of files if it is empty

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
@@ -500,12 +500,13 @@ class DatafileManager {
       throw new IllegalStateException("Target map file already exist " + newDatafile);
     }
 
-    // rename before putting in metadata table, so files in metadata table should
-    // always exist
-    rename(tablet.getTabletServer().getFileSystem(), tmpDatafile.getPath(), newDatafile.getPath());
-
     if (dfv.getNumEntries() == 0) {
       tablet.getTabletServer().getFileSystem().deleteRecursively(newDatafile.getPath());
+    } else {
+      // rename before putting in metadata table, so files in metadata table should
+      // always exist
+      rename(tablet.getTabletServer().getFileSystem(), tmpDatafile.getPath(),
+          newDatafile.getPath());
     }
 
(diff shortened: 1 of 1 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
```diff
@@ -500,12 +500,13 @@ StoredTabletFile bringMajorCompactionOnline(Set<StoredTabletFile> oldDatafiles,
       throw new IllegalStateException("Target map file already exist " + newDatafile);
     }
 
-    // rename before putting in metadata table, so files in metadata table should
-    // always exist
-    rename(tablet.getTabletServer().getFileSystem(), tmpDatafile.getPath(), newDatafile.getPath());
-
     if (dfv.getNumEntries() == 0) {
       tablet.getTabletServer().getFileSystem().deleteRecursively(newDatafile.getPath());
```
keith-turner:
```suggestion
      tablet.getTabletServer().getFileSystem().deleteRecursively(tmpDatafile.getPath());
```

milleruntime:
Thanks @keith-turner

PR #1547, opened 2020-03-04
Drop obscuring overrides in thrift client code

milleruntime:
* Inline method overrides only called once or twice. They obscure actual
method calls and bloat the code


milleruntime:
I ran "mvn clean verify javadoc:jar -DskipITs" locally and it built successfully.

PR #1548, opened 2020-03-05
Document transport pool unreserved queue

keith-turner:


keith-turner:
The formatter does not mess with it.  I had to put empty lines between the bullet points to keep the formatter from squishing them all into a paragraph.

milleruntime:
Related to these changes, tests were added to the TransportCachingIT [here](https://github.com/apache/accumulo/pull/1241/files#diff-a8fae6237f477626112730e174b19a78R144-R166) but I don't believe they are reliable tests or testing the behaviour you updated.   The method called ```getAnyTransport``` in the IT can return a random entry of the server set, which may not be the one the test is expecting.

milleruntime:
> Related to these changes, tests were added to the TransportCachingIT [here](https://github.com/apache/accumulo/pull/1241/files#diff-a8fae6237f477626112730e174b19a78R144-R166) but I don't believe they are reliable tests or testing the behaviour you updated. The method called `getAnyTransport` in the IT can return a random entry of the server set, which may not be the one the test is expecting.

Occasional failure can be seen here:
 https://github.com/apache/accumulo/issues/1542

PR #1552, opened 2020-03-06
Fix and simplify TransportCachingIT

keith-turner:


test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
```diff
@@ -18,87 +18,53 @@
...
-          } else {
-            log.error("Failed to add ThriftTransportKey servers - Failing TransportCachingIT test");
-            org.junit.Assert
-                .fail("Failed to add ThriftTransportKey servers - Failing TransportCachingIT test");
-          }
-        }
-      }
+      List<ThriftTransportKey> servers = Lists.transform(tservers,
+          serverStr -> new ThriftTransportKey(HostAndPort.fromString(serverStr), rpcTimeout,
+              context));
```
ctubbsii: This might snag on the modernizer plugin. Even if it doesn't, I think the pure Java streams form looks a little cleaner (or maybe I've just been using streams a lot to have acquired an affinity for them):

```suggestion
      List<ThriftTransportKey> servers = tservers.stream().map(serverStr -> {
        return new ThriftTransportKey(HostAndPort.fromString(serverStr), rpcTimeout, context);
      }).collect(Collectors.toList());
```
(not necessarily formatted)

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
```diff
@@ -158,7 +124,7 @@ public void testCachedTransport() throws InterruptedException {
       while (fifth == null) {
         try {
           // Get a cached transport
-          fifth = pool.getAnyTransport(servers, true).getSecond();
+          fifth = pool.getAnyTransport(servers.subList(0, 1), true).getSecond();
```
ctubbsii: Do we want this to loop indefinitely? If this is null, could it ever become non-null?

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
keith-turner: I doubt modernizer would complain because there is no exact analogue.  Using streams requires copying/allocating a new list where Lists.transform does not.  However in this case performance does not matter so using streams is better.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
```diff
@@ -18,87 +18,53 @@
...
-  private static int ATTEMPTS = 0;
 
   @Test
   public void testCachedTransport() throws InterruptedException {
     try (AccumuloClient client = Accumulo.newClient().from(getClientProps()).build()) {
-      while (client.instanceOperations().getTabletServers().isEmpty()) {
+
+      List<String> tservers;
+
+      while ((tservers = client.instanceOperations().getTabletServers()).isEmpty()) {
```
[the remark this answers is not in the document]
keith-turner: I don't think the test needs more than one tserver.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
keith-turner: The looping seemed kinda sketchy to me, but I didn't look into it in detail.

keith-turner: If on the first attempt it gets a TTransportException and on the 2nd attempt it succeeds then it seems like it could become non-null, right?  So maybe the looping is not so sketchy after all.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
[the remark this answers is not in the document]
ctubbsii: In that case, we should override mini's config to only start up one.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
ctubbsii: I'm not really sure. I was confused by what the test was trying to do, and the possible outcomes it was trying to account for with the loop, since none of that is documented :smiley_cat:
I will defer to you.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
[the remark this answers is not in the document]
keith-turner: I think two are fine now.  I would prefer for test to not have config unless needed.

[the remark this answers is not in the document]
ctubbsii: That's fair.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
```diff
@@ -18,87 +18,53 @@
...
-import org.apache.accumulo.fate.zookeeper.ZooCache;
+import org.apache.accumulo.core.util.HostAndPort;
 import org.apache.accumulo.harness.AccumuloClusterHarness;
 import org.apache.thrift.transport.TTransport;
 import org.apache.thrift.transport.TTransportException;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.collect.Lists;
```
ctubbsii: I think the imports need to be cleaned up after applying the suggested change to use Java streams. After that, this PR should be good to merge.

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
```diff
@@ -148,7 +114,7 @@ public void testCachedTransport() throws InterruptedException {
       while (fourth == null) {
         try {
           // Get a non-cached transport
-          fourth = pool.getAnyTransport(servers, false).getSecond();
+          fourth = pool.getAnyTransport(servers.subList(0, 1), false).getSecond();
```
milleruntime: How does getting the same transport from a list of one test anything?  The last ```assertSame``` now doesn't seem like it is testing anything.

ctubbsii: I think the test is similar between the first/second. Where those were checking basic caching behavior, this one was specifically testing LIFO behavior, as mentioned in the comment above: *// ensure the LIFO scheme with a fourth and fifth entry*

However, there are still some nuances I don't understand. Like this fourth seems like it is supposed to be for the same server as one of first, second, or third. Fifth is supposed to make sure that it matches fourth, and not the same server from first/second/third (LIFO behavior). However, because we have more than one server, it doesn't seem like there's any guarantee that fourth/fifth refer to the same server as one of first/second/third, so this test might occasionally not test LIFO behavior, and basically do the same as first/second (basic caching) but for a different server than that one chose.

I think some of this can be resolved if we select `servers.subList(0,1)` at the beginning of the test, and stick with that one the whole time, so we make sure we're testing with the same server.

keith-turner: I made the following improvements to the test

- Use servers.subList(0,1) for everything
- Strengthened the LIFO check by pulling three things off to check for LIFO order instead of just one
- Pulled the redundant while loops into a function
- Improved the comments

test/src/main/java/org/apache/accumulo/test/TransportCachingIT.java
```diff
@@ -49,122 +45,78 @@
...
+
+      // The following three asserts ensure the per server queue is LIFO
+      TTransport fifth = getAnyTransport(servers, pool, true);
       assertSame("Expected fourth and fifth transport to be the same instance", fourth, fifth);
+
+      TTransport sixth = getAnyTransport(servers, pool, true);
+      assertSame("Expected third and sixth transport to be the same instance", third, sixth);
+
+      TTransport seventh = getAnyTransport(servers, pool, true);
+      assertSame("Expected third and sixth transport to be the same instance", second, seventh);
```
ctubbsii: Comment is incorrect:
```suggestion
      assertSame("Expected second and seventh transport to be the same instance", second, seventh);
```

PR #1571, opened 2020-03-25
Fix ArrayIndexOutOfBoundsException in importdir

arvindshmicrosoft:
**Fix details**
The importdirectory command implementation within the shell now checks for the expected number of arguments and safely handles the case where no arguments are supplied for importdirectory.

**Bug details**
The issue is trivial to reproduce (in master and in publicly released Accumulo 2.0.0 as well). Just switch to a table context within the Accumulo shell and issue an importdirectory command without any arguments:
```
root@uno t1> importdirectory 
2020-03-24T22:16:16,107 [shell.Shell] ERROR: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
```


milleruntime:
Thanks for the fix @arvindshmicrosoft

PR #1594, opened 2020-04-24
Handle spurious failure were rename succeeded but reported it failed (#1588)

### LINKED ISSUE #1588: Investigate avoiding file rename in major compactions
Opening this in response to a discussion on slack about running Accumulo on S3.  Major compactions create files with a `tmp` suffix and rename the file when its complete before adding an entry to the metadata table.  It seems like this rename could be avoided.

The only benefit I know of the current scheme is that it allows tablets to delete any tmp files from hdfs on load.  Would need to to determine how files from failed compactions could be cleaned up.  Need to determine if there are any other benefits.

tushardhadiwal:
Fix #1588 : Refine rename during minor compaction
PR addresses a few issues:	

- Spurious rename failures where rename failed but actually succeeded
- The check for file existence should only be done before attempting rename
- Should not delete file if it exists (an admin may want to investigate).

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
@@ -329,16 +329,26 @@ class DatafileManager {
     StoredTabletFile newFile;
     // rename before putting in metadata table, so files in metadata table should
     // always exist
+    boolean attemptedRename = false;
     do {
       try {
         if (dfv.getNumEntries() == 0) {
+
           tablet.getTabletServer().getFileSystem().deleteRecursively(tmpDatafile.getPath());
         } else {
-          if (tablet.getTabletServer().getFileSystem().exists(newDatafile.getPath())) {
+          if (!attemptedRename
+              && tablet.getTabletServer().getFileSystem().exists(newDatafile.getPath())) {
             log.warn("Target map file already exist {}", newDatafile);
-            tablet.getTabletServer().getFileSystem().deleteRecursively(newDatafile.getPath());
+            throw new RuntimeException("File unexpectedly exists " + newDatafile.getPath());
           }
-
+          // the following checks for spurious rename failures that succeeded but gave an IoE
+          if (attemptedRename
+              && tablet.getTabletServer().getFileSystem().exists(newDatafile.getPath())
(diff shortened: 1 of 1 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
```diff
@@ -329,23 +329,34 @@ void bringMinorCompactionOnline(TabletFile tmpDatafile, TabletFile newDatafile,
...
+    boolean attemptedRename = false;
     do {
       try {
         if (dfv.getNumEntries() == 0) {
+
           tablet.getTabletServer().getFileSystem().deleteRecursively(tmpDatafile.getPath());
-        } else {
-          if (tablet.getTabletServer().getFileSystem().exists(newDatafile.getPath())) {
+        } else // i.e. dfv.getNumEntries() > 0
+        {
```
arvindshmicrosoft, line 339: @tushardhadiwal some minor comments:
* Please modify the placement of this curly brace to match the project's expected style, otherwise the CI fails. The [patch submission steps](https://github.com/apache/accumulo/blob/master/CONTRIBUTING.md#patch-submission) are very useful - ```mvn clean verify -DskipTests``` will usually auto-format code and fix this
* Related to above - personally I'd also remove the // i.e. dfv.getNumEntries() > 0 comment :-)

keith-turner:
I think #1588 is a more general issue that could still be beneficial and be done at some point in the future.  Could change the commit message for this PR to something like `Handle spurious failure were rename succeeded but reported it failed` and omit #1588 or just reference #1588 in the commit message without the preceding `fix`

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/DatafileManager.java
tushardhadiwal: @arvindshmicrosoft I have removed the comment & fixed the curly brace.

keith-turner:
@tushardhadiwal  thanks for the contribution.  If you would like to be listed as a contributor on the [people page](https://accumulo.apache.org/people), please [submit a PR to add yourself](https://github.com/apache/accumulo-website/edit/master/pages/people.md).  Let me know if you have any questions about editing the website.

PR #1595, opened 2020-04-24
Update manual array to collections copy method.

jmark99:
Replace instances of copying of array contents to a collection where each element is added individually using a for loop with a call to Collections.addAll().

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/mapreduce/lib/InputConfigurator.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/mapreduce/lib/InputConfigurator.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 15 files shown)


core/src/main/java/org/apache/accumulo/core/data/Key.java
```diff
@@ -1159,7 +1159,7 @@ public int getSize() {
   }
 
   private static boolean isEqual(byte[] a1, byte[] a2) {
-    if (a1 == a2)
+    if (Arrays.equals(a1, a2))
```
EdColeman, line 1162: I'm not sure this keeps the same behaviour. The == will test if the passed arrays are the same reference.  Replacing with Arrays.equals is also going to check if the contents are the same.  This would make the rest of the test redundant - and from the code comments is trying to optimize the comparison.

jmark99: Thanks for the catch Ed! I will revert that change. It didn't really go with the rest of the changes anyway.

core/src/main/java/org/apache/accumulo/core/clientImpl/mapreduce/lib/InputConfigurator.java
```diff
@@ -360,9 +361,7 @@ public static void fetchColumns(Class<?> implementingClass, Configuration conf,
     List<String> serialized = new ArrayList<>();
     if (confValue != null) {
       // Split and include any trailing empty strings to allow empty column families
-      for (String val : confValue.split(",", -1)) {
-        serialized.add(val);
-      }
+      Collections.addAll(serialized, confValue.split(",", -1));
```
keith-turner, line 364: This is a handy method, I didn't know it existed.

PR #1614, opened 2020-05-19
Create max tablets property in new bulk import

milleruntime:
* Creates the master.bulk.max.tablets property to allow restricting the
number of tablets in a single bulk import

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
@@ -432,6 +432,7 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
       String fileName = entry.getKey();
       List<Destination> destinations = entry.getValue();
       Set<KeyExtent> extents = mapDestinationsToExtents(tableId, extentCache, destinations);
+      log.debug("The file {} mapped to {} tablets.", fileName, extents.size());
       checkTabletCount(maxTablets, extents.size(), fileName);
 
       long estSize = (long) (fileLens.get(fileName) / (double) extents.size());
@@ -565,7 +566,7 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
             pathLocations.put(ke, new Bulk.FileInfo(filePath, estSizes.getOrDefault(ke, 0L)));
           }
           long t2 = System.currentTimeMillis();
-          log.trace("Mapped {} to {} tablets in {}ms", filePath, pathLocations.size(), t2 - t1);
+          log.debug("Mapped {} to {} tablets in {}ms", filePath, pathLocations.size(), t2 - t1);
           return pathLocations;
         } catch (Exception e) {
           throw new CompletionException(e);
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
@@ -134,6 +134,14 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
 
     SortedMap<KeyExtent,Bulk.Files> mappings;
     TableOperationsImpl tableOps = new TableOperationsImpl(context);
+
+    int maxTablets = 0;
+    for (var prop : tableOps.getProperties(tableName)) {
+      if (prop.getKey().equals(Property.TABLE_BULK_MAX_TABLETS.getKey())) {
+        maxTablets = Integer.parseInt(prop.getValue());
+        break;
+      }
+    }
     Retry retry = Retry.builder().infiniteRetries().retryAfter(100, MILLISECONDS)
         .incrementBy(100, MILLISECONDS).maxWait(2, MINUTES).backOffFactor(1.5)
         .logInterval(3, TimeUnit.MINUTES).createRetry();
@@ -142,18 +150,13 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
     boolean shouldRetry = true;
     while (shouldRetry) {
       if (plan == null) {
-        mappings = computeMappingFromFiles(fs, tableId, srcPath);
+        mappings = computeMappingFromFiles(fs, tableId, srcPath, maxTablets);
       } else {
-        mappings = computeMappingFromPlan(fs, tableId, srcPath);
+        mappings = computeMappingFromPlan(fs, tableId, srcPath, maxTablets);
       }
 
       if (mappings.isEmpty())
         throw new IllegalArgumentException("Attempted to import zero files from " + srcPath);
-
-      long tabletMaxSize = conf.getCount(Property.MASTER_BULK_MAX_TABLETS);
-      if (tabletMaxSize > 0 && mappings.keySet().size() > tabletMaxSize)
-        throw new IllegalArgumentException("Attempted to import to " + mappings.keySet().size()
-            + " tablets. Max tablets allowed set to " + tabletMaxSize);
 
       BulkSerialize.writeLoadMapping(mappings, srcPath.toString(), fs::create);
 
@@ -391,7 +394,7 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
   }
 
   private SortedMap<KeyExtent,Files> computeMappingFromPlan(FileSystem fs, TableId tableId,
-      Path srcPath)
+      Path srcPath, int maxTablets)
       throws IOException, AccumuloException, AccumuloSecurityException, TableNotFoundException {
 
     Map<String,List<Destination>> fileDestinations =
@@ -428,7 +431,8 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
     for (Entry<String,List<Destination>> entry : fileDestinations.entrySet()) {
       String fileName = entry.getKey();
       List<Destination> destinations = entry.getValue();
-      Set<KeyExtent> extents = mapDesitnationsToExtents(tableId, extentCache, destinations);
+      Set<KeyExtent> extents = mapDestinationsToExtents(tableId, extentCache, destinations);
+      checkTabletCount(maxTablets, extents.size(), fileName);
 
       long estSize = (long) (fileLens.get(fileName) / (double) extents.size());
 
@@ -445,7 +449,7 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
     return row == null ? null : new Text(row);
   }
 
-  private Set<KeyExtent> mapDesitnationsToExtents(TableId tableId, KeyExtentCache kec,
+  private Set<KeyExtent> mapDestinationsToExtents(TableId tableId, KeyExtentCache kec,
       List<Destination> destinations)
       throws AccumuloException, AccumuloSecurityException, TableNotFoundException {
     Set<KeyExtent> extents = new HashSet<>();
@@ -476,7 +480,7 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
   }
 
   private SortedMap<KeyExtent,Bulk.Files> computeMappingFromFiles(FileSystem fs, TableId tableId,
-      Path dirPath) throws IOException {
+      Path dirPath, int maxTablets) throws IOException {
 
     Executor executor;
     ExecutorService service = null;
@@ -492,7 +496,7 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
     }
 
     try {
-      return computeFileToTabletMappings(fs, tableId, dirPath, executor, context);
+      return computeFileToTabletMappings(fs, tableId, dirPath, executor, context, maxTablets);
     } finally {
       if (service != null) {
         service.shutdown();
@@ -529,8 +533,8 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
     return fileList;
   }
 
-  public static SortedMap<KeyExtent,Bulk.Files> computeFileToTabletMappings(FileSystem fs,
-      TableId tableId, Path dirPath, Executor executor, ClientContext context) throws IOException {
+  public SortedMap<KeyExtent,Bulk.Files> computeFileToTabletMappings(FileSystem fs, TableId tableId,
+      Path dirPath, Executor executor, ClientContext context, int maxTablets) throws IOException {
 
     KeyExtentCache extentCache = new ConcurrentKeyExtentCache(tableId, context);
 
@@ -546,21 +550,22 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
     CryptoService cs = CryptoServiceFactory.newDefaultInstance();
 
     for (FileStatus fileStatus : files) {
+      Path filePath = fileStatus.getPath();
       CompletableFuture<Map<KeyExtent,Bulk.FileInfo>> future = CompletableFuture.supplyAsync(() -> {
         try {
           long t1 = System.currentTimeMillis();
-          List<KeyExtent> extents = findOverlappingTablets(context, extentCache,
-              fileStatus.getPath(), fs, fileLensCache, cs);
-          Map<KeyExtent,Long> estSizes = estimateSizes(context.getConfiguration(),
-              fileStatus.getPath(), fileStatus.getLen(), extents, fs, fileLensCache, cs);
+          List<KeyExtent> extents =
+              findOverlappingTablets(context, extentCache, filePath, fs, fileLensCache, cs);
+          // make sure file isn't going to too many tablets
+          checkTabletCount(maxTablets, extents.size(), filePath.toString());
+          Map<KeyExtent,Long> estSizes = estimateSizes(context.getConfiguration(), filePath,
+              fileStatus.getLen(), extents, fs, fileLensCache, cs);
           Map<KeyExtent,Bulk.FileInfo> pathLocations = new HashMap<>();
           for (KeyExtent ke : extents) {
-            pathLocations.put(ke,
-                new Bulk.FileInfo(fileStatus.getPath(), estSizes.getOrDefault(ke, 0L)));
+            pathLocations.put(ke, new Bulk.FileInfo(filePath, estSizes.getOrDefault(ke, 0L)));
           }
           long t2 = System.currentTimeMillis();
-          log.trace("Mapped {} to {} tablets in {}ms", fileStatus.getPath(), pathLocations.size(),
-              t2 - t1);
+          log.trace("Mapped {} to {} tablets in {}ms", filePath, pathLocations.size(), t2 - t1);
           return pathLocations;
         } catch (Exception e) {
           throw new CompletionException(e);
@@ -582,7 +587,12 @@ public class BulkImport implements ImportDestinationArguments, ImportMappingOpti
         Thread.currentThread().interrupt();
         throw new RuntimeException(e);
       } catch (ExecutionException e) {
-        throw new RuntimeException(e);
+        // clean up exception for user
+        Throwable t = e.getCause();
+        if (t instanceof IllegalArgumentException) {
+          throw (IllegalArgumentException) t;
+        } else
+          throw new RuntimeException(t);
       }
     }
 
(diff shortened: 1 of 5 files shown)


milleruntime:
This PR addresses #1559 for 2.1.  Related to #1560

EdColeman:
Would there be any use for this being able to be overridden on a table basis?  My thought would be that "normally" there would be a system limit, but because of some external factor there was a need to allow this to be on a case by case basis.  It could go either way - say, I know this is not optimal, but for this table I want to allow more files, or the other way would be that the system allows X, but for this table I want it to be less because at X bad things happen and I'd rather fail.

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
```diff
@@ -183,6 +184,23 @@ public void testSingleTabletSingleFileOffline() throws Exception {
...
+  public void testMaxTablets() throws Exception {
+    String maxTablets = "0";
+    try (AccumuloClient client = Accumulo.newClient().from(getClientProps()).build()) {
+      maxTablets = client.instanceOperations().getSystemConfiguration()
+          .get(Property.MASTER_BULK_MAX_TABLETS.getKey());
+      client.instanceOperations().setProperty(Property.MASTER_BULK_MAX_TABLETS.getKey(), "1");
+      testBulkFile(false, false);
+      fail("Expected IllegalArgumentException for " + Property.MASTER_BULK_MAX_TABLETS);
+    } catch (IllegalArgumentException e) {} finally {
+      try (AccumuloClient client = Accumulo.newClient().from(getClientProps()).build()) {
```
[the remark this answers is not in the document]
milleruntime: I was thinking of the situation where an exception other than IllegalArgumentException is thrown, the property won't be changed back.  Typically this doesn't matter, but I was under the impression we are still striving towards getting the ITs to work as a standalone cluster (even if its still currently fails).

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
```diff
@@ -149,6 +150,11 @@ public void load()
       if (mappings.isEmpty())
         throw new IllegalArgumentException("Attempted to import zero files from " + srcPath);
 
+      long tabletMaxSize = conf.getCount(Property.MASTER_BULK_MAX_TABLETS);
+      if (tabletMaxSize > 0 && mappings.keySet().size() > tabletMaxSize)
```
keith-turner, line 154: This check is different than the one in #1560, which ensured a single file did not go to too many tablets.  This is checking that all files do not go to too many tablets.

milleruntime:
> Would there be any use for this being able to be overridden on a table basis? 

That is a good question.  I hadn't thought about that level of granularity but it would make sense as a next step.  And the bulk import API does operate on a per table basis so that makes sense too.  I will look into it as I am not sure how that would work with the implementation in Master.

keith-turner:
I like the suggestion for a table prop, that aligns more closely with the user experience rather than how its implemented.

The check on the client side is nice for fail fast, but we may want a check on the server side to ensure correctness.  However the server side check is really hard to test.

Are there any plans to make the old bulk import code respect this property?

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
milleruntime: Yes that is correct.  I am not sure which is better but based on the original description in #1559, I took it as we want to prevent files from mapping to too many tablets.

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
```diff
@@ -183,6 +184,23 @@ public void testSingleTabletSingleFileOffline() throws Exception {
     }
   }
 
+  @Test
+  public void testMaxTablets() throws Exception {
+    String maxTablets = "0";
+    try (AccumuloClient client = Accumulo.newClient().from(getClientProps()).build()) {
+      maxTablets = client.instanceOperations().getSystemConfiguration()
+          .get(Property.MASTER_BULK_MAX_TABLETS.getKey());
+      client.instanceOperations().setProperty(Property.MASTER_BULK_MAX_TABLETS.getKey(), "1");
```
keith-turner, line 193: I am not completely sure, but these two lines may not work as expected sometimes.  What I think may be able to happen is that the property is set on tserver A by the first line and then the second line reads props from tserver B.  If server B does not clear zoocache, then it could return a stale copy w/o the prop set on the first line.  I am not sure if the client side code to get props forces a prop cache clear.

keith-turner: If it were a table prop, then table could be created with the prop and would not need to clear in finally.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
milleruntime: I guess the danger with this check is stopping ingest due to false positives?  Like if I have 100 tablets and 10 files going to all tablets could be normal. I would rather stop 1 file going to all 100 then 10 going to all 100.

milleruntime:
> Are there any plans to make the old bulk import code respect this property?

I made some comments on #1560 about this as I think it would be worth it but should be discussed since it would incur adding a new feature to a deprecated API.  The implementation in 2.1 was a priority first though.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
keith-turner: I just looked at #1559 and saw the following sentence.  I would be in favor of preventing a single file from going to too many tablets.  If someone has a large number of tablets and each file is going to a few tablets, they are probably ok with that like you said.

> It would be nice if we could establish a threshold in the bulk import process to abort when encountering a rfile that maps to more than a specified number of extents.

milleruntime:
@EdColeman @keith-turner I made some improvements based on your feedback.  I had to move the check inside the computeMapping methods so it would fail fast on a single file.  This also lead me to clean up the exception as it comes out of a future so the user will just get a single IllegalArgumentException.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
```diff
@@ -133,6 +134,14 @@ public void load()
...
     SortedMap<KeyExtent,Bulk.Files> mappings;
     TableOperationsImpl tableOps = new TableOperationsImpl(context);
+
+    int maxTablets = 0;
+    for (var prop : tableOps.getProperties(tableName)) {
+      if (prop.getKey().equals(Property.TABLE_BULK_MAX_TABLETS.getKey())) {
+        maxTablets = Integer.parseInt(prop.getValue());
+        break;
+      }
+    }
```
milleruntime, line 144: I couldn't find a cleaner way to get a single table property from the client side...

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
```diff
@@ -576,7 +587,12 @@ private Text toText(byte[] row) {
         Thread.currentThread().interrupt();
         throw new RuntimeException(e);
       } catch (ExecutionException e) {
-        throw new RuntimeException(e);
+        // clean up exception for user
+        Throwable t = e.getCause();
+        if (t instanceof IllegalArgumentException) {
+          throw (IllegalArgumentException) t;
+        } else
+          throw new RuntimeException(t);
```
milleruntime, line 595: I thought it was cleaner for the user to just drop the ExecutionException coming out of the future since that is just implementation specific.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
```diff
@@ -576,7 +587,12 @@ private Text toText(byte[] row) {
...
       }
```
keith-turner, line 596: I would wrap `e` so that stack traces are not lost and anyone getting the exceptions can trace the full code path from their code to the background thread.

```suggestion
        Throwable t = e.getCause();
        if (t instanceof IllegalArgumentException) {
          throw new IllegalArgumentException(e);
        } else
          throw new RuntimeException(e);
      }
```

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
keith-turner: This could lead to a loss of information for someone debugging.  Like they see a stack trace from a background thread in Accumulo, but they don't know anything about the calling code in the foreground thread.

keith-turner: I am not sure if the ExecutionException provides valuable information, but definitely want to wrap the IllegalArgumentException.  I would err on the side of caution with omiting the excution exception in that even if does not currently provide anything useful, it may in future versions of Java.

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
```diff
@@ -186,18 +186,26 @@ public void testSingleTabletSingleFileOffline() throws Exception {
...
+    } catch (IllegalArgumentException e) {
+      // expected
+    }
+    // test max tablets hit using load plan
+    try {
+      testBulkFile(false, true);
+      fail("Expected IllegalArgumentException for " + Property.TABLE_BULK_MAX_TABLETS);
+    } catch (IllegalArgumentException e) {
+      // expected
     }
```
keith-turner: I recently learned about this new feature in Junit.  It also returns the expected exception if you want to check the message.

```suggestion
   Assert.assertThrows(IllegalArgumentException.class, () -> testBulkFile(false, true));
```

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
milleruntime: Good point.  Better to error on the side of too much information.

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
milleruntime: Cool thanks.  I am not sure what happened but Github isn't letting me apply your suggestion.  I will just change it manually.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
milleruntime: I am not sure this is better than just the Runtime.  I think this will wrap an IllegalArgument around the whole stack, which already has an IllegalArgument at the bottom.  I will play around.

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
milleruntime: I settled with keeping the exception handling the way it is now.  I was able to clean up the IT using assertThrows and just checked the causes manually.  I think this is as clean as it is going to get until we move to Junit5.  I did find this library which I thought was interesting: https://assertj.github.io/doc/

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
milleruntime: See my other comment.  You OK with me keeping the exception handling as-is?

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
```diff
@@ -183,6 +186,29 @@ public void testSingleTabletSingleFileOffline() throws Exception {
...
   }
 
+  @Test
+  public void testMaxTablets() throws Exception {
+    try (AccumuloClient client = Accumulo.newClient().from(getClientProps()).build()) {
+      tableName = "testMaxTablets_table1";
+      NewTableConfiguration newTableConf = new NewTableConfiguration();
+      // set logical time type so we can set time on bulk import
+      Map<String,String> props = new HashMap<>();
+      props.put(Property.TABLE_BULK_MAX_TABLETS.getKey(), "1");
```
keith-turner, line 196:
```suggestion
     var props = Map.of(Property.TABLE_BULK_MAX_TABLETS.getKey(), "1");
```

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
```diff
@@ -183,6 +186,29 @@ public void testSingleTabletSingleFileOffline() throws Exception {
...
+      newTableConf.setProperties(props);
+      client.tableOperations().create(tableName, newTableConf);
+
+      // test max tablets hit while inspecting bulk files
+      var thrown = assertThrows(RuntimeException.class, () -> testBulkFile(false, false));
+      var c = thrown.getCause();
+      assertTrue("Wrong exception: " + c, c instanceof ExecutionException);
+      assertTrue("Wrong exception: " + c.getCause(),
+          c.getCause() instanceof IllegalArgumentException);
+
```
keith-turner, line 206: Not sure if this is possible, but it would be really nice to confirm that exception message contains the offending file name.  Whenever someone runs into this error message, knowing which file caused the problem will be extremely helpful to them.

If the test does not do this, I would also recommend creating multiple files.  One that exceeds the limit and few that do not.  Want to ensure in this case the troublesome file is listed in the message.

core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java
keith-turner: Sorry for the delay @milleruntime.  This fell off my radar.

test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java
milleruntime: I have been looking at BulkNewIT to see how we could test this scenario and I think there is an issue with the test data.  This line doesn't make sense to me and seems wrong: https://github.com/milleruntime/accumulo/blob/f71a2d1ba175dbb74a8f9e26151b778c270eaa2c/test/src/main/java/org/apache/accumulo/test/functional/BulkNewIT.java#L315

This also made me wonder about exclusivity when creating the load plan.  We don't mention whether the start and end rows are inclusive/exclusive on the load API [here](https://github.com/apache/accumulo/blob/3fd5cad92f9b63ac19e4466f3f2d5237b905262c/core/src/main/java/org/apache/accumulo/core/data/LoadPlan.java#L187).  I was wondering this myself and eventually found it mentioned in the ```RangeType``` enum but think it should also be mentioned on the ```loadFileTo``` methods, with the ```startRow``` and ```endRow``` parameters.

keith-turner: I looked at BulkNewIT line 315,  I think what is going is that the file with a hash of h3 overlaps two tablets.  It overlaps a tablet with an end row of 1333 and another tablet with an end row of 1666.  So that file is added as expected for those two tablets.

keith-turner: > but think it should also be mentioned on the loadFileTo methods, with the startRow and endRow parameters.

For the loadFileTo method, its javadoc could say that how the startRow and endRow are interpreted depends on the RangeType (with a javadoc link to RangeType).

milleruntime:
@keith-turner I added new data for the max tablet test but now I am getting different tablet counts when using the LoadPlan.  I can't figure out why.  The changes i just pushed up in 3ff589a don't throw an IllegalArgumentException when using the LoadPlan.  Here are the counts I get:
```
2020-06-02T11:35:34,990 [bulk.BulkImport] DEBUG: The file f3.rf mapped to 1 tablets.
2020-06-02T11:35:34,991 [bulk.BulkImport] DEBUG: The file f4.rf mapped to 2 tablets.
2020-06-02T11:35:34,991 [bulk.BulkImport] DEBUG: The file f2.rf mapped to 1 tablets.
2020-06-02T11:35:34,991 [bulk.BulkImport] DEBUG: The file f1.rf mapped to 1 tablets.
```

keith-turner:
> @keith-turner I added new data for the max tablet test but now I am getting different tablet counts when using the LoadPlan. I can't figure out why. The changes i just pushed up in 3ff589a don't throw an IllegalArgumentException when using the LoadPlan

I took a look at the code quickly. I think what is happening is that for the RangeType.TABLE it does nothing for this on the client side and maps the range to tablets on the server side.

keith-turner:
@milleruntime this is the code that I think is just serving as a pass through on the client side

https://github.com/apache/accumulo/blob/3ff589ac91ac668b433e78fdb3f6a735d5da7091/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java#L461

milleruntime:
> @milleruntime this is the code that I think is just serving as a pass through on the client side
> 
> https://github.com/apache/accumulo/blob/3ff589ac91ac668b433e78fdb3f6a735d5da7091/core/src/main/java/org/apache/accumulo/core/clientImpl/bulk/BulkImport.java#L461

OK so it looks like a file can only ever have 1 tablet when using ```RangeType.TABLE```.  Is this as intended?  Is this how Accumulo creates files when splitting?

keith-turner:
> OK so it looks like a file can only ever have 1 tablet when using RangeType.TABLE. Is this as intended? Is this how Accumulo creates files when splitting?

It can correspond to one or more tablets, but must map to existing table split points.

https://github.com/apache/accumulo/blob/3fd5cad92f9b63ac19e4466f3f2d5237b905262c/core/src/main/java/org/apache/accumulo/core/data/LoadPlan.java#L71

milleruntime:
> It can correspond to one or more tablets, but must map to existing table split points.

Than how come this line in my last commit only maps to 1 tablet?
```     
.loadFileTo("f2.rf", RangeType.TABLE, row(333), row(1333))
```
Since I used the same splits as the other tests, ```"0333 0666 0999 1333 1666 null"``` I thought this would map to 3 tablets 0666-0334, 0999-0667, 1333-1000...

keith-turner:
> I thought this would map to 3 tablets 0666-0334, 0999-0667, 1333-1000...

I am thinking it will on the server side.  So the client passes that single range and then it gets expanded into one or more tablets on the servers side.  

https://github.com/apache/accumulo/blob/3ff589ac91ac668b433e78fdb3f6a735d5da7091/server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/LoadFiles.java#L339

milleruntime:
> I am thinking it will on the server side. So the client passes that single range and then it gets expanded into one or more tablets on the servers side.

Ah I see.  If that is the case, I should probably add the check for max tablets on the server side as well.  I could see how 1 tablet could easily get expanded beyond the limit.

keith-turner:
> Ah I see. If that is the case, I should probably add the check for max tablets on the server side as well. I could see how 1 tablet could easily get expanded beyond the limit.

If its possible, then I think the best place to do it on the servers side would be `PrepBulkImport.checkForMerge()` because this is before any files have been loaded to any tablets.  I am not sure if its possible though, because at that point I think the data is grouped by range like `Map<Range, List<File>>`.  To easily do the check we need `Map<File, List<Range>>`. 

Another option if that does not work is when Range.TABLE is seen AND maxTablets > 0, then we actually try to resolve the range to tablets.   However this kinda defeats the purpose of Range.TABLE as it is for the case when user has the most knowledge and Accumulo can do the least work in the fastest manner.

Not sure what is best, just trying to think through the options.  Need to look at the code some more too.

keith-turner:
@milleruntime the following branch has a change I was experimenting with for adding a server side check... its the top commit

https://github.com/keith-turner/accumulo/tree/accumulo-1614

milleruntime:
> @milleruntime the following branch has a change I was experimenting with for adding a server side check... its the top commit
> 
> https://github.com/keith-turner/accumulo/tree/accumulo-1614

Thanks.  It took a bit more work than I expected but I think I got the server side (master FATE) check working.   I also added a note to the property description.

milleruntime:
@keith-turner want to take a last look at this?  I think it is ready to merge.

keith-turner:
@milleruntime I went ahead and merged this because you said you thought it was ready to merge and I worked up a follow on test in #1623.

PR #1632, opened 2020-06-18
Fix #1609 dynamically reinitialize compaction services when config ch…

### LINKED ISSUE #1609: Reinitialize compaction services when configuration changes
The changes in #1605 create compaction services based on configuration at tablet server startup.  If the configuration in zookeeper changes after startup, the compaction services never change.  It would be nice to update the compaction services whenever configuration changes.

When the configuration for an existing compaction service goes away, instances of that service should finish any running compactions and not accept any new work while waiting.

When the configuration for existing compaction service changes, it should resize existing thread pools and/or delete thread pools that are no longer needed.

When the configuration for a compaction service that does not currently exist is noticed, then it should be created.

keith-turner:
…anges

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionManager.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionManager.java
@@ -97,6 +97,10 @@ public class CompactionManager {
       return false;
     }
 
+    @Override
+    public int hashCode() {
+      return Objects.hash(planners, options);
+    }
   }
 
   private void mainLoop() {
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionManager.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionManager.java
@@ -20,33 +20,84 @@ package org.apache.accumulo.tserver.compactions;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.accumulo.core.conf.AccumuloConfiguration;
 import org.apache.accumulo.core.conf.Property;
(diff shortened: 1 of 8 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionManager.java
```diff
@@ -20,34 +20,89 @@
...
 
+import com.google.common.collect.Sets;
+
 public class CompactionManager {
 
   private static final Logger log = LoggerFactory.getLogger(CompactionManager.class);
 
   private Iterable<Compactable> compactables;
-  private Map<CompactionServiceId,CompactionService> services;
+  private volatile Map<CompactionServiceId,CompactionService> services;
```
milleruntime, line 47: Did you consider using a ConcurrentHashMap here?  I see you made the map volatile, but was wondering if it would help prevent concurrency issues to have a ConcurrentHashMap as well.

keith-turner: I am using Map.copyOf() for the maps, so its a volatile pointer to an immutable map.   Whenever its updated the entire map is replaced with another immutable map.  This pattern is good for maps that are concurrently read a lot and updated infrequently.  The internal pointers in concurrent hash map are volatile, making all operations on the map slower.  The internal pointers in an immutable map are probably not volatile, just our pointer to the entire map is.

milleruntime: Thanks for the explanation.  I think what you have is fine.  I was just wondering what is the best way to guard against future changes which could introduce concurrency issues.  For example, you used a coding pattern which may not be obvious to a new developer, versus using a concrete type that is less likely to change.  Your code is definitely more efficient but it was not initially clear to me why.  Perhaps just some comments with your explanation or creating a method to specifically update the map would help.

keith-turner: I will add a comment.

Issue #1620, opened 2020-06-03
Possible loss of parallelism in VolumeImpl


milleruntime:
The implementation of VolumeManager may be negating some of the work done in other areas of code where we are using threadpools and futures to improve calls to the Namenode.  I think this is only noticeable during the Bulk import processes (old and new) when we are calling [VolumeManager.rename()](https://github.com/apache/accumulo/blob/dd4a488a9077cdd5e90cbff89fbcf7e9147381a7/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java#L282) when moving the files.  

@ivakegg and I recently saw a lot of threads waiting on a single rename during a Bulk import in 1.9.  It is possible the way that we are reusing the hadoop FileSystem in the VolumeManager code in 1.9 is causing this.  It is also possible that this was fixed with the refactor done in #1553

keith-turner:
There is a possibly related change for import table #1569

I think it would be nice to have a VolumeManager.bulkRename() method for renaming lots of files.  Then we can have a single implementation of this in the code to fix and maintain.

milleruntime:
It could also be that Hadoop is returning the same object either way since it caches FileSystem objects on the static FileSystem get(uri, conf) call.  Sorry can't find a link to the source code in Hadoop :roll_eyes: 
The code I am looking at will always return this:
```return CACHE.get(uri, conf);```

milleruntime:
> There is a possibly related change for import table #1569
> 
> I think it would be nice to have a VolumeManager.bulkRename() method for renaming lots of files. Then we can have a single implementation of this in the code to fix and maintain.

I started looking into this and noticed some subtle differences in debugging and exception handling betwen tableImport and bulk import use of the thread pool.  I am not sure if the behavior is specific to the actual FATE operation or it is just how the code evolved independently.  For example, BulkImportMove has this [check](https://github.com/apache/accumulo/blob/15a95feebc74bd10e5ba73d4c2ecfc726435bc55/server/master/src/main/java/org/apache/accumulo/master/tableOps/bulkVer2/BulkImportMove.java#L131) for existence while the [MoveExportedFiles](https://github.com/apache/accumulo/blob/27acec198401942fbd5d90edfea71ac859533470/server/master/src/main/java/org/apache/accumulo/master/tableOps/tableImport/MoveExportedFiles.java#L117) does not.

keith-turner:
[...]

PR #1635, opened 2020-06-19
Create bulkRename method in VolumeManager

milleruntime:
* Refactor fate code in BulkImportMove (bulkVer2) and MoveExportedFiles to call the
new method for handling renames of lots of files
* Work related to #1620 
* Made bulkRename method return the list of completed Futures so FATE operations can handle errors accordingly

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
@@ -150,8 +150,11 @@ public interface VolumeManager extends AutoCloseable {
   // volumes
   boolean rename(Path path, Path newPath) throws IOException;
 
-  // rename lots of files at once in a thread pool
-  // returns once all the threads have completed
+  /**
+   * Rename lots of files at once in a thread pool and return once all the threads have completed.
+   * This operation should be idempotent to allow calling multiple times in the case of a partial
+   * completion.
+   */
   void bulkRename(Map<Path,Path> oldToNewPathMap, int poolSize, String poolName,
       String transactionId) throws IOException;
 
--- a/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
@@ -22,9 +22,13 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.UnknownHostException;
 import java.util.Collection;
+import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Future;
 
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
+import org.apache.accumulo.core.util.SimpleThreadPool;
 import org.apache.accumulo.core.volume.Volume;
 import org.apache.accumulo.core.volume.VolumeConfiguration;
 import org.apache.accumulo.server.ServerConstants;
@@ -148,6 +152,11 @@ public interface VolumeManager extends AutoCloseable {
   // forward to the appropriate FileSystem object, throws an exception if the paths are in different
   // volumes
   boolean rename(Path path, Path newPath) throws IOException;
+
+  // rename lots of files at once in a thread pool
+  // returns the results once all the threads have completed
+  List<Future<Boolean>> bulkRename(Map<Path,Path> oldToNewPathMap, SimpleThreadPool workerPool,
+      String transactionId) throws Exception;
 
   // forward to the appropriate FileSystem object
   boolean moveToTrash(Path sourcePath) throws IOException;
--- a/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
@@ -28,7 +28,6 @@ import java.util.Set;
 import java.util.concurrent.Future;
 
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
-import org.apache.accumulo.core.util.SimpleThreadPool;
 import org.apache.accumulo.core.volume.Volume;
 import org.apache.accumulo.core.volume.VolumeConfiguration;
 import org.apache.accumulo.server.ServerConstants;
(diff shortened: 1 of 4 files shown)


server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
```diff
@@ -289,6 +294,38 @@ public boolean rename(Path path, Path newPath) throws IOException {
...
+        if (!exists(newPath) || exists(oldPath)) {
+          throw e;
+        }
+        log.debug(
+            "Ignoring rename exception for {} because destination already exists. orig: {} new: {}",
+            transactionId, oldPath, newPath, e);
+        success = true;
+      }
+      if (!success) {
+        log.error("Rename operation {} returned false. orig: {} new: {}", transactionId, oldPath,
```
keith-turner, line 317: should this throw an exception?

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
```diff
@@ -289,6 +294,38 @@ public boolean rename(Path path, Path newPath) throws IOException {
     return source.rename(path, newPath);
   }
 
+  @Override
+  public List<Future<Boolean>> bulkRename(Map<Path,Path> oldToNewPathMap,
+      SimpleThreadPool workerPool, String transactionId) throws InterruptedException {
+    List<Future<Boolean>> results = new ArrayList<>();
+    oldToNewPathMap.forEach((oldPath, newPath) -> results.add(workerPool.submit(() -> {
```
keith-turner, line 301: A lambda within a lambda. I was looking for a way to add a third inner lambda, but did not see an opportunity.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
```diff
@@ -289,6 +294,38 @@ public boolean rename(Path path, Path newPath) throws IOException {
...
+      }
+      if (!success) {
+        log.error("Rename operation {} returned false. orig: {} new: {}", transactionId, oldPath,
+            newPath);
+      } else if (log.isTraceEnabled()) {
+        log.trace("{} moved {} to {}", transactionId, oldPath, newPath);
+      }
+      return success;
+    })));
+    workerPool.shutdown();
```
keith-turner, line 324: If this method is going to shutdown the pool, I think it would be better if it created the pool also.  This method could accept a number of threads and operation name instead of a thread pool and use those to create the pool.

milleruntime: That makes more sense to isolate the pool.  I will make that change.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
milleruntime: My plan was to preserve current behavior of both methods.  Making this throw an exception would change the behavior where one file failing to rename fails the whole operation.

keith-turner: The code that handles the futures deals with execution exceptions and false returns.  Could possibly make this simpler and not return anything `Future<Void>` and only deal with execution exceptions.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
```diff
@@ -289,6 +294,39 @@ public boolean rename(Path path, Path newPath) throws IOException {
...
+    List<Future<Boolean>> results = new ArrayList<>();
+    SimpleThreadPool workerPool = new SimpleThreadPool(poolSize, poolName);
+    oldToNewPathMap.forEach((oldPath, newPath) -> results.add(workerPool.submit(() -> {
+      boolean success;
+      try {
+        success = rename(oldPath, newPath);
+      } catch (IOException e) {
+        // The rename could have failed because this is the second time its running (failures
+        // could cause this to run multiple times).
+        if (!exists(newPath) || exists(oldPath)) {
```
keith-turner, line 309: I wonder if this check should also be done in the case where rename returns false.

milleruntime: In that case would we throw a new IOException?

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
milleruntime: This does seem to simplify things some.  You still suggesting we throw an error here when ```!success```?

keith-turner: I took a quick look at the code calling bulk rename and I think it always fails if any of the futures fail.   I was wondering if it would make sense to move the future handling inside bulk rename and just have bulkRename throw an IOE if any future fails.  As long as this can still result in good error messages to users, I think may simplify the code calling bulk rename alot.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
milleruntime: I decided since we are already logging an error per file when rename returns false, we should not also throw a new error, failing the transation.  It seems like this is just to rethrow any errors we may care about.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManagerImpl.java
milleruntime: This did seem to clean up the handling in the FATE methods.  Checkout my latest, I think this is ready to go now.

server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java
```diff
@@ -149,6 +150,11 @@ FSDataOutputStream createSyncable(Path logPath, int buffersize, short replicatio
   // volumes
   boolean rename(Path path, Path newPath) throws IOException;
 
+  // rename lots of files at once in a thread pool
+  // returns once all the threads have completed
```
keith-turner, line 154: could add something to the docs about this operation being idempotent, safe to call again after previous partial run of bulkRename

PR #1646, opened 2020-06-30
fixes #1644: minor compaction retry

### LINKED ISSUE #1644: minor compaction never retries
I came across a situation where a subset of minor compactions failed and were never retried (version 1.9.3).  This then resulted in a major compaction hanging in the TabletMemory.waitForMinC method forever.  Eventually the system backed with files on those tablets.
The reason the minor compaction thread died was because our system was being modified with a new context and the VFSClassLoader was changing classpaths.  This resulted in a NoClassDefFoundError while trying to load the iterators for that table.  The purpose of this ticket is not to fix anything to do with the VFSClassLoader mechanism, but rather to ensure that if a minor compaction fails for any reason, it will be subsequently retried perhaps with some sort of exponential backoff.

ivakegg:
* Updated the minor compaction to gracefully handle temporary VFS class loader issues while minor compacting

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
@@ -138,7 +138,7 @@ public class MinorCompactor extends Compactor {
           ProblemReports.getInstance(tabletServer).report(new ProblemReport(
               getExtent().getTableId(), ProblemType.FILE_WRITE, outputFileName, e));
           reportedProblem = true;
-        } catch (RuntimeException e) {
+        } catch (RuntimeException | NoClassDefFoundError e) {
           // if this is coming from a user iterator, it is possible that the user could change the
           // iterator config and that the
           // minor compaction would succeed


server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
```diff
@@ -148,6 +149,13 @@ public CompactionStats call() {
           reportedProblem = true;
         } catch (CompactionCanceledException e) {
           throw new IllegalStateException(e);
+        } catch (Throwable t) {
+          // This is an unexpected situation which is critical. If we cannot minor compact
+          // then this tablet will eventually become unusable if we are still ingesting data.
+          log.error("MinC failed ({}).  Aborting.", t.getMessage(), t);
+          // TODO: Should we simply unload this tablet instead?
+          Halt.halt("Unexpected MinC failure: " + t.getMessage(), 2);
+          throw t;
```
milleruntime, line 158: We typically try to either log an error or throw, not both.  Instead of halting in MinorCompactor, we may want halt [here in Tablet](https://github.com/apache/accumulo/blob/15a95feebc74bd10e5ba73d4c2ecfc726435bc55/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java#L829) where we catch Exception or Error.  That code currently does not catch Throwable.

ivakegg: Good call.  I will move the halt logic.  The question still remains whether halting is the appropriate action here.  I could try unloading the tablet instead.....

ivakegg: Looking closer that might not work because of the IllegalStateException thrown when the minor compaction is cancelled....

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
```diff
@@ -148,6 +149,13 @@ public CompactionStats call() {
           reportedProblem = true;
         } catch (CompactionCanceledException e) {
           throw new IllegalStateException(e);
+        } catch (Throwable t) {
```
keith-turner, line 152:
```suggestion
        } catch (Error t) {
```

Catching throwable will mask checked exceptions in future code changes.  For example if the checked exceptions on a method called here are changed, and those checked exceptions should be handled.. then catching Throwable would hide the need to handle those.

ivakegg: Throwable was intentional, but reducing it to an Error would be fine I guess.  My take was if a "checked" exception is thrown down the road that is not actually caught, then the system will end up in the same defunct state.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
milleruntime: Following the stack further up, we are already catching Throwable [here in MinorCompactionTask](https://github.com/apache/accumulo/blob/94844b3ef88bc9e02b5cb5c8eeff81411ac8221d/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactionTask.java#L111).  So if we decide to do anything to the tablet, I think that would be the best place since that method is already operating on the tablet object.  The finally in that try will set the ```minorCompactionState``` so we one option would be to add another state to the [enum in Tablet](https://github.com/apache/accumulo/blob/15a95feebc74bd10e5ba73d4c2ecfc726435bc55/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java#L217) to indicate failed state.  I don't know what all this state is used for but it may be less complicated than unloading the Tablet and less problematic than halting the tserver.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
keith-turner: I was just thinking if this code catches Error and and a checked exception is added elsewhere then this code will not compile the developer will have to decided what to do.

ivakegg: Now I am thinking that a Halt just does not feel right, and I fear that given the appropriate circumstances could result in a cascade of tserver deaths.  I am reworking this to retry failed minor compactions in the thread that triggers minor compactions.

ivakegg: @keith-turner @milleruntime Please let me know what you think about this approach: compactions that do not successfully completed get marked as failed, and then the minor compaction initiator thread will pick those up and rerun them.  Thoughts?

keith-turner: > Now I am thinking that a Halt just does not feel right, and I fear that given the appropriate circumstances could result in a cascade of tserver deaths. 

That could happen.  Part of the problem is that some errors are benign and others are likely and indication of a catastrophe.  Maybe configuration is the best option could have three configuration items.

 * Configurable list of errors to retry on.
 * Configurable list of errors to halt on.
 * Action to take for errors that fall in neither list : hang, retry, halt. 

This would allow when a new error is encountered like in #1644 that the error class could just be added to config for retry.

>  It could be that all of these changes go away and I simply catch the NoClassDefFoundError and call it a day.

Maybe a follow on issue would be best.  I feel like this is a wider issue, because any thread could encounter an error and having a single mechanism for handling unexpected errors in server processes seems useful.

I am also wondering if emitting metrics for errors would be useful.  If the cardinality is deemed low enough, could emit metrics for each error class name.  Then in the metrics system could see when a tserver has an out of memory error.  This would be a follow on issue if it seems like something that might be useful.

milleruntime: > Maybe a follow on issue would be best. I feel like this is a wider issue, because any thread could encounter an error and having a single mechanism for handling unexpected errors in server processes seems useful.

I agree.  I think since its clear that we need to retry flushes on the NoClassDefFoundError that should be fixed here, with anything further being follow on work.

ivakegg: Thank you for your observations.  I will take out all of the retry stuff and simply handle the condition that this ticket warrants.  A more general and generic way of handling critical thread failures is needed if we want.  This one was kind of weird given the TabletMemory holds onto the memory being compacted at the same time a new memory block is used ongoing mutations so there are multiple states to handle at the same time.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/MinorCompactor.java
```diff
@@ -138,7 +138,7 @@ public CompactionStats call() {
           ProblemReports.getInstance(tabletServer).report(new ProblemReport(
               getExtent().getTableId(), ProblemType.FILE_WRITE, outputFileName, e));
           reportedProblem = true;
-        } catch (RuntimeException e) {
+        } catch (RuntimeException | NoClassDefFoundError e) {
```
keith-turner, line 141: This has nothing to do with your change.  I was puzzled what the diff between this catch block and the prev one was looking at the code.  AFAICT the only diff is that one passes the exception to the log.warn and the other does not.

ivakegg: That is curious and perhaps not the cleanest.  In the case of a NoClassDefFoundError I think I would want to see the exception in case the issue is more dire than a temporary VFS issue.

PR #1651, opened 2020-07-07
#1473 Move TabletsMetadata builder to Ample

cradal:


core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
```diff
@@ -67,7 +67,7 @@
  */
 public class TabletsMetadata implements Iterable<TabletMetadata>, AutoCloseable {
 
-  private static class Builder implements TableRangeOptions, TableOptions, RangeOptions, Options {
+  public static class Builder implements TableRangeOptions, TableOptions, RangeOptions, Options {
```
ivakegg, line 70: How about package private?

cradal:
This code change makes the Builder class public instead of private. This was done to give readTablets() access to instantiate a Builder object, allowing use TabletsMetadata Builder methods and implementations.

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
cradal: @ivakegg I think this is the way to go. One tumbling block is crafting a unit test that is outside the ...metadata.schema package. A test created outside that package won't have access to the Builder methods. For the moment, I will push a version with a public Builder class that is tested using a mock table.

cradal:
@keith-turner @ivakegg This version of the solution changes the modifier of TabletsMetadata Builder from private to public. This allows readTablets() in AmpleImpl to create a Builder object. I created a test using a mock table, which runs successfully. I think that package-private is a better choice for Builder, just need to figure out a way to test things outside of TabletsMetadata's package(...schema.metadata).

milleruntime:
@cradal I noticed you opened this PR in addition to another one you still have open: https://github.com/apache/accumulo/pull/1563.   If this PR supersedes #1563 then please close that one.

cradal:
@keith-turner This is ready to be reviewed. I went with making Builder a protected class vice package-private. I think this provides a good balance between encapsulating Builder and making it available to the test classes.

cradal:
@ctubbsii Just a reminder that this PR is still available for review.

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
```diff
@@ -79,6 +79,7 @@
     private boolean checkConsistency = false;
     private boolean saveKeyValues;
     private TableId tableId;
+    public AccumuloClient _client;
```
ctubbsii, line 82: public non-final variables should be avoided.

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
```diff
@@ -93,6 +94,20 @@ public TabletsMetadata build(AccumuloClient client) {
       }
     }
 
+    // No-parameter build method that allows the use of AmpleImpl's Accumulo client.
+    // Ample methods can assign AmpleImple's Accumulo client to _client of this Builder object.
+    public TabletsMetadata build() {
+      Preconditions.checkState(level == null ^ table == null);
```
ctubbsii, line 100: This might be slightly more readable as:
```suggestion
      Preconditions.checkState((level == null) != (table == null));
```

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
```diff
@@ -371,6 +388,10 @@ private TabletsMetadata(Scanner scanner, Iterable<TabletMetadata> tmi) {
     this.tablets = tmi;
   }
 
+  public TabletsMetadata() {
+
+  }
+
```
ctubbsii, line 394: This doesn't appear to be used. Not sure it's needed.

ctubbsii (commented):
First pass: Presumably, this replaces existing code. Where is that? Can the other code be removed and anything calling that be migrated to use this?

test/src/main/java/org/apache/accumulo/test/functional/AccumuloClientIT.java
```diff
@@ -220,6 +238,120 @@ public void testClose() throws Exception {
     expectClosed(() -> tops.create("expectFail"));
     expectClosed(() -> tops.cancelCompaction(tableName));
     expectClosed(() -> tops.listSplits(tableName));
+  }
 
+  @Test
+  public void testAmpleReadTablets() throws Exception {
+
+    try (AccumuloClient accumuloClient = Accumulo.newClient().from(getClientProps()).build()) {
+      accumuloClient.securityOperations().grantTablePermission(accumuloClient.whoami(),
```
keith-turner, line 247: An alternative way to write this test would be to create a table w/ desired splits and then get the table ID.

test/src/main/java/org/apache/accumulo/test/functional/AccumuloClientIT.java
```diff
@@ -220,6 +238,120 @@ public void testClose() throws Exception {
     expectClosed(() -> tops.create("expectFail"));
     expectClosed(() -> tops.cancelCompaction(tableName));
     expectClosed(() -> tops.listSplits(tableName));
+  }
 
+  @Test
+  public void testAmpleReadTablets() throws Exception {
```
keith-turner, line 244: Why put this test in AccumuloClientIT?

core/src/main/java/org/apache/accumulo/core/metadata/schema/AmpleImpl.java
```diff
@@ -42,4 +43,12 @@ public TabletMetadata readTablet(KeyExtent extent, ColumnType... colsToFetch) {
       return Iterables.getOnlyElement(tablets);
     }
   }
+
+  @Override
+  public TabletsMetadata.Builder readTablets() {
+    this.builder = new TabletsMetadata.Builder();
+    builder._client = this.client;
```
keith-turner, line 50: It would be cleaner to pass the client to the constructor if possible.

core/src/main/java/org/apache/accumulo/core/metadata/schema/AmpleImpl.java
```diff
@@ -27,6 +27,7 @@
 
 public class AmpleImpl implements Ample {
   private final AccumuloClient client;
+  private TabletsMetadata.Builder builder;
```
keith-turner, line 30: Why is this an instance variable?

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
cradal: This was done as part of an overall attempt take the builder interfaces in TabletsMetadata and move them to Ample, while leaving the implementation of the interfaces in TabletsMetadata.The readTablets() method in AmplImpl generates a TabletsMetadata builder object and assigns the AmpleImpl client to the public client in that builder object. This allows readTablets() to be called from a ClientContext. However, at this point, I think I should rethink the whole design because it breaks too many coding conventions (case in point, above).

core/src/main/java/org/apache/accumulo/core/metadata/schema/AmpleImpl.java
cradal: This was done as part of an overall attempt take the builder interfaces in TabletsMetadata and move them to Ample, while leaving the implementation of the interfaces in TabletsMetadata.The readTablets() method in AmplImpl generates a TabletsMetadata builder object and assigns the AmpleImpl client to the public client in that builder object. This allows readTablets() to be called from a ClientContext. However, at this point, I think I should rethink the whole design because it breaks too many coding conventions (case in point, above).

cradal:
Most recent push encapsulates Builder member vars and moves Builder class over to AmpleImpl. Refactored ZooKeeperInstance to use new AmpleImpl method.

core/src/main/java/org/apache/accumulo/core/metadata/schema/AmpleImpl.java
cradal: Latest push removes this var.

test/src/main/java/org/apache/accumulo/test/functional/AccumuloClientIT.java
cradal: This is here to facilitate calling the method from a AccumuloClient.

milleruntime:
Is this change complete?  I would like to use this to read metadata in a new admin command.

core/src/main/java/org/apache/accumulo/core/metadata/schema/Ample.java
```diff
@@ -119,6 +119,15 @@ public static DataLevel of(TableId tableId) {
...
   TabletMetadata readTablet(KeyExtent extent, ColumnType... colsToFetch);
 
+  /**
+   * Entry point for reading multiple tablets' metadata. Generates a TabletsMetadata builder object
+   * and assigns the AmpleImpl client to that builder object. This allows readTablets() to be called
+   * from a ClientContext. Associated methods of the TabletsMetadata Builder class are used to
+   * generate the metadata.
+   */
+
+  AmpleImpl.Builder readTablets();
```
milleruntime, line 129: This exposes the Impl class on the interface.  This breaks the contract of the interface.

core/src/main/java/org/apache/accumulo/core/metadata/schema/AmpleImpl.java
```diff
@@ -34,12 +70,212 @@ public AmpleImpl(AccumuloClient client) {
 
   @Override
   public TabletMetadata readTablet(KeyExtent extent, ColumnType... colsToFetch) {
-    Options builder = TabletsMetadata.builder().forTablet(extent);
+    TabletsMetadata.Options builder = TabletsMetadata.builder().forTablet(extent);
     if (colsToFetch.length > 0)
       builder.fetch(colsToFetch);
 
     try (TabletsMetadata tablets = builder.build(client)) {
       return Iterables.getOnlyElement(tablets);
```
milleruntime, line 78: A very simple alternative to this change, would be to just have a shared method return `TabletsMetadata` (before calling getOnlyElement).  Is there a reason just returning `TabletsMetadata` for readTablets() wouldn't suffice?

The method on the interface could look like this:
`public TabletsMetadata readTablets(KeyExtent extent, ColumnType... colsToFetch)`

milleruntime: Nevermind, it doesn't look like this would work with the way that we use TabletsMetadata.

core/src/main/java/org/apache/accumulo/core/metadata/schema/Ample.java
milleruntime: I think this method should return `TabletsMetadata.TableOptions`.  That will limit the next method calls to what is needed for the fluent entry point and help prevent coding errors.

We want to select the table first like this:
```
ample.readTablets().forTable(tableId).fetch(LOCATION, PREV_ROW).build();
```
And not be able to skip the required parameters like this:
```
ample.readTablets().build(); //coding error
```

milleruntime (changes requested):

core/src/main/java/org/apache/accumulo/core/metadata/schema/Ample.java
cradal: Incorporated suggestions on latest push.

server/base/src/test/java/org/apache/accumulo/server/master/state/RootTabletStateStoreTest.java
```diff
@@ -54,6 +55,11 @@ public TabletMetadata readTablet(KeyExtent extent, ColumnType... colsToFetch) {
       return RootTabletMetadata.fromJson(json).convertToTabletMetadata();
     }
 
+    @Override
+    public AmpleImpl.Builder readTablets() {
+      return null;
+    }
```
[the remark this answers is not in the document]
cradal: Added UnsupportedOperationException.

core/src/main/java/org/apache/accumulo/core/metadata/schema/AmpleImpl.java
```diff
@@ -34,12 +70,212 @@ public AmpleImpl(AccumuloClient client) {
...
     }
   }
+
+  @Override
+  public AmpleImpl.Builder readTablets() {
+    Builder builder = new Builder(client);
+    return builder;
+  }
+
+  public static class Builder implements Iterable<TabletMetadata> {
```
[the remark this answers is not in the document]
cradal: Moved test to MetadataIT.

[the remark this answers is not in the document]
cradal: Incorporated suggestions on latest push.

test/src/main/java/org/apache/accumulo/test/functional/MetadataIT.java
```diff
@@ -156,4 +171,122 @@ public void batchScanTest() throws Exception {
...
+  @Test
+  public void testAmpleReadTablets() throws Exception {
+
+    try (AccumuloClient accumuloClient = Accumulo.newClient().from(getClientProps()).build()) {
+      accumuloClient.securityOperations().grantTablePermission(accumuloClient.whoami(),
+          MetadataTable.NAME, TablePermission.WRITE);
+      BatchWriter bw =
+          accumuloClient.createBatchWriter(MetadataTable.NAME, new BatchWriterConfig());
+      ClientContext cc = (ClientContext) accumuloClient;
+      // Create a fake METADATA table with these splits
```
milleruntime, line 184: Why not just use the API to create a table with those splits and let Accumulo populate the metadata?

cradal: That idea is more economical. I guess an effective test does not always need a full-blown client. I will update the test.

core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java
```diff
@@ -79,6 +79,15 @@
...
     private TableId tableId;
+    private AccumuloClient _client;
+
+    Builder(AccumuloClient client) {
+      this._client = client;
+    }
+
+    Builder() {
+
+    }
```
milleruntime, line 90: Should only have one constructor and one build method to prevent coding errors.  Since you have access to the AccumuloClient object at the time you are calling the method, I think it would be best to go with the `Builder(AccumuloClient client)` constructor and then end it with the `build()` method.  This will only pass in the client once and prevent it from being null when calling `build()`.

cradal: One issue with this is that there are ~15 classes that already use the pre-existing no-parameter constructor. Perhaps I should find a different way to pass the AmpleImpl Accumulo client into the TabletsMetadata builder object.

cradal: I might be able to refactor those classes to use the new Builder(Accumulo client) constructor.

milleruntime: That is OK.  I should only be 1 or 2 lines per class.

cradal: In the latest push: Refactored the other classes to use the new constructor. Cleaned up/streamlined unit test in MetadataIT.

ctubbsii:
@cradal There are several comments from older iterations of this, that are possibly no longer relevant. For those conversations, if you have already incorporated the suggestions, you can mark them as "Resolved", so that way the UI doesn't show them as prominently. That will make it easier to track current, ongoing discussions related to the latest iteration of this PR.

milleruntime:
@cradal Do you have plans on finishing this PR? If not, I can get take it over if you'd like. I recently used this code and have at least one place it could be used in #1821

cradal:
I would like to finish it up. I just wasn't sure if the changes I last made
were okay or not.

ctubbsii:
I attempted to resolve the merge conflicts to help out in getting this PR back into a healthy state. I will leave the rest to @cradal and/or @milleruntime

test/src/main/java/org/apache/accumulo/test/functional/AccumuloClientIT.java
```diff
@@ -46,11 +46,15 @@
...
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Iterables;
 
 public class AccumuloClientIT extends AccumuloClusterHarness {
 
+  Logger log = LoggerFactory.getLogger(AccumuloClientIT.class);
+
```
milleruntime, line 57: Old modifications that can be dropped.

milleruntime (commented):
Your most recent changes look good.  I only had one comment on some leftover code and then I think its ready to merge.

milleruntime:
@cradal If you don't have any objections, I will go ahead and finish up the work on this PR.

cradal:
@milleruntime No objections.

milleruntime:
Replaced by #1945

PR #1654, opened 2020-07-09
Added unit test to exercise bulk import range sanity check code #1618

keith-turner:


keith-turner:
@cradal  would you be interested in looking over this?  Its a follow up to #1637

cradal:
@keith-turner Sure.

core/src/test/java/org/apache/accumulo/core/clientImpl/bulk/BulkImportTest.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.TreeMap;
+
+import org.apache.accumulo.core.clientImpl.bulk.Bulk.FileInfo;
+import org.apache.accumulo.core.clientImpl.bulk.Bulk.Files;
+import org.apache.accumulo.core.data.TableId;
+import org.apache.accumulo.core.dataImpl.KeyExtent;
+import org.apache.hadoop.io.Text;
+import org.junit.Assert;
+import org.junit.Test;
+
```
cradal, line 33: @keith-turner Try adding this line. Looks like the checkstyle is execting Assert methods to be static imports.

import static org.junit.Assert.assertEquals;

cradal:
@keith-turner The test runs fine. Looks good.

core/src/test/java/org/apache/accumulo/core/clientImpl/bulk/BulkImportTest.java
keith-turner: I updated to static imports in bc6fdc1

PR #1684, opened 2020-08-14
Update jcommander to 1.78

milleruntime:
I thought upgrading JC was an issue but I think those problems may have been fixed since 1.71

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,7 +158,7 @@
       <dependency>
         <groupId>com.beust</groupId>
         <artifactId>jcommander</artifactId>
-        <version>1.48</version>
+        <version>1.78</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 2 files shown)


milleruntime:
I was able to run this update in uno and do some things in the shell no problem.  So I think it is OK.

milleruntime:
The main branch was already updated to 1.78 and now the ShellConfigTest is different...  Ugh

milleruntime:
I will make the one test in ShellConfigTest the same in 1.10 and then try and do a merge fwd.

milleruntime:
> I will make the one test in ShellConfigTest the same in 1.10 and then try and do a merge fwd.

Nevermind... I am just doing a no-op merge forward, the test changed a lot.

Issue #1689, opened 2020-08-19
Tserver in bad state may be writing corrupted files during compactions.


EdColeman:
The symptom is that the system is encountering an IOException (incorrect data check) where the zlib decompressor cannot uncompress a file.  The file would have been created during a previous compaction.

The tserver seems to be in a bad state, and rather than writing corrupt files, it would be preferable if the condition(s) can't be corrected, then stop the server.

On processing the corrupted files:

The IOException is being thrown on lines 348 or 385 of Compactor.compactLocalityGroup. (exceptions occur on both)
https://github.com/apache/accumulo/blob/7a2d12eaf785f924f555733a54f40828cdb2414f/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Compactor.java#L348
or
https://github.com/apache/accumulo/blob/7a2d12eaf785f924f555733a54f40828cdb2414f/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Compactor.java#L385

The file(s) can be partially processed with rfile-info.  You can examine the file metadata, but if you do something that reads the entire file (like -keyStats) then it fails with an exception reported in line 815 of RFile LocalityGroupReader.

When the files are being compacted it looks like the tserver was in an unhealthy state.  Potential contributors:

1) The dynamic class loader kept continually rebuilding.  AccumuloReloadingVFSClassLoader run loop just keeps executing - the message at line 83 appears frequently in the log.
https://github.com/apache/accumulo/blob/7a2d12eaf785f924f555733a54f40828cdb2414f/start/src/main/java/org/apache/accumulo/start/classloader/vfs/AccumuloReloadingVFSClassLoader.java#L83

2) The tablet servers are throwing a null pointer exception that is being logged by https://github.com/apache/accumulo/blob/7a2d12eaf785f924f555733a54f40828cdb2414f/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServerResourceManager.java#L452

There may be other issues, but these stand out in the logs.

EdColeman:
I removed the 1.10 tag so that it does not impend the 1.10 release.

[...]

PR #1709, opened 2020-09-17
DO NOT MERGE - concept for discussion - code would not work as is

EdColeman:
This is just a sample for discussion.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServerResourceManager.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServerResourceManager.java
@@ -36,7 +36,6 @@ import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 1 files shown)


EdColeman:
Closing this and will explore suggests from @dlmarion w.r.t. UnexpectedException handler in https://github.com/apache/accumulo/issues/1689

server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServerResourceManager.java
```diff
@@ -449,7 +584,33 @@ private void manageMemory() {
           mma = memoryManager.getMemoryManagementActions(tabletStates);
 
         } catch (Throwable t) {
+
           log.error("Memory manager failed {}", t.getMessage(), t);
+
+          // treat error as unrecoverable - get zookeeper and delete server lock to kill instance.
```
keith-turner, line 590: If a tablet server needs to kill itself, it would probably be better to halt rather than deleting the lock.  If things are not going well inside the process, deleting the lock may not kill the process.  Halting probably has a better chance.

PR #1726, opened 2020-10-07
Drop FileRef class

milleruntime:
* Replace remaining uses with Path objects and removed unnecessary
wrapping of intermediate objects
* Replaced with more specific types introduced in #1519 #1501 #1471 
* This class was confusing and ambiguous


ctubbsii (changes requested):
This looks okay, but I added my feedback for additional improvements in a PR against your branch at https://github.com/milleruntime/accumulo/pull/14

milleruntime:
All the ITs are passing, except a known failure in #1574 
https://github.com/apache/accumulo/runs/1227340841?check_suite_focus=true

ctubbsii (approved):

PR #1751, opened 2020-10-22
WIP Design: Interface for handling tablet state migration

milleruntime:
Some objected oriented design ideas to clean up Master code in ```TabletGroupWatcher```.  The 
idea is to replace the ```TabletGoalState``` enum with concrete classes, to eliminate some of the complex if/else statements and switch cases.  It starts with a Goal interface, implemented by two Abstract types (AbstractGoal + AbstractHostedGoal).  The concrete types are classes which combine current state + goal state.  The classes are one of two goal categories of HOSTED or not (UNASSIGNED, DELETED, SUSPENDED), since the work is the same for the non hosted goals.  I haven't figured out how to integrate the ```MergeState``` that is also used to determine the Goal state [here](https://github.com/apache/accumulo/blob/62110a3c27b9a27d8a7c0bc272aa6778910acad7/server/manager/src/main/java/org/apache/accumulo/master/Master.java#L587).

### DIFF
--- a/server/manager/src/main/java/org/apache/accumulo/master/Master.java
+++ b/server/manager/src/main/java/org/apache/accumulo/master/Master.java
@@ -543,24 +543,23 @@ public class Master extends AbstractServer
     }
   }
 
-  TabletGoalState getSystemGoalState(TabletLocationState tls) {
+  TabletGoalState getSystemGoalState(KeyExtent extent) {
     switch (getMasterState()) {
       case NORMAL:
         return TabletGoalState.HOSTED;
       case HAVE_LOCK: // fall-through intended
       case INITIAL: // fall-through intended
       case SAFE_MODE:
-        if (tls.extent.isMeta()) {
+        if (extent.isMeta()) {
           return TabletGoalState.HOSTED;
         }
         return TabletGoalState.UNASSIGNED;
       case UNLOAD_METADATA_TABLETS:
-        if (tls.extent.isRootTablet()) {
+        if (extent.isRootTablet()) {
           return TabletGoalState.HOSTED;
         }
         return TabletGoalState.UNASSIGNED;
       case UNLOAD_ROOT_TABLET:
-        return TabletGoalState.UNASSIGNED;
       case STOP:
         return TabletGoalState.UNASSIGNED;
       default:
@@ -587,8 +586,9 @@ public class Master extends AbstractServer
   TabletGoalState getGoalState(TabletLocationState tls, MergeInfo mergeInfo) {
     KeyExtent extent = tls.extent;
     // Shutting down?
-    TabletGoalState state = getSystemGoalState(tls);
+    TabletGoalState state = getSystemGoalState(extent);
     if (state == TabletGoalState.HOSTED) {
+      // TODO shouldn't this be in the upgrade code? this seems very inefficient
       if (!upgradeCoordinator.getStatus().isParentLevelUpgraded(extent)) {
         // The place where this tablet stores its metadata was not upgraded, so do not assign this
         // tablet yet.
--- a/server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
+++ b/server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
@@ -218,9 +218,9 @@ abstract class TabletGroupWatcher extends Daemon {
             var mStats = currentMerges.get(k);
             return mStats != null ? mStats : new MergeStats(new MergeInfo());
           });
-          TabletGoalState goal = master.getGoalState(tls, mergeStats.getMergeInfo());
           // TODO replace TabletGoalState with Goal interface so this will look like
           // Goal goal = master.getGoalState(tls, mergeStats.getMergeInfo());
+          Goal goal = new AssignedTabletHostedGoal(master, tls); // temp
           TServerInstance server = tls.getServer();
           TabletState state = tls.getState(currentTServers.keySet());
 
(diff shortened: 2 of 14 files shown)


keith-turner:
Is the following the code you are trying to simplify?  

https://github.com/apache/accumulo/blob/353c611891a8b88b307dcba5e884467d4191b1f2/server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java#L252-L343

How much of your goals could accomplished by replacing code blocks w/ functions like below?   I think you are also trying to simplify master.getGoalState() and possibly consolidate concerns that currently straddle master.getGoalState() and TabletGroupWatcher?  If that is true, then it seems like the following would not accomplish all of your goals.

```java
          if (goal == TabletGoalState.HOSTED) {
            if (state != TabletState.HOSTED && !tls.walogs.isEmpty()) {
              if (master.recoveryManager.recoverLogs(tls.extent, tls.walogs))
                continue;
            }
            switch (state) {
              case HOSTED:
                if (server.equals(master.migrations.get(tls.extent)))
                  master.migrations.remove(tls.extent);
                break;
              case ASSIGNED_TO_DEAD_SERVER:
                hostTabletAssignedToDeadServer(...);
                break;
              case SUSPENDED:
                hostSuspendedTablet(...);
                break;
              case UNASSIGNED:
                hostUnassignedTablet(....);
                break;
              case ASSIGNED:
                hostAssignedTablet(...);
                break;
            }
          } else {
            switch (state) {
              case SUSPENDED:
                unassignSuspendedTablet(...);
                break;
              case UNASSIGNED:
                cancelOfflineTableMigrations(tls);
                break;
              case ASSIGNED_TO_DEAD_SERVER:
                unassignAssignedToDeadServerTablet(...);
                break;
              case HOSTED:
                unassignHostedTablet(....);
                break;
              case ASSIGNED:
                break;
            }
          }
          counts[state.ordinal()]++;
        }
```

As for the merge code, that code straddles FATE and the Tablet goal state code in a way that is hard to follow.  It may need improvement, but I am uncertain what the improvement is.

milleruntime:
> Is the following the code you are trying to simplify?
> 
> [TabletGroupWatcher link](https://github.com/apache/accumulo/blob/353c611891a8b88b307dcba5e884467d4191b1f2/server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java#L252-L343)

Yes that is where I started.

> How much of your goals could accomplished by replacing code blocks w/ functions like below? 

This is certainly an improvement.  We could make this change, regardless of any other refactoring I do.  It makes this code block easier to read at least.

> I think you are also trying to simplify master.getGoalState() and possibly consolidate concerns that currently straddle master.getGoalState() and TabletGroupWatcher? If that is true, then it seems like the following would not accomplish all of your goals.

Correct.

> As for the merge code, that code straddles FATE and the Tablet goal state code in a way that is hard to follow. It may need improvement, but I am uncertain what the improvement is.

I agree.  I was aiming for a design that would allow adding more states to handle this code as well.

server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
```diff
@@ -244,6 +246,7 @@ public void run() {
             }
```
milleruntime, line 246: With this design, the "Always follow through with assignments" check would not need to exist, they would have this logic built into the types. For example, ```AssignedTabletHostedGoal``` is the only class that has a current state of ```ASSIGNED```. 

I am not sure what this second check is trying to do... something with ordering during shutdown of the whole cluster?

server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
```diff
@@ -230,117 +232,25 @@ public void run() {
           sendChopRequest(mergeStats.getMergeInfo(), state, tls);
           sendSplitRequest(mergeStats.getMergeInfo(), state, tls);
 
-          // Always follow through with assignments
-          if (state == TabletState.ASSIGNED) {
-            goal = TabletGoalState.HOSTED;
-          }
+          // Always follow through with assignments will give AssignedTabletHostedGoal
```
milleruntime, line 235: With this design, the "Always follow through with assignments" check would not need to exist, they would have this logic built into the types. For example, ```AssignedTabletHostedGoal``` is the only class that has a current state of ```ASSIGNED```.

server/manager/src/main/java/org/apache/accumulo/master/Master.java
```diff
@@ -587,8 +586,9 @@ TabletGoalState getTableGoalState(KeyExtent extent) {
   TabletGoalState getGoalState(TabletLocationState tls, MergeInfo mergeInfo) {
     KeyExtent extent = tls.extent;
     // Shutting down?
-    TabletGoalState state = getSystemGoalState(tls);
+    TabletGoalState state = getSystemGoalState(extent);
     if (state == TabletGoalState.HOSTED) {
+      // TODO shouldn't this be in the upgrade code? this seems very inefficient
```
milleruntime, line 591: I feel like this check should be moved to the upgrade code. Perhaps it was missed in the upgrade code wrangling?  Checking on every single tablet assignment seems very inefficient.

server/manager/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java
```diff
@@ -230,117 +232,25 @@ public void run() {
...
 
           // if we are shutting down all the tabletservers, we have to do it in order
-          if (goal == TabletGoalState.SUSPENDED && state == TabletState.HOSTED) {
+          // TODO figure out what this is optimization is doing
+          if (goal instanceof HostedTabletNotHostedGoal) {
             if (master.serversToShutdown.equals(currentTServers.keySet())) {
               if (dependentWatcher != null && dependentWatcher.assignedOrHosted() > 0) {
-                goal = TabletGoalState.HOSTED;
+                goal = new HostedTabletHostedGoal(master, tls);
```
milleruntime, line 242: It looks like this optimization is only when shutting down entire cluster but I am not sure what it is doing or why.  We remove the goal if the tablet is trying to suspend when shutting down.  This is also the only spot that ```dependentWatcher``` is used.  I feel like this could be done better.

milleruntime: Master.getGoalState() which is called before this, returns a ```goal = TabletGoalState.SUSPENDED``` if the tablet is on the serversToShutdown.  Which this check is most likely overriding... if its in the list for this tablet's dependentWatcher?  Ugh this code is nearly impossible to follow.

milleruntime:
After much pain looking at this code, I created #1761 with the best I can come up with for now. @keith-turner I incorporated some of your ideas into this change.  I will close this Draft design PR for now since #1761 may be the best that comes of it.

PR #1771, opened 2020-11-06
Add tablename option to shell insert command.

jmark99:
This change adds the <code>--tablename</code> option to the insert command when inside the shell or when using the accumulo command with <code>accumulo shell -e \<task\></code>. The option is not required, but if used it allows data to be inserted into a table in situations where the shell is not within a table context or when in a table context to insert data into another table.

Although I would not expect this option to be utilized very often, I found myself in need of it when working with the various accumulo examples in the accumulo-examples repo. I had many situations where I wished to insert some test data from a bash shell using the form:

<code>accumulo shell -u user -p pwd -e "insert r f q v"</code>

but this would fail due to there being no table name context for the insertion. This update allows data inserts from a command line using the form: 

<code>accumulo shell -u user -p pwd -e "insert r f q v --tablename destination_table" </code>

Note, there is no shortened form of the option to minimize confusion with the -ts option.

### DIFF
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
@@ -171,7 +171,7 @@ public class InsertCommand extends Command {
         "durability to use for insert, should be one of \"none\" \"log\" \"flush\" or \"sync\"");
     o.addOption(durabilityOption);
 
-    o.addOption(OptUtil.tableOpt("table in which data is to be inserted"));
+    o.addOption(OptUtil.tableOpt("table into which data will be inserted"));
 
     return o;
   }
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
@@ -50,6 +50,7 @@ public class InsertCommand extends Command {
   private Option insertOptAuths, timestampOpt;
   private Option timeoutOption;
   private Option durabilityOption;
+  private Option tableNameOption;
 
   protected long getTimeout(final CommandLine cl) {
     if (cl.hasOption(timeoutOption.getLongOpt())) {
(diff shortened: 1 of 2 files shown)


shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
```diff
@@ -63,81 +64,96 @@ protected long getTimeout(final CommandLine cl) {
...
+    String initialTable = null;
     try {
-      bw.close();
-    } catch (MutationsRejectedException e) {
-      final ArrayList<String> lines = new ArrayList<>();
-      if (!e.getSecurityErrorCodes().isEmpty()) {
-        lines.add("\tAuthorization Failures:");
+      if (cl.hasOption(tableNameOption.getOpt())) {
+        initialTable = shellState.getTableName();
+        shellState.setTableName(cl.getOptionValue(tableNameOption.getOpt()));
```
ctubbsii, line 72: This command should use the table specified, but it should not update the table state for the entire shell. All you need to do is set a local variable to either the specified table, or default to the current shell state's table, and use that local variable in the line below that creates the batch writer.

ctubbsii (changes requested):

shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
jmark99: Thanks for pointing that out @ctubbsii. I've updated the code to use a temp variable and left the shellstate as is.

shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
```diff
@@ -171,6 +180,11 @@ public Options getOptions() {
         "durability to use for insert, should be one of \"none\" \"log\" \"flush\" or \"sync\"");
     o.addOption(durabilityOption);
 
+    tableNameOption =
+        new Option("", "tablename", true, "name of table in which data is being inserted");
+    tableNameOption.setArgName("tablename");
+    o.addOption(tableNameOption);
+
```
ctubbsii, line 187: What's the reason for not supporting `-t` short name?

jmark99: I chose not to use the short form to minimize any confusion with the <code>timestamp</code> parameter (-ts). There is also a <code>timeout<code> parameter that does not off a shortened name. If you think it would be nice to include one, I have no problem adding one. Could always use '-n' for 'name' perhaps. Let me know your thoughts.

ctubbsii: If there's not already a `-t`, I don't think it's a problem. I don't think there's any concern over conflicting with `-ts`. In fact, most commands, use the standard `OptUtil.tableOpt`. See  https://github.com/apache/accumulo/blob/main/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java#L93 and https://github.com/apache/accumulo/blob/main/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java#L395.
I would recommend doing the same here, if there's no existing `-t`.

shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
```diff
@@ -171,6 +171,8 @@ public Options getOptions() {
         "durability to use for insert, should be one of \"none\" \"log\" \"flush\" or \"sync\"");
     o.addOption(durabilityOption);
 
+    o.addOption(OptUtil.tableOpt("table in which data is to be inserted"));
```
ctubbsii, line 174: Minor grammatical nit:
```suggestion
    o.addOption(OptUtil.tableOpt("table into which data will be inserted"));
```

ctubbsii (changes requested):
Sorry, very minor grammatical nit, but otherwise looks good to me.

shell/src/main/java/org/apache/accumulo/shell/commands/InsertCommand.java
jmark99: No problem. Didn't quite sound right to me, but I was blanking on how to re-word it.

ctubbsii (approved):

EdColeman:
I realize this has been merged - but I have a more general question.  This is at least the third time that a table option has been added to a shell command (I added it for importdir, there was another, but I don't recall specifics) and then this one.  Would it be worth creating an issue and checking that we are consistent across all of the commands that operate on a table that they accept this as an option?

ctubbsii:
@EdColeman I think that would be a reasonable thing to do if you want to pursue it.

jmark99:
Hi Ed,

I think that would be a good idea. Unless you wish to pursue it, I would be willing to take a look since I’m currently reviewing the Accumulo-examples code and coming across various commands that may need the option. Let me know which you would prefer.

Mark

EdColeman:
I do not mind at all!

jmark99:
I created #1776 to address this issue.

PR #1802, opened 2020-11-20
Convert Translator usages to lambdas

### LINKED ISSUE #1533: Convert Translators to lambdas

lbschanno:
Replace Translator.java and Translators.java with native Java lambda
functions.

Fixes #1533


ctubbsii (approved):
Thanks @lbschanno . These changes look good to me.

PR #1823, opened 2020-12-04
Rename TabletState interface to TabletMem in tserver

milleruntime:
* Refactor tserver internal interface, renaming from TabletState to
TabletMem to differentiate from TabletState in master

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/tabletserver/LargestFirstMemoryManager.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/tabletserver/LargestFirstMemoryManager.java
@@ -147,7 +147,7 @@ public class LargestFirstMemoryManager {
     return context.getTableConfiguration(tableId) != null;
   }
 
-  public MemoryManagementActions getMemoryManagementActions(List<TabletState> tablets) {
+  public MemoryManagementActions getMemoryManagementActions(List<TabletMem> tablets) {
     if (maxMemory < 0)
       throw new IllegalStateException(
           "need to initialize " + LargestFirstMemoryManager.class.getName());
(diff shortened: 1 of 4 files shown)


ctubbsii (commented):
A rename is good, but I don't know what "TabletMem" means from its name. It would be *really* helpful to have a more descriptive name (rather than just a unique one to differentiate it). Or at the very least *some* kind of javadoc on the interface definition, so that when somebody sees "TabletMem" in the code and asks "what is that?", they could gain an understanding relatively quickly from the javadoc.

Off the top of my head, I'm thinking "TabletMemState" might be a little better, so even if I don't know what "Mem" means, I know the class represents some kind of "state" pertaining to a "tablet".

But, the javadoc would still be helpful to understand what kind of "state" and what the "mem" means, like what's this state for? how is it used? why does it exist? The method names in the interface provide clues, but without diving deeper, I don't really know the answer to these questions... they are just the first thing I think when seeing a name like "TabletMem" (or even "TabletMemState").

milleruntime:
> A rename is good, but I don't know what "TabletMem" means from its name. It would be _really_ helpful to have a more descriptive name (rather than just a unique one to differentiate it). Or at the very least _some_ kind of javadoc on the interface definition, so that when somebody sees "TabletMem" in the code and asks "what is that?", they could gain an understanding relatively quickly from the javadoc.
> 
> Off the top of my head, I'm thinking "TabletMemState" might be a little better, so even if I don't know what "Mem" means, I know the class represents some kind of "state" pertaining to a "tablet".

I actually had TabletMemState but changed it to TabletMem because it was shorter.  The derived classes all seemed too longer to me... TabletMemStateImpl or TestTabletMemState, especially since the class is usually the impl quite a bit.  The only thing I extracted about this inner class from looking at the code was that it had to do with Memory.  What about MemoryState, TabMemory, TMemoryState, MemState?

> But, the javadoc would still be helpful to understand what kind of "state" and what the "mem" means, like what's this state for? how is it used? why does it exist? The method names in the interface provide clues, but without diving deeper, I don't really know the answer to these questions... they are just the first thing I think when seeing a name like "TabletMem" (or even "TabletMemState").

I agree but your guess is as good as mine.  I attempted to get rid of the interface all together but it is needed since the test uses KeyExtent but the actual server code uses Tablet.

milleruntime:
Superseded by #1826

PR #1860, opened 2021-01-07
Use long instead of int for monitor scan rate (#1858)

brianloss:
Ran into a case on a large cluster where the scan rate was so high that the graph flatlined at Java's max integer value.

### DIFF
--- a/server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
+++ b/server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
@@ -154,10 +154,10 @@ public class Monitor {
       Collections.synchronizedList(new MaxList<Integer>(MAX_TIME_PERIOD));
   private static final List<Pair<Long,Double>> lookupsOverTime =
       Collections.synchronizedList(new MaxList<Double>(MAX_TIME_PERIOD));
-  private static final List<Pair<Long,Integer>> queryRateOverTime =
-      Collections.synchronizedList(new MaxList<Integer>(MAX_TIME_PERIOD));
-  private static final List<Pair<Long,Integer>> scanRateOverTime =
-      Collections.synchronizedList(new MaxList<Integer>(MAX_TIME_PERIOD));
+  private static final List<Pair<Long,Long>> queryRateOverTime =
+      Collections.synchronizedList(new MaxList<Long>(MAX_TIME_PERIOD));
+  private static final List<Pair<Long,Long>> scanRateOverTime =
+      Collections.synchronizedList(new MaxList<Long>(MAX_TIME_PERIOD));
   private static final List<Pair<Long,Double>> queryByteRateOverTime =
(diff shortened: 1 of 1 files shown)


keith-turner (approved):
The changes look good.  Curious if you were able to test the changes and see scan rates higher than max int on the monitor.

brianloss:
> Curious if you were able to test the changes and see scan rates higher than max int on the monitor.

I don't have the large cluster available right now, but I was able to simulate by modifying the monitor code and adding `Integer.MAX_VALUE` to every result coming back. That worked well:

image

ctubbsii:
@brianloss When you merge, it's okay to do a squash merge. That's what we've typically been doing. It helps keep the history a bit cleaner than it might otherwise be. Also, I see you didn't merge this forward into the main branch after merging it into 1.10. Was that an oversight, or are you still working on merging it to main? Does this issue apply to the main branch?

brianloss:
@ctubbsii, thanks. There was only the 1 commit, so I didn't think a squash merge was necessary. Did I miss anything?
I have a local merge to the main branch. It was a bit more work, since the monitor is split up. However, when testing it, I find that I'm getting a flat line graph at integer max value. I'm digging through that now. I think it may be some issue with the plotting library used.

ctubbsii:
> @ctubbsii, thanks. There was only the 1 commit, so I didn't think a squash merge was necessary. Did I miss anything?

When you merge from the UI, even with only one commit, GitHub creates a merge commit. That's why we squash, even if there's only 1. :smiley_cat: 

> I have a local merge to the main branch. It was a bit more work, since the monitor is split up. However, when testing it, I find that I'm getting a flat line graph at integer max value. I'm digging through that now. I think it may be some issue with the plotting library used.

Okay, that's fine. If you want to merge and deconflict by not making any changes, that will ensure that the branches are ready for others to merge (as needed) while you work through the issues. Then, you can create a new PR (or new commit) that would apply directly to the main branch later.

brianloss:
> > @ctubbsii, thanks. There was only the 1 commit, so I didn't think a squash merge was necessary. Did I miss anything?
> 
> When you merge from the UI with the default merge button, even with only one commit, GitHub creates a merge commit. That's why we tend to select the squash option from the dropdown, even if there's only 1. 😺
> 

Gotcha, my bad. I'll squash everything from now on.

> > I have a local merge to the main branch. It was a bit more work, since the monitor is split up. However, when testing it, I find that I'm getting a flat line graph at integer max value. I'm digging through that now. I think it may be some issue with the plotting library used.
> 
> Okay, that's fine. If you want to merge and deconflict by not making any changes, that will ensure that the branches are ready for others to merge (as needed) while you work through the issues. Then, you can create a new PR (or new commit) that would apply directly to the main branch later.

I thought maybe I could resolve the discrepancy quickly, but it's getting late enough for me today that I agree it's better to merge up as a no-op and do the changes from scratch on the main branch.

Issue #1844, opened 2020-12-21
Error recovering metadata tablet


milleruntime:
I saw 2 metadata tablets fail to recover during a long running session of the Randomwalk test.  The error that was reported read:
"Error recovering tablet !0;51e7e10b1827d193;490d5011198f9047 from log files"
Here is the stacktrace from the tserver trying to load the tablet:
```
2020-12-20 00:45:53,512 [tserver.TabletServer] WARN : exception trying to assign tablet !0;51e7e10b1827d193;490d5011198f9047 hdfs://muchoshacluster/accumulo/tables/!0/t-0000yt4
java.lang.RuntimeException: Error recovering tablet !0;51e7e10b1827d193;490d5011198f9047 from log files
        at org.apache.accumulo.tserver.tablet.Tablet.<init>(Tablet.java:499)
        at org.apache.accumulo.tserver.TabletServer$AssignmentHandler.run(TabletServer.java:2413)
        at org.apache.accumulo.fate.util.LoggingRunnable.run(LoggingRunnable.java:35)
        at org.apache.accumulo.tserver.ActiveAssignmentRunnable.run(ActiveAssignmentRunnable.java:64)
        at org.apache.htrace.wrappers.TraceRunnable.run(TraceRunnable.java:57)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at org.apache.accumulo.fate.util.LoggingRunnable.run(LoggingRunnable.java:35)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.io.IOException: Unable to find recovery files for extent !0;51e7e10b1827d193;490d5011198f9047 logEntry: !0;51e7e10b1827d193; hdfs://muchoshacluster/accumulo/wal/worker1+9997/e41008f5-edcf-4e18-aa
92-ba5ff021685c
        at org.apache.accumulo.tserver.TabletServer.recover(TabletServer.java:3311)
        at org.apache.accumulo.tserver.tablet.Tablet.<init>(Tablet.java:437)
        ... 8 more
```

[...]

PR #1863, opened 2021-01-11
Add check for finished to LogSorter

milleruntime:
* Check for finished file in LogSorter before performing sort
* LogSorter was always removing the destination sort directory even if
it had previously been successfully sorted, creating lots of duplicate
and unnecessary work during recovery

I discovered this issue while looking into #1844

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/log/LogSorter.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/log/LogSorter.java
@@ -93,7 +93,6 @@ public class LogSorter {
       }
 
       try {
-        log.info("Copying " + src + " to " + dest);
         sort(sortId, new Path(src), dest);
       } finally {
         currentWork.remove(sortId);
@@ -110,7 +109,13 @@ public class LogSorter {
       String formerThreadName = Thread.currentThread().getName();
       int part = 0;
       try {
+        // check for finished first since another thread may have already done the sort
+        if (fs.exists(SortedLogState.getFinishedMarkerPath(destPath))) {
+          log.debug("Sorting already finished at {}", destPath);
+          return;
+        }
 
+        log.info("Copying " + srcPath + " to " + destPath);
         // the following call does not throw an exception if the file/dir does not exist
         fs.deleteRecursively(new Path(destPath));
 


Manno15 (approved):
These look good to me.

server/tserver/src/main/java/org/apache/accumulo/tserver/log/LogSorter.java
```diff
@@ -110,7 +109,13 @@ public void sort(String name, Path srcPath, String destPath) {
       String formerThreadName = Thread.currentThread().getName();
       int part = 0;
       try {
+        // check for finished first since another thread may have already done the sort
+        if (fs.exists(SortedLogState.getFinishedMarkerPath(destPath))) {
```
keith-turner, line 113: Did you have specific reasons for putting the check here (and not for example at the beginning of the method?).  I noticed putting the check here will execute the finally block, which seems like a good thing.

milleruntime: Mainly for the try/catch/finally since the `exists()` call has a checked IOException.  I wanted to add it earlier, in the process method before it gets added to the `currentWork` map but there isn't a try/catch there.  Then the more I thought about it, its probably not a bad thing it gets added to the `currentWork` map for this check since the `exists()` method is a Namenode operation.

milleruntime:
I tested this on my test cluster and it works well.  It prevented this duplicate work from occurring:
<pre>
tserver_2_ip-10-113-12-25.debug.log:2021-01-11 07:50:36,821 [log.LogSorter] INFO : Finished log sort 5bce00d4-1920-416a-8bfb-41e4cf16e9f6 93 bytes 1 parts in 310ms
tserver_4_ip-10-113-12-25.debug.log:2021-01-11 07:50:57,490 [log.LogSorter] INFO : Finished log sort 5bce00d4-1920-416a-8bfb-41e4cf16e9f6 93 bytes 1 parts in 479ms
tserver_4_ip-10-113-12-25.debug.log:2021-01-11 07:51:37,700 [log.LogSorter] INFO : Finished log sort 5bce00d4-1920-416a-8bfb-41e4cf16e9f6 93 bytes 1 parts in 66ms
</pre>

It should also help with the errors I am seeing in #1844

server/tserver/src/main/java/org/apache/accumulo/tserver/log/LogSorter.java
milleruntime: I think ideally the DistributedWorkQueue would be able to know that the sort was finished but that would be too much work for 1.10.  Maybe we could make improvements for 2.1

keith-turner: Thanks for the explanation @milleruntime, it helped me understand the larger context.

>  Then the more I thought about it, its probably not a bad thing it gets added to the currentWork map for this check since the exists() method is a Namenode operation.

That is a nice benefit.

> I think ideally the DistributedWorkQueue would be able to know that the sort was finished but that would be too much work for 1.10. Maybe we could make improvements for 2.1

I have not looked at the DistributedWorkQueue in a while.  If it will always have race conditions and possibly run things multiple times, then your approach here of doing this check as late as possible will be hard to beat.

keith-turner (approved):

ctubbsii (approved):
Need to be careful merging into the main branch, but I don't think this code has changed much, so it shouldn't be too much of a problem.

milleruntime:
> Need to be careful merging into the main branch, but I don't think this code has changed much, so it shouldn't be too much of a problem.

Since it is an improvement we definitely want in the main branch, I was going to hold off on merging this until I can look at the conflicts first.

PR #1864, opened 2021-01-12
Fixes #1321 makes metadata consistency level configurable for reads

### LINKED ISSUE #1321: Review how Ample reads root tablet metadata from zookeeper
Currently Ample reads root tablet metadata using ZooCache.  I am not sure this is always appropriate.  Its certainly appropriate and desirable in some cases, but I am not sure about all.  May want to add a consistency level to the Ample API and make all code choose if it wants eventual or immediate consistency.

keith-turner:


ctubbsii (approved):

milleruntime (approved):
Nice improvement.  I like that the ReadConsistency parameter is very explicit.

PR #1877, opened 2021-01-22
Update zookeeper 3.5.x version to 3.5.9

EdColeman:
This change bumps the zookeeper version to the latest 3.5.x release. With 2.1 we are changing the zookeeper version to 3.5.x, this would keep the 2.1 release consistent with the current zookeeper release.


brianloss (approved):
LGTM - 3.5.9 fixes [several CVEs](https://zookeeper.apache.org/doc/r3.5.9/releasenotes.html)

ctubbsii (approved):

PR #1910, opened 2021-02-04
Adding JShell Accumulo Feature

slackwinner:
In response to apache/accumulo#1870, this commit includes basic functionality to boot up JShell with Accumulo's API imported.


slackwinner:
This is still in a work in progress but I wanted to share what I have done so far. Essentially, I added a "jshell entry" in bin/accumulo so that the user can easily activate jshell with the classpath imported. In addition, there is a jshell-init.jsh file allows the user to load in the default Accumulo public APIs. Also the user can customize the jshell-init by adding or removing public APIs. Currently jshell-init has hard coded public APIs as I am  still figuring out how to import the classes automatically as well as making sure the classes are always up to date during a build.

ctubbsii:
@slackwinner Since you're still working on this, I converted it to a Draft pull request.

assemble/conf/create-jshell.sh
```diff
@@ -0,0 +1,134 @@
...
+      echo "$line" >> "$jPath"
+    done < "$header"
+    
+    printf "%s\n" "*/" >> "$jPath" 
+    echo " " >> "$jPath"
+  else
+    echo "Cannot add license header in $jPath"
+    echo "Please ensure jshell-init.jsh exists"
+    exit 1
+  fi
```
ctubbsii, line 40: The generated file doesn't hold any meaningful/copyrightable content. It's just generated import statements, so it's not strictly necessary to add the license header. The rat check error from the missing license header can be avoided by generating the file in the build tree, rather than the source tree. For maven, that means generating it in `assemble/target/` directory rather than `assemble/conf/`

That should simplify things a bit.

assemble/pom.xml
```diff
@@ -520,6 +520,16 @@
...
             </configuration>
           </execution>
+          <execution>
+            <id>create-jshell</id>
+            <goals>
+              <goal>exec</goal>
+            </goals>
+            <phase>compile</phase>
+            <configuration>
+              <executable>${basedir}/conf/create-jshell.sh</executable>
```
ctubbsii, line 530: Since this file should not be included as part of the distribution tarball, but is instead part of the build tools, this `create-jshell.sh` file should live in `assemble/src/main/scripts/`

```suggestion
              <executable>src/main/scripts/create-jshell.sh</executable>
```

assemble/bin/accumulo
```diff
@@ -58,7 +58,30 @@ function main() {
...
+    for arg in "${!cmdArr[@]}"; do
+       if [[ "${cmdArr[arg]}" =~ "-startup" ]]; then
+         # Insert "DEFAULT" JShell script after -startup argument
+         cmdArr=( "${cmdArr[@]:0:arg}" "DEFAULT" "${cmdArr[@]:arg+1}" )
+       fi
+    done
+    # Load in default java libraries first followed by custom file(s)
+    CLASSPATH="$CLASSPATH" "${cmdArr[@]}"
+    exit 0
+  fi
```
ctubbsii, line 83: CLASSPATH is probably already exported, so you might not need that part. If you do, then `export CLASSPATH` on the preceding line before the exec might be more clear.

Also, you can probably inject the startup file (if it exists) before the user's own arguments, as in:

```suggestion
  jShellPath="$conf/jshell-init.jsh"
  if [[ $cmd == "jshell" ]]; then
    shift
    if [[ -f "$jShellPath" ]]; then
      exec "$cmd" --startup "$jShellPath" "$@"
    else
      exec "$cmd" "$@"
    fi
  fi
```

assemble/conf/create-jshell.sh
```diff
@@ -0,0 +1,134 @@
...
+  SOURCE="${BASH_SOURCE[0]}"
+  while [[ -h "${SOURCE}" ]]; do
+    bin="$( cd -P "$( dirname "${SOURCE}" )" && pwd )"
+    SOURCE="$(readlink "${SOURCE}")"
+    [[ "${SOURCE}" != /* ]] && SOURCE="${bin}/${SOURCE}"
+  done
+    
+  # Establish file and folder paths for JShell config
+  export conf="$( cd -P "$( dirname "${SOURCE}" )" && pwd )"; 
+  export jPath="$conf/jshell-init.jsh"
```
ctubbsii, line 102: You should generate this in the `assemble/target` directory during the build, and then update the `assemble/src/main/assemblies/component.xml` file to ensure it is copied to the appropriate `conf/` directory by the `maven-assembly-plugin` for the distribution tarball.

assemble/conf/create-jshell.sh
```diff
@@ -0,0 +1,134 @@
...
+  
+  # Does a valid JShell path and source directory exists?
+  if [[ -e "$jPath" ]] && [[ -d "$srcDir" ]]; then
+    # Add API category designator in jshell-init.jsh 
+    case "$srcDir" in
+        *"$client"*) echo "// Accumulo Client API" >> "$jPath";;
+        *"$data"*) echo "// Accumulo Data API" >> "$jPath";;
+        *"$security"*) echo "// Accumulo Security API" >> "$jPath";;
+        *"$mini"*) echo "// Accumulo Minicluster API" >> "$jPath";;
+        *"$hadoop"*) echo "// Accumulo Hadoop API" >> "$jPath";;
```
ctubbsii, line 74: I like how you added section headers to the generated file!

ctubbsii (changes requested):
Nice work. See my suggestions below.

dlmarion:
As part of https://issues.apache.org/jira/browse/ACCUMULO-1399 I added the ability for users to write scripts in a supported JSR-223 engine and run them in the shell. How does adding jshell to the shell compare to this? Is it duplicative, should the script command be deprecated?

ctubbsii:
> As part of https://issues.apache.org/jira/browse/ACCUMULO-1399 I added the ability for users to write scripts in a supported JSR-223 engine and run them in the shell. How does adding jshell to the shell compare to this? Is it duplicative, should the script command be deprecated?

@dlmarion Rhino, the default scripting engine that feature used is already removed from Java, and replaced with Nashorn, which is now also deprecated and scheduled for removal in JDK 15. So, that feature won't work anymore anyway. So, yes, that script command should be deprecated.

I don't know all the use cases for the shell script command, but I suspect that they can all be replaced with the use of JShell. JShell is even better, actually, because it can be used interactively.

dlmarion:
Nashorn might be planned for removal, but the ability to plug in a JSR-223 compatible script library is not. Users can still plug in their own library, like https://github.com/eclipsesource/J2V8.

I'm fine with deprecation, we just need to make sure it makes it into the release notes so that users have time to migrate their javascript / python / whatever files to java.

assemble/conf/create-jshell.sh
slackwinner: @ctubbsii Thank you for reviewing my code submission. I'll make the proposed updates and test out the code before pushing a revised copy later.

assemble/bin/accumulo
```diff
@@ -58,6 +58,18 @@ function main() {
...
     exit 0
   fi
+  
+  # Set up path variable for default import config file
+  export jShellPath="$conf/jshell-init.jsh"
+  
+  if [[ $cmd == "jshell" ]]; then
+    shift
+    if [[ -f "$jShellPath" ]]; then
+      exec "$cmd" --startup DEFAULT "$jShellPath" "$@"
```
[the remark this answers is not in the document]
ctubbsii: I just learned that we can actually specify more than one `--startup` option, so if we don't want to customize the `java.*` imports for now, and just want the defaults, we can do:

```suggestion
      exec "$cmd" --startup DEFAULT --startup "$jShellPath" "$@"
```

ctubbsii (changes requested):

assemble/bin/accumulo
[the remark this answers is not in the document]
slackwinner: Yeah for the DEFAULT script, I originally left that as a start up parameter to ensure the user can see the default import library and the Accumulo API when they first boot up JShell. I believe having two start-up statements should do the trick. Also I believe I found a work around regarding the Scanner. Whenever a Scanner is defined, we need to explicitly specify which Scanner are we referring to. For example: 

**Referencing Accumulo Scanner:**
(org.apache.accumulo.core.client.Scanner scan = client.createScanner("GothamPD", Authorizations.EMPTY);

**Referencing Java Scanner:**
 java.util.Scanner myObj = new java.util.Scanner(System.in);

I created a Java test code where there one function uses the Accumulo Scanner and another function uses the Java Scanner. I then executed the code in JShell with Accumulo and both functions where able to execute with no issues.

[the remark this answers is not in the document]
ctubbsii: I was aware of the workaround, but it's still annoying to specify the fully-qualified class name. Since the java.util.Scanner is very unlikely to be used by Accumulo devs using this, I was thinking we could try to exclude it by replacing `java.util.*` from the defaults with a narrower set of useful imports from `java.util`, but I'm not sure what that set would look like. It would probably include a bunch of Collection classes, but I'm not sure what else.

The alternative is that we just settle for the workaround for now, and consider a new name for Scanner in Accumulo's public API in the future to make it easier to coexist with java.util.Scanner.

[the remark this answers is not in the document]
slackwinner: Would you like me to manually import all the classes from java.util.* into the jshell-init.jsh file or just go with the alternative solution?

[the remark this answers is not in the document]
ctubbsii: Let's just go with the double-startup option for now. `--startup DEFAULT --startup "$jShellPath"`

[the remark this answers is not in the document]
slackwinner: Cool. I just pushed a commit to apply your double-startup option idea into the accumulo script file.

assemble/src/main/scripts/create-jshell.sh
```diff
@@ -0,0 +1,110 @@
...
+  touch "$jPath"
+ 
+  # Create and add Accumulo APIs into API storage 
+  apiStorage=("$CLIENT" "$DATA" "$SECURITY" "$MINI" "$HADOOP")
+   
+  # Traverse through each source directory and load in Accumulo APIs
+  for srcDir in "${apiStorage[@]}"; do
+    addAccumuloAPI "$srcDir" "$jPath"
+  done
+  exit 0
```
ctubbsii, line 108: Normally at the end of a function, you'd want to let the return code of the last command propagate outside of the function. If this loop terminates abnormally, we shouldn't mask its return code with our own using `exit 0`, which forces it to terminate successfully, even if there was an error.

Also, I was thinking, you don't need to pass in the filename to the function or create it ahead of time. Instead, you can just create the file right here at the output of the whole for loop. That would simplify the implementation of the addAccumuloAPI command a lot.

```suggestion
    addAccumuloAPI "$srcDir"
  done > "$jPath"
```
One other suggestion I had was that you could move the printing of the section headers to outside that method, that way you don't need to create so many variables in that method to do the section header checks. You could either move the if statements to here, inside this for loop, before calling `addAccumuloAPI`, or you remove the loop, and do something like the following. You wouldn't even need all the extra variables this way, because you could just inline them:

```bash
  {
    echo "// section header here for client"
    addAccumuloAPI "$CLIENT"
    echo "// section header here for data"
    addAccumuloAPI "$DATA"
    # etc...
  } > "$jPath"
```

ctubbsii (changes requested):
@slackwinner I tested this end-to-end, and everything works great, functionally speaking. I did have lots of minor suggestions for polish/improving the scripts.

The one other thing that I didn't mention elsewhere is that it might be useful to have a `System.out.println("Preparing JShell for Apache Accumulo")` or similar in the startup file also. That way, people know that it was started with the custom startup script.

assemble/bin/accumulo
```diff
@@ -58,6 +58,18 @@ function main() {
     echo "$CLASSPATH"
     exit 0
   fi
+  
+  # Set up path variable for default import config file
+  export jShellPath="$conf/jshell-init.jsh"
+
```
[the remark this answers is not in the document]
slackwinner: @ctubbsii Thank you for the helpful improvements/suggestions. This code review is helping me understand bash scripting in general. I'll apply these improvements and push a revised copy soon.

assemble/src/main/scripts/create-jshell.sh
```diff
@@ -0,0 +1,92 @@
...
+    addAccumuloAPI "${apiStorage[1]}"
+    echo "// Accumulo Security API"
+    addAccumuloAPI "${apiStorage[2]}"
+    echo "// Accumulo MiniCluster API"
+    addAccumuloAPI "${apiStorage[3]}"
+    echo "// Accumulo Hadoop API"
+    addAccumuloAPI "${apiStorage[4]}"
+  } > "$jPath"
+}
+main "$@"
```
ctubbsii, line 92: I had a few other minor tips for this script, but probably not worth going back and forth over them, so here they are as one lump suggestion:

* Minor style improvements (indentation using 2 spaces instead of 3,
  space before opening brace in function declaration and around pipes
* Declare additional variables as local (SOURCE, bin) and place in
  single line at top of each function
* Do validation of srcDirs in called method, rather than in a separate
  loop, to avoid needing to use an array for the srcDirs
* Reduce number of variables needed by inlining several
* Remove redundant clearing of destination file
* Use single quote for strings that don't have embedded variables
* Move System.out.println to after the imports

```suggestion
function addAccumuloAPI() {
  local srcDir="$1" api apiPath
  # Validate each source directory before populating JShell-Init file
  if [[ ! -d "$srcDir" ]]; then
    echo "$srcDir is not a valid directory. Please make sure it exists."
    exit 1
  fi

  # Extract API info from provided source directory
  mapfile -t api < <(find "$srcDir" -type f -name '*.java' -print0 | xargs -0 -n1 dirname | sort -u)

  # Load in API and format source directory into Java import statements
  for apiPath in "${api[@]}"; do
    echo "import ${apiPath##*/java/}.*" | tr / .
  done
  echo
}

function main() {
  local SOURCE bin scriptPath mainBase corePath
  # Establish Accumulo's main base directory
  SOURCE="${BASH_SOURCE[0]}"
  while [[ -h "${SOURCE}" ]]; do
    bin="$( cd -P "$( dirname "${SOURCE}" )" && pwd )"
    SOURCE="$(readlink "${SOURCE}")"
    [[ "${SOURCE}" != /* ]] && SOURCE="${bin}/${SOURCE}"
  done

  # Establish file and folder paths for JShell config
  scriptPath="$( cd -P "$( dirname "${SOURCE}" )" && pwd )"
  mainBase="$( cd -P "${scriptPath}"/../../../.. && pwd )"
  corePath="$mainBase/core/src/main/java/org/apache/accumulo/core"

  # Create new jshell-init file
  mkdir -p "$mainBase/assemble/target"
  echo 'Generating JShell-Init file'
  {
    echo '// Accumulo Client API'
    addAccumuloAPI "$corePath/client"
    echo '// Accumulo Data API'
    addAccumuloAPI "$corePath/data"
    echo '// Accumulo Security API'
    addAccumuloAPI "$corePath/security"
    echo '// Accumulo MiniCluster API'
    addAccumuloAPI "$mainBase/minicluster/src/main/java/org/apache/accumulo/minicluster"
    echo '// Accumulo Hadoop API'
    addAccumuloAPI "$mainBase/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce"
    echo
    echo 'System.out.println("Preparing JShell for Apache Accumulo")'
    echo
  } > "$mainBase/assemble/target/jshell-init.jsh"
}

main "$@"

```

ctubbsii (approved):

assemble/src/main/scripts/create-jshell.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo "$srcDir is not a valid directory. Please make sure it exists."
+    exit 1
+  fi
+
+  # Extract API info from provided source directory
+  mapfile -t api < <(find "$srcDir" -type f -name '*.java' -print0 | xargs -0 -n1 dirname | sort -u)
+
+  # Load in API and format source directory into Java import statements
+  for apiPath in "${api[@]}"; do
+    echo "import ${apiPath##*/java/}.*" | tr / .
```
ctubbsii, line 34: Trailing semicolon isn't strictly necessary for JShell, but it's nice to have.

```suggestion
    echo "import ${apiPath##*/java/}.*;" | tr / .
```

assemble/src/main/scripts/create-jshell.sh
```diff
@@ -0,0 +1,74 @@
...
+    echo '// Accumulo Data API'
+    addAccumuloAPI "$corePath/data"
+    echo '// Accumulo Security API'
+    addAccumuloAPI "$corePath/security"
+    echo '// Accumulo MiniCluster API'
+    addAccumuloAPI "$mainBase/minicluster/src/main/java/org/apache/accumulo/minicluster"
+    echo '// Accumulo Hadoop API'
+    addAccumuloAPI "$mainBase/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce"
+    echo
+    echo 'System.out.println("Preparing JShell for Apache Accumulo")'
```
ctubbsii, line 69: While playing with these changes, I noticed that `Text` from Hadoop would be nice to have by default, as it is frequently used in our own APIs.

```suggestion
    echo '// Essential Hadoop API
    echo 'import org.apache.hadoop.io.Text;'
    echo
    echo '// Initialization code'
    echo 'System.out.println("Preparing JShell for Apache Accumulo");'
```

assemble/src/main/scripts/create-jshell.sh
```diff
@@ -0,0 +1,77 @@
...
+    addAccumuloAPI "$corePath/client"
+    echo '// Accumulo Data API'
+    addAccumuloAPI "$corePath/data"
+    echo '// Accumulo Security API'
+    addAccumuloAPI "$corePath/security"
+    echo '// Accumulo MiniCluster API'
+    addAccumuloAPI "$mainBase/minicluster/src/main/java/org/apache/accumulo/minicluster"
+    echo '// Accumulo Hadoop API'
+    addAccumuloAPI "$mainBase/hadoop-mapreduce/src/main/java/org/apache/accumulo/hadoop/mapreduce"
+    echo '// Essential Hadoop API
```
ctubbsii, line 68: Oops. I broke this with the missing quote
```suggestion
    echo '// Essential Hadoop API'
```

dlmarion:
Is there an example of how to use this?

keith-turner:
> Is there an example of how to use this?

I used this recently to quickly create some data for testing compactions.  Below is a bit of bash I wrote to use this new feature.

```bash
bin/accumulo jshell -<<EOF
var client=Accumulo.newClient().from("conf/accumulo-client.properties").build();
var writer=client.createBatchWriter("testTable");
String prev = "";
for(int i = 0; i < 2000000; i++){ var uuid = UUID.randomUUID().toString(); Mutation m = new Mutation(uuid); m.put("","",prev); writer.addMutation(m); prev = uuid;}
writer.close()
/exit
EOF
```

PR #1927, opened 2021-02-11
Document master/manager rename in Upgrader9to10 re #1922

brianloss:


server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -95,7 +95,17 @@
 import com.google.common.base.Preconditions;
 
 /**
- * Handles upgrading from 2.0 to 2.1
+ * Handles upgrading from 2.0 to 2.1.
```
brianloss, line 98: @milleruntime does this cover what you were looking for?

milleruntime: This is great, thanks. I was also thinking we could link the methods associated with the change but that might not always be easy

brianloss: I added a reference to the associated upgrade method.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -95,7 +95,17 @@
...
+ * <ul>
+ * <li><strong>Rename master properties (Issue
+ * <a href="https://github.com/apache/accumulo/issues/1640">#1640</a>):</strong> Rename any
+ * ZooKeeper system properties that start with "master." to the equivalent property starting with
+ * "manager." instead (see the {@code renameOldMasterPropsinZK(ServerContext)} method). Note that
```
milleruntime, line 103: I am not sure which spot in the javadoc is more likely for users to see (at the class or method) but you could link to the method like this:

```suggestion
 * "manager." instead (see the {@link #renameOldMasterPropsinZK(ServerContext)} method). Note that
```

brianloss: There isn't any javadoc on that private method, so I thought having a link to it would be more confusing.

milleruntime (approved):

PR #1933, opened 2021-02-16
Improvements to TabletStateChangeIteratorIT

milleruntime:
* Modified copyTable method to gather info about the metadata first
before creating the copy, preventing information about the copy itself
from being seen.
* Added debugging information

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/functional/TabletStateChangeIteratorIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/functional/TabletStateChangeIteratorIT.java
@@ -19,10 +19,12 @@ package org.apache.accumulo.test.functional;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.junit.Assert.assertEquals;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.List;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.SortedSet;
@@ -63,6 +65,8 @@ import org.apache.accumulo.server.master.state.TabletStateChangeIterator;
 import org.apache.accumulo.server.zookeeper.ZooLock;
 import org.apache.hadoop.io.Text;
 import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Sets;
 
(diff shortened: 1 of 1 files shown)


milleruntime:
These improvements seem to have fixed the issue in 2.1 as well so as soon as the checks pass I will go ahead and merge.

DomGarguilo:
Another small improvement may be to pre-split the table in the createTable method. I'm not too sure what is going on with the online/offline of the table, however if it is found that the split does not need to occur after the table is created for this test, it has been found that pre-splitting the table leads to greater stability and efficiency.

The updated code could look something like this:
```java
private void createTable(String t, boolean online) throws AccumuloSecurityException, AccumuloException, TableNotFoundException, TableExistsException {
    Connector conn = getConnector();
    SortedSet<Text> partitionKeys = new TreeSet<>(Arrays.asList(new Text("some split"));
    NewTableConfiguration ntc = new NewTableConfiguration().withSplits(partitionKeys);
    conn.tableOperations().create(t, ntc);
    conn.tableOperations().online(t, true);
    if (!online) {
      conn.tableOperations().offline(t, true);
    }
  }
```

milleruntime:
Thanks @DomGarguilo good idea.

EdColeman (approved):

milleruntime:
> Another small improvement may be to pre-split the table in the createTable method. I'm not too sure what is going on with the online/offline of the table, however if it is found that the split does not need to occur after the table is created for this test, it has been found that pre-splitting the table leads to greater stability and efficiency.
> 
> The updated code could look something like this:
> 
> ```java
> private void createTable(String t, boolean online) throws AccumuloSecurityException, AccumuloException, TableNotFoundException, TableExistsException {
>     Connector conn = getConnector();
>     SortedSet<Text> partitionKeys = new TreeSet<>(Arrays.asList(new Text("some split"));
>     NewTableConfiguration ntc = new NewTableConfiguration().withSplits(partitionKeys);
>     conn.tableOperations().create(t, ntc);
>     conn.tableOperations().online(t, true);
>     if (!online) {
>       conn.tableOperations().offline(t, true);
>     }
>   }
> ```
> 
> Not sure if this is something worth including, just thought I would point it out though.

This won't work in 1.10. The ability to pre-split tables wasn't added until 2.0. Ugh! I forgot about how crappy our API used to be. Be glad that you aren't cursed with that knowledge :grin:

DomGarguilo:
> This won't work in 1.10. 

Whoops. Maybe something to keep in mind if changes are made to the current version in that case.

PR #1944, opened 2021-02-22
Closes #1929. Add property verification for Sampler Options

### LINKED ISSUE #1929: Compaction from Shell hangs on unknown property
Initial bug report here: https://issues.apache.org/jira/browse/ACCUMULO-4486

**Describe the bug**
The compact command will cause the shall to hang if the user enters an incorrect property (see example in that issue).

**Versions:**
 I confirmed it still occurs on 2.1.0 and most likely still happens on every version still.

**To Reproduce**
There is a complete example in the JIRA issue that will reproduce the error. I will add the snipet here:
<details>
<pre>
createtable sampex
insert 9255 doc content 'abcde'
insert 9255 doc url file://foo.txt
insert 8934 doc content 'accumulo scales'
insert 8934 doc url file://accumulo_notes.txt
insert 2317 doc content 'milk, eggs, bread, parmigiano-reggiano'
insert 2317 doc url file://groceries/9.txt
insert 3900 doc content 'EC2 ate my homework'
insert 3900 doc uril file://final_project.txt
config -t sampex -s table.sampler=org.apache.accumulo.core.client.sample.RowSampler
config -t sampex -s table.sampler.opt.haser=murmur3_32
config -t sampex -s table.sampler.opt.hasher=murmur3_32
config -t sampex -s table.sampler.opt.modulus=3
compact -t sampex --sf-no-sample
</pre>
</details>

**Expected behavior**
For the shell to catch the `IllegalArgumentException` and properly handle the error so the user can regain control of the shell.

Manno15:
Closes #1929. 

Sampler options didn't have proper validity checks like other properties (it only made sure the prefix was valid but not the options part). This PR adds that verification so the case the issue specified cannot occur in the future.  It does stand to reason that we might want to adjust the shell API in the future so something cannot hang when an error occurs but with proper validity checks that might not be necessary.


core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1352,6 +1352,17 @@ public static boolean isValidZooPropertyKey(String key) {
...
+   * Checks if the given sampler option is valid
+   *
+   * @param option
+   *          property option
+   * @return true if option is a valid sampler option
+   */
+  public static boolean isValidSamplerOption(String option) {
+    return option.endsWith("hasher") || option.endsWith("modules");
+  }
+
```
ctubbsii, line 1365: Sampler options are arbitrary, and implementation-specific. The available options depend on the user's provided sampler implementation. They do not have to only be one of these options. These are merely example properties placed in the docs that an implementation might have.

Manno15: That is what I was afraid of. Guess there isn't an easy way to guarantee validity.

ctubbsii:
Another way you might provide early validation of these properties is to add some sort of options validation to samplers, either by adding a validateOptions method or similar to the interface as a default method, and then call it when validating the config.

Or, the hanging issue could be addressed by ensuring any exception thrown in the sampler's init method will trigger an error in the logs and skip the sampler, rather than trigger a retry behavior that hangs.

Manno15:
> Another way you might provide early validation of these properties is to add some sort of options validation to samplers, either by adding a validateOptions method or similar to the interface as a default method, and then call it when validating the config. Or, the hanging issue could be addressed by ensuring any exception thrown in the sampler's init method will trigger an error in the logs and skip the sampler, rather than trigger a retry behavior that hangs.

Do you know if users have to implement `AbstractHashSampler` when creating a new sampler? Most of the already configured samplers have validity checks already. Which is where the original error came from. Seems the best way to tackle this is to handle the hanging issue.

Manno15:
@ctubbsii I went ahead and added the validateOptions in the sampler interface. With what I have now, it will log an error message, and the sampler returns `null` so the rest of the compaction can complete as normal. The shell thread now returns and does not hang.  However, the shell still doesn't see the error so the user will have to realize that their sampler options didn't go through and make those corrections by looking at the monitor/logs. I marked this as a draft since I am still working on some things to hopefully improve functionality but I wanted to get the base idea out there for you to look at when you get the chance. Thanks!

core/src/main/java/org/apache/accumulo/core/client/sample/AbstractHashSampler.java
```diff
@@ -77,13 +74,6 @@ public void init(SamplerConfiguration config) {
     String hasherOpt = config.getOptions().get("hasher");
     String modulusOpt = config.getOptions().get("modulus");
 
-    requireNonNull(hasherOpt, "Hasher not specified");
-    requireNonNull(modulusOpt, "Modulus not specified");
```
ctubbsii, line 81: This was removed, but was not replaced with any kind of validation that the required options were set with actual values.

ctubbsii (changes requested):

core/src/main/java/org/apache/accumulo/core/client/sample/AbstractHashSampler.java
Manno15: My recent commit moved the null check to where the other argument is checked.

core/src/main/java/org/apache/accumulo/core/client/sample/Sampler.java
```diff
@@ -61,4 +61,11 @@
    *         Return false if it should not be included.
    */
   boolean accept(Key k);
+
+  /**
+   * @param option
+   *          Option to validate.
+   * @return True if option is a valid sampler option. False otherwise.
+   */
+  boolean isValidOption(String option);
```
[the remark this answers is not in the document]
Manno15: Good idea, I will look into adding that improvement.

core/src/main/java/org/apache/accumulo/core/client/sample/AbstractHashSampler.java
ctubbsii: Not quite. What you added was a check that the key was not null, which might be redundant, since the fact that it exists implies it is not null. This code here was checking that the values (not the keys) that correspond to these option keys was not null.

Manno15: Ah, you are correct. I misread that. My future changes should handle this.

core/src/main/java/org/apache/accumulo/core/client/sample/Sampler.java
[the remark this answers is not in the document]
Manno15: Is there anything we should do for the default case? I currently have it verified to not be null. Not sure what else to add to the interface itself.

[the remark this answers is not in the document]
ctubbsii: The default implementation should probably be a no-op. It is assumed that the map of all the options will be non-null, because Accumulo would be responsible for preparing that map itself, but it's fine if you want to check for that, to be sure.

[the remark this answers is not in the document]
Manno15: I went ahead and kept the check. If the user forgets to even mention the option before compacting, it will cause the shell to hang.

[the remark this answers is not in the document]
Manno15: Nevermind. You are correct. I just realized the null gets thrown before reaching my validation check.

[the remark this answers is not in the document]
Manno15: Made these changes in my recent commit.

Manno15:
Marking this back to ready for review.

core/src/main/java/org/apache/accumulo/core/sample/impl/SamplerFactory.java
```diff
@@ -38,13 +42,16 @@ public static Sampler newSampler(SamplerConfigurationImpl config, AccumuloConfig
...
+      sampler.validateOptions(config.getOptions());
       sampler.init(config.toSamplerConfiguration());
 
       return sampler;
 
     } catch (ReflectiveOperationException e) {
       throw new RuntimeException(e);
+    } catch (IllegalArgumentException | NullPointerException e) {
+      log.error("Cannot init sampler, invalid configuration: {}", e.getMessage(), e);
+      return null;
```
ctubbsii, line 54: It's a little weird that if the class can't be loaded, then it throws a runtime exception, but if it can't be started because of invalid config or some other problem, then it logs and returns null. I'm thinking these should probably be consistent. I would do something like:

```suggestion
    } catch (ReflectiveOperationException | RuntimeException e) {
      log.error("Cannot initialize sampler {}: {}", config.getClassName(), e.getMessage(), e);
      return null;
```

core/src/main/java/org/apache/accumulo/core/client/sample/RowColumnSampler.java
```diff
@@ -80,8 +83,11 @@ private boolean hashField(SamplerConfiguration config, String field) {
   }
 
   @Override
-  protected boolean isValidOption(String option) {
-    return super.isValidOption(option) || VALID_OPTIONS.contains(option);
+  public void validateOptions(Map<String,String> config) {
+    super.validateOptions(config);
+    for (Map.Entry<String,String> entry : config.entrySet()) {
```
ctubbsii, line 88: Since you're only using the key, and not validating any values in this implementation, you could just loop over the `keySet()` rather than the `entrySet()`.

core/src/main/java/org/apache/accumulo/core/client/sample/AbstractHashSampler.java
```diff
@@ -57,14 +58,28 @@
...
-  protected boolean isValidOption(String option) {
-    return VALID_OPTIONS.contains(option);
+      if (entry.getKey().equals("modulus"))
+        checkArgument(Integer.parseInt(entry.getValue()) > 0,
+            "Improper Integer value for modulus: %s", entry.getValue());
+    }
+  }
+
+  protected boolean isValid(String entry) {
+    return VALID_OPTIONS.contains(entry) || VALID_VALUES_HASHER.contains(entry);
```
ctubbsii, line 82: Each half of this method's implementation is checking for different things. The left side of the `||` checks keys, whereas the right side checks values. Each case should probably just be inlined into their respective areas of the code, and this method can be deleted.

ctubbsii: Or, since this method is protected (public API), it should be deprecated, and its original implementation left intact.

ctubbsii (changes requested):

core/src/main/java/org/apache/accumulo/core/client/sample/RowColumnSampler.java
Manno15: Good catch

core/src/main/java/org/apache/accumulo/core/sample/impl/SamplerFactory.java
Manno15: Good idea, I like this change.

core/src/main/java/org/apache/accumulo/core/client/sample/AbstractHashSampler.java
Manno15: My lasted commit includes these changes. I did revert that function and deprecated it. Let me know if I need to do something else for it or not.

ctubbsii (approved):

core/src/main/java/org/apache/accumulo/core/client/sample/AbstractHashSampler.java
```diff
@@ -57,12 +58,33 @@
...
+      if (entry.getKey().equals("modulus"))
+        checkArgument(Integer.parseInt(entry.getValue()) > 0,
+            "Improper Integer value for modulus: %s", entry.getValue());
+    }
+  }
 
   /**
    * Subclasses with options should override this method and return true if the option is valid for
    * the subclass or if {@code super.isValidOption(opt)} returns true.
+   * @deprecated since 2.1.0, replaced by {@link #validateOptions(Map)}
```
ctubbsii, line 85:
```suggestion
   *
   * @deprecated since 2.1.0, replaced by {@link #validateOptions(Map)}
```

Manno15: That is what caused the other build error. I couldn't win with either.

Manno15: I guess it was something on my end to create extra white space. I had what you committed before and the build failed.

ctubbsii: When the formatter automatically adds a line, it leaves a space after the `*` to indent the section. That causes checkstyle to complain. However, if you delete that extra space, the formatter is fine with it and will not reformat it to add it again. My change added the line back without the extra space that caused it to fail previously. It's just an annoying quirk of the formatter.

ctubbsii: I actually think it might be useful to add the trailing space removal as a feature to https://github.com/revelc/formatter-maven-plugin since the Eclipse formatter library it uses doesn't have a feature to remove the trailing whitespace itself.

Manno15: > When the formatter automatically adds a line, it leaves a space after the * to indent the section. That causes checkstyle to complain. However, if you delete that extra space, the formatter is fine with it and will not reformat it to add it again. My change added the line back without the extra space that caused it to fail previously. It's just an annoying quirk of the formatted.

ah, that makes sense. Thanks for fixing it.

Issue #1451, opened 2019-12-06
Support external compactions in containers


keith-turner:
For use cases like large scale filtering data on an accumulo table, it may be useful to support running compactions externally from a tserver in a system like kubernetes. This feature could support the following operations and behaviors.

 * Client side Accumulo API that selects tablets and files to compact that returns serializable+runnable objects for each external compaction. The objects can be serialized and run anywhere that has access to DFS.
 * A lease for every file that has been selected for external compaction.  This lease prevents other compactions from processing the files.
 * A client side API for listing, committing, and canceling outstanding external compactions. 

If all of the selection decisions for tablets and files to compact are made in the client side, then a user could pass a lambda to Accumulo to make these decisions.  This approach would avoid having to put that code on tservers.  The Accumulo client code could make RPCs to bring all needed information to the client side. Accumulo could also automatically handle the case of things changing and call the lambda again.

External compactions could have early or late binding for the set of files to compact.  Early binding is much easier to implement and run idempotently on a cluster.  There is one problem with early binding : leases could be held on lots of small file for a long time negatively impacting scans.  One possible way to avoid this would be compact all files less than size X on tserver before starting an external compaction and then only select files over size X for the external compaction.

EdColeman:
Concerning leases and lease management.  The apache river / jini project has a lease manager specification that may provide some guidance with this type of functionality.

See https://river.apache.org/release-doc/current/specs/html/leaseutil-spec.html

ivakegg:
A few comments:
1) I would imagine that this is primarily for handling larger files that may not be compacted using the configured compaction manager.
2) Using something like yarn instead of kubernetes might be a good initial implementation.
3) I would like to add the ability to do this while accumulo is off line.  This might be very difficult since we still need to make modifications to the accumulo metadata but there might be a way using fate operations, conditional mutations, zookeeper, other....

ivakegg:
Now that I think about it, doing this while accumulo is off line might not be worth the effort.  Doing this for an off-line table may be reasonable however.

keith-turner:
> Using something like yarn instead of kubernetes might be a good initial implementation.

The Accumulo API's will return a list of serializeable runnables.  Then a user could run those wherever they want.  This could lead to multiple little bridge projects that help run compactions in different execution environments.  This is an approach we took with Fluo when creating [fluo-yarn](https://github.com/apache/fluo-yarn) and [fluo-docker](https://github.com/apache/fluo-docker).

[...]

PR #1946, opened 2021-02-23
re #1451: Initial external compaction thrift APIs

dlmarion:
Initial work on the external compaction thrift API (sans the tablet server changes). This should allow me to start working on the Coordinator and Compactor processes.


core/src/main/thrift/compaction-coordinator.thrift
```diff
@@ -0,0 +1,134 @@
...
+}
+
+service CompactionCoordinator {
+
+  /*
+   * Called by TabletServer (or CLI) to add, update, or cancel a compaction for a tablet
+   */
+  void AddOrUpdateCompaction(
+    1:string tableId
+    2:string endRow
```
keith-turner, line 78: Will probably end up needed to use TKeyExtent so that we can distinguish tablets that have split or merged.

core/src/main/thrift/compaction-coordinator.thrift
```diff
@@ -0,0 +1,134 @@
...
+    3:string queueName
+    4:i64 priority
+    5:Action action
+  )
+  
+  /*
+   * Called by TabletServer (or CLI) to get current status of compaction for a tablet
+   */
+  Status getCompactionStatus(
```
keith-turner, line 87: How do you envision this being used?

core/src/main/thrift/compaction-coordinator.thrift
```diff
@@ -0,0 +1,134 @@
...
+    1:string tableId
+    2:string endRow
+    3:string queueName
+  )
+  
+  /*
+   * Called by TabletServer (or CLI) to get complete status of compaction for a tablet
+   */
+  list<Status> getCompactionStatusHistory(
```
keith-turner, line 96: How do you envision this being used?

core/src/main/thrift/compaction-coordinator.thrift
```diff
@@ -0,0 +1,134 @@
...
+  # Compactor should set state to CANCELLED to acknowledge that it has stopped compacting 
+  CANCELLED
+}
+
+struct CompactionJob {
+  1:trace.TInfo traceInfo
+  2:security.TCredentials credentials
+  3:i64 compactionId
+  5:data.TKeyExtent extent
+  6:list<string> files
```
keith-turner, line 53: This will also need an output file.

keith-turner (commented):
Think the tserver thrift idl will need two methods added.  One to reserve a compaction and get a CompactionJob and another to commit a compaction.  Maybe something like

```
   CompactionJob getCompactionJob(TKeyExtent extent, string queuename, compactorId cid)
  void commitCompaction(CompactionJob job)
```

core/src/main/thrift/compaction-coordinator.thrift
dlmarion: I think there might be two cases, which might end up being the same:

1. The compaction manager is constantly checking to see if the tablets need to be compacted. We had talked about, for external compactions, having the tserver just send a compaction request to the coordinator and having the coordinator just ignore the request is one is already queued. An alternate approach would be for the tserver to have (or the ability to know) the if a compaction is occurring or what the state is.
2. Likewise, if a compaction for a tablet is started, and then the tablet is re-hosted, giving the ability to the tablet server to determine if a compaction is occurring or queued for that tablet.

core/src/main/thrift/compaction-coordinator.thrift
dlmarion: I'll add.

core/src/main/thrift/compaction-coordinator.thrift
dlmarion: I can modify, I assume that this file might change several times as we work through issues.

core/src/main/thrift/compaction-coordinator.thrift
dlmarion: I was thinking that the Compactor would call Coordiator.updateCompactionState as the compaction is started, in progress, and comleted (or failed). This method, if we stored the state history, would provide that history of the state changes. It could potentially be used for finding out how far along a tablet is in the compaction process and what host is it running on, or if it's not making progress.

core/src/main/thrift/compaction-coordinator.thrift
keith-turner: >  giving the ability to the tablet server to determine if a compaction is occurring or queued for that tablet.

I Was thinking the coordinator would maintain the state of the queue and tablest would maintain the state of external running compactions.  Tablets on tservers will constantly need this information for planning purposes. Was thinking tablets would track running external compactions in their portion of the metadata table.  When an external compaction is committed they can atomically change the files in the metadata table and remove the information about the external compaction.  When a tablet loads on a new tserver it always reads it metadata and would see information about running compactions.

keith-turner (approved):
We can work out changes as we go.

PR #1962, opened 2021-03-08
closes #1955 - Move log message

### LINKED ISSUE #1955: Log spam while running ZooLockIT
**Describe the bug**
While running ZooLockIT, there is a debug log that gets logged several hundred times per second. This issue was uncovered while working on #1952. Initial discussion suggested this may be a problem rather than a case of incorrect log type(debug vs. trace). Example of the log:

```java
2021-03-03T10:45:13,618 [zookeeper.ZooLock] DEBUG: [zlock#3db3b47d-4fc2-4b38-83cd-c733880121bf#] Renewing watch on prior node  /zltestDeleteWaiting-587915658-l1/zlock#947ce341-f843-4742-8b28-79f2de9cc7da#0000000000
2021-03-03T10:45:13,618 [zookeeper.ZooLock] DEBUG: [zlock#3db3b47d-4fc2-4b38-83cd-c733880121bf#] Renewing watch on prior node  /zltestDeleteWaiting-587915658-l1/zlock#947ce341-f843-4742-8b28-79f2de9cc7da#0000000000
2021-03-03T10:45:13,619 [zookeeper.ZooLock] DEBUG: [zlock#3db3b47d-4fc2-4b38-83cd-c733880121bf#] Renewing watch on prior node  /zltestDeleteWaiting-587915658-l1/zlock#947ce341-f843-4742-8b28-79f2de9cc7da#0000000000
2021-03-03T10:45:13,619 [zookeeper.ZooLock] DEBUG: [zlock#3db3b47d-4fc2-4b38-83cd-c733880121bf#] Renewing watch on prior node  /zltestDeleteWaiting-587915658-l1/zlock#947ce341-f843-4742-8b28-79f2de9cc7da#0000000000
```

**To Reproduce**
This can be reproduced by running ZooLockIT and viewing the logs.

dlmarion:


ctubbsii (approved):
Code change looks good. However, git log subject line is a bit long. Please take a look at [these tips](https://lists.apache.org/thread.html/r5773a166ef2ada07a9ad769af0e1b729614312e763aac80961d5051a%40%3Cdev.accumulo.apache.org%3E) before you merge.

EdColeman:
@dlmarion - thinking about this, there is only one edge case where a sequential, ephemeral node would ever find a node with a lower sequential number that was created "after" the current node.  The sequential numbers are assigned based on the parent, so each child node is created with a 1-up number.  For a lowered numbered node to wink into existence it would take something like, (and hopefully zookeeper prevents this):
 
```
process A creates a sequential node - this reserves the number, but the node is not "created" yet.
process B creates a sequential node - this reserves the "next, higher" number`, the node IS created.
time passes...
the node created by A comes into existence.
```
If zookeeper protects against this by atomically assigning a sequential number and creating the node, then if you create a sequential node, sequential nodes with lower numbers either exist - or if they don't they can never be recreated using the zookeeper sequential API. Any subsequent sequential nodes created will always have a higher sequential number, never lower.

The sequential numbers always increasing is baked into zookeeper and fundamental to the leader election recipe.  So, if you have a sequential node and look for nodes with lower numbers, they either exist or they never will.  If you checked for existence and then added the watcher the only edge case would be if the node was deleted between the check and the watch add - in which case we'd probably want to remove the watcher.

dlmarion:
Looking at the [note](https://zookeeper.apache.org/doc/r3.6.2/zookeeperProgrammers.html#ch_zkGuarantees) in the `Consistency Guarantees` section of the ZK programmers guide,  I'm confused by the sentence `ZooKeeper does not guarantee that at every instance in time, two different clients will have identical views of ZooKeeper data.`. Does it mean `data` in the general sense, or `data` in a node?

If it's the latter, then ZooLock may need to perform a `sync` before getting the list of children to ensure that it gets a complete list from the ZK server that it's connected to. In your example above, the node created by process A was created but the ZK server to which process B is connected may not know about it?

dlmarion:
There is no mention of a client having to use `sync` in the Locks or Leader Election section of [Recipes](https://zookeeper.apache.org/doc/r3.6.2/recipes.html). So, maybe it's just the data in the node.

PR #1980, opened 2021-03-30
Add created column to Fate Print command

Manno15:
From https://issues.apache.org/jira/browse/ACCUMULO-4161. An age/created column was added to the `fate print ` command in the shell. 

Image below is what the timestamp looks like in its current state. 

image

### DIFF
--- a/core/src/main/java/org/apache/accumulo/fate/AdminUtil.java
+++ b/core/src/main/java/org/apache/accumulo/fate/AdminUtil.java
@@ -80,10 +80,10 @@ public class AdminUtil<T> {
     private final List<String> hlocks;
     private final List<String> wlocks;
     private final String top;
-    private final String timeTopCreated;
+    private final String timeCreated;
 
     private TransactionStatus(Long tid, TStatus status, String debug, List<String> hlocks,
-        List<String> wlocks, String top, String timeTopCreated) {
+        List<String> wlocks, String top, String timeCreated) {
 
       this.txid = tid;
       this.status = status;
@@ -91,7 +91,7 @@ public class AdminUtil<T> {
       this.hlocks = Collections.unmodifiableList(hlocks);
       this.wlocks = Collections.unmodifiableList(wlocks);
       this.top = top;
-      this.timeTopCreated = timeTopCreated;
+      this.timeCreated = timeCreated;
 
     }
 
@@ -138,8 +138,8 @@ public class AdminUtil<T> {
     /**
      * @return The timestamp of when the operation was created.
      */
-    public String getTimeTopCreated() {
-      return timeTopCreated;
+    public String getTimeCreated() {
+      return timeCreated;
     }
   }
 
@@ -382,7 +382,7 @@ public class AdminUtil<T> {
 
       TStatus status = zs.getStatus(tid);
 
-      String timeTopCreated = zs.timeTopCreated(tid);
+      String timeCreated = zs.timeCreated(tid);
 
       zs.unreserve(tid, 0);
 
@@ -390,7 +390,7 @@ public class AdminUtil<T> {
           || (filterStatus != null && !filterStatus.contains(status)))
         continue;
 
-      statuses.add(new TransactionStatus(tid, status, debug, hlocks, wlocks, top, timeTopCreated));
+      statuses.add(new TransactionStatus(tid, status, debug, hlocks, wlocks, top, timeCreated));
     }
 
     return new FateStatus(statuses, heldLocks, waitingLocks);
@@ -412,7 +412,7 @@ public class AdminUtil<T> {
       fmt.format(
           "txid: %s  status: %-18s  op: %-15s  locked: %-15s locking: %-15s top: %-15s created: %s%n",
           txStatus.getTxid(), txStatus.getStatus(), txStatus.getDebug(), txStatus.getHeldLocks(),
-          txStatus.getWaitingLocks(), txStatus.getTop(), txStatus.getTimeTopCreated());
+          txStatus.getWaitingLocks(), txStatus.getTop(), txStatus.getTimeCreated());
     }
     fmt.format(" %s transactions", fateStatus.getTransactions().size());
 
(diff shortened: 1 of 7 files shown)


EdColeman:
From a quick look, some general things you may want to consider:

  - Format the printed time in UTC with ISO-8601 format instead of java Date default.
  - Provide more descriptive naming in functions instead of timestamp, maybe topCreated() or timeTopCreated()?  

The rational for a more descriptive function name than getTimestamp - there are multiple timestamps important in overall FATE transaction life cycle, some of which might be useful if additional FATE timing information was ever extended. If the function naming reflected the operation from an "external" perspective then it may be easier differentiate what time this is retrieving (FATE top creation) rather than some of the other possibilities.

Manno15:
Good ideas, especially on the naming convention. I wasn't sure on the time zone part which is why I stuck with the default. I will look into making that change.

core/src/main/java/org/apache/accumulo/fate/ZooStore.java
```diff
@@ -478,6 +480,18 @@ public Serializable getProperty(long tid, String prop) {
...
 
+  @Override
+  public String getTimestamp(long tid) {
+    verifyReserved(tid);
+
+    try {
+      Stat stat = zk.getZooKeeper().exists(getTXPath(tid), false);
+      return new Date(stat.getCtime()).toString();
+    } catch (Exception e) {
+      throw new RuntimeException(e);
```
EdColeman, line 491: Consider returning an empty string (or N/A, or unavailable?) instead of throwing an exception?  This is for information and continuing without this data point (especially if the exception was transient) may allow other operations to continue if they can.  What is returned versus what is displayed may want to consider a user wanting to parse the output (say with awk or sort) along what is clear to a user.

Manno15:
> Format the printed time in UTC with ISO-8601 format instead of java Date default.

The original ticket specified a more human-readable print statement. My latest commit does that. What are your thoughts on this @EdColeman? I did keep it in the same order as expect with iso_8601 (I believe). Picture below: 
image

core/src/main/java/org/apache/accumulo/fate/ZooStore.java
Manno15: I can see the value in this. I was trying to match the other functions in this class with the exception handling but I can look into this more.

core/src/main/java/org/apache/accumulo/fate/ZooStore.java
```diff
@@ -486,7 +488,8 @@ public String timeTopCreated(long tid) {
 
     try {
       Stat stat = zk.getZooKeeper().exists(getTXPath(tid), false);
-      return new Date(stat.getCtime()).toString();
+      DateTimeFormatter dtf = DateTimeFormatter.ofPattern("E yyyy MMM dd z hh:mm:ss");
```
EdColeman, line 491: iso time like 2021-03-31T10:00:00Z would be easier for external parsing / sorting

Manno15: That's fair. I will add it back in my next commit.

Manno15: image

core/src/main/java/org/apache/accumulo/fate/ZooStore.java
EdColeman: Specifically - if NoNodeExists could possibly be thrown (say it completed asynchronously while doing the gathering) then a blank instead of throwing an exception would allow other values to be displayed.

Manno15: I made this change in the latest commit. I went with "ERROR" but I am open to something else or a more specific message depending on what actual exception occurred.

core/src/main/java/org/apache/accumulo/fate/ReadOnlyTStore.java
```diff
@@ -139,4 +139,12 @@
...
   List<Long> list();
 
+  /**
+   * Retrieve timestamp of a transaction.
+   *
+   * @param tid
+   *          Transaction id, previously reserved.
+   * @return timestamp of transaction.
+   */
+  String timeTopCreated(long tid);
```
ctubbsii, line 149: I'm just confused on the `Top` portion of the name. What does that mean?

Manno15: Table operation, I believe. Some of the other functions in that class use that name. I do agree it could be confusing. `timeCreated` may just be better.

EdColeman: TOP is the operation that created the operation stack. I'm not sure if it is table operation or not, but FATE operations create a stack of operations and top represents the parent (I think)  This is coming from someone that took an embarrassing long time to realize that REPO stood for REPeatable Operation and not just generic repository. So...

Manno15: Oh, that makes sense. I will go ahead and change the function names to avoid any confusion in the future.

core/src/main/java/org/apache/accumulo/fate/AdminUtil.java
```diff
@@ -80,16 +80,18 @@ public AdminUtil(boolean exitOnError) {
...
-        List<String> wlocks, String top) {
+        List<String> wlocks, String top, String timeCreated) {
 
       this.txid = tid;
       this.status = status;
       this.debug = debug;
       this.hlocks = Collections.unmodifiableList(hlocks);
       this.wlocks = Collections.unmodifiableList(wlocks);
       this.top = top;
+      this.timeCreated = timeCreated;
```
EdColeman, line 94: Did you give any thought to storing / passing the timeCreated as a long?  That could make use of the transaction status easier for anything that wanted to access the timestamp for other purposes.  This would defer the formatting closer to where its needed and other uses would not need to parse the string.  Sorry that this didn't occur to me earlier.

Manno15: What should be in the catch block inside `ZooStore` then? Should I just return 0 or 1/-1 or revert back to throwing the runtimeException?

EdColeman: I guess 0?  The if > 0 the print can format the time otherwise ERROR

Manno15:
I went ahead and added a format function people can use and also one that just returns the long value for it.

EdColeman (approved):

PR #1981, opened 2021-03-30
WIP: Convert Hadoop Metrics to Micrometer 

### LINKED ISSUE #1134: Look into using abstraction for metrics.
There may be a chance to abstract the metrics implementation to allow plug-able sources and sinks.  One approach may be https://github.com/micrometer-metrics/micrometer.

First need to determine requirements and then see what is available / suitable candidates.

DomGarguilo:
Fixes #1134

Converting from the use of hadoop metrics to micrometer metrics.

### DIFF
--- a/server/base/pom.xml
+++ b/server/base/pom.xml
@@ -57,8 +57,18 @@
       <artifactId>commons-codec</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-core</artifactId>
+      <version>1.6.4</version>
+    </dependency>
+    <dependency>
       <groupId>org.apache.accumulo</groupId>
       <artifactId>accumulo-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.accumulo</groupId>
+      <artifactId>accumulo-iterator-test-harness</artifactId>
+      <version>2.1.0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>org.apache.accumulo</groupId>
--- a/server/base/pom.xml
+++ b/server/base/pom.xml
@@ -67,11 +67,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.accumulo</groupId>
-      <artifactId>accumulo-iterator-test-harness</artifactId>
-      <version>2.1.0-SNAPSHOT</version>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.accumulo</groupId>
       <artifactId>accumulo-start</artifactId>
     </dependency>
     <dependency>
--- a/server/gc/src/main/java/org/apache/accumulo/gc/metrics/GcMetrics.java
+++ b/server/gc/src/main/java/org/apache/accumulo/gc/metrics/GcMetrics.java
@@ -44,7 +44,7 @@ import io.micrometer.core.instrument.logging.LoggingRegistryConfig;
 public class GcMetrics extends Metrics {
 
   // use common prefix, different that just gc, to prevent confusion with jvm gc metrics.
-  public static final String GC_METRIC_PREFIX = "acc.gc.";
+  public static final String GC_METRIC_PREFIX = "AccGc";
 
   private static final String jmxName = "GarbageCollector";
   private static final String description = "Accumulo garbage collection metrics";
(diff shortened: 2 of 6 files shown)


server/base/pom.xml
```diff
@@ -56,10 +56,20 @@
...
+    </dependency>
     <dependency>
       <groupId>org.apache.accumulo</groupId>
       <artifactId>accumulo-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.accumulo</groupId>
+      <artifactId>accumulo-iterator-test-harness</artifactId>
+      <version>2.1.0-SNAPSHOT</version>
+    </dependency>
```
EdColeman, line 72: Why include the test-harness? This seems like an undesirable dependency,

server/manager/src/main/java/org/apache/accumulo/manager/metrics/fate/FateMetrics.java
```diff
@@ -29,12 +29,17 @@
...
+// import org.apache.hadoop.metrics2.lib.MetricsRegistry;
+// import org.apache.hadoop.metrics2.lib.MutableGaugeLong;
 import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
```
EdColeman, line 41: For an example, this may be okay - but for production I think I'd expect either a CompositeMeterRegisry that was created / managed by the metrics context - or would use the Metrics.globalRegisrty.

DomGarguilo: I agree. This is one of the things I was hoping to get input on. Both seem like they could work however I'm not sure which would be best.

server/base/pom.xml
DomGarguilo: This must have been added automatically somehow. I'm not sure why.

server/manager/src/main/java/org/apache/accumulo/manager/metrics/fate/FateMetrics.java
EdColeman: With the current initialization occurring in a base class, the composite seems the likely candidate, but global may be just as well - it may depend on / if there are multiple tenant situations - say in a cluster all things run in a separate process, then there would be no conflicts, but if say in mini-cluster (or another situation) where the registry was shared by multiple "processes" then there may need to be a way to differentiate between them - so that if there were metrics named the same, but for different processes they would not step on each other.

With the global registry, this could be mitigated by include a process name tag (or something)

In general, a process tag might already be baked into the current naming scheme so it would be a non-issue just to carry that forward.  There may be guidance in naming conventions and what makes a suitable tag.

server/manager/src/main/java/org/apache/accumulo/manager/metrics/fate/FateMetrics.java
```diff
@@ -141,46 +197,51 @@ protected void prepareMetrics() {
...
+     * super.getRegistry().newGauge(metricNameHelper(FATE_TX_STATE_METRIC_PREFIX, key),
+     * "By transaction state count for " + key, value)) .set(value));
+     */
     // the op types are dynamic and the metric gauges generated when first seen. After
     // that the values need to be cleared and set to any new values present. This is so
     // that the metrics system will report "known" values once seen. In operation, the
     // number of types will be a fairly small set and should populate a consistent set
     // with normal operations.
 
+    // DONT NEED THIS SINCE MICROMETER GAUGE DOESNT NEED TO BE SET
```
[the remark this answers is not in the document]
DomGarguilo: I think I understand, I'm not sure this same approach will work with micrometer gauges since the value will just be recomputed each time the gauge is read so even if we set it to 0, 0 will not be returned once we request the gauges value.

I'll have to look into a solution for this.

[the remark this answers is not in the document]
EdColeman: Using the example from their docs (https://micrometer.io/docs/concepts#_gauges) one approach my be to create AtomicIntegers when first seen, store them in the map, register those with registry.gauge(key, atomic_value_ref) - then the metrics updates would 1) set each atomic_value_ref in the map to 0, then process the current counts and set the current values.  Creating the AtomicIntegers and storing them in a map (keyed by opname) would also satisfy the need to keep a strong reference to the values.

This assumes that something like the following would be valid

```
AtomicInteger x = new AtomicInteger(0);
...
registry.gauge(key, x);
...
x.set(value)

(x would be stored in the map, keyed with optype name)
```

works and you don't need the reference returned from registry.gauge(....)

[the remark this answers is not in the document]
DomGarguilo: @EdColeman, I did my best to address this in 85f71d3.

server/gc/src/main/java/org/apache/accumulo/gc/metrics/GcMetrics.java
```diff
@@ -35,7 +44,7 @@
 public class GcMetrics extends Metrics {
 
   // use common prefix, different that just gc, to prevent confusion with jvm gc metrics.
-  public static final String GC_METRIC_PREFIX = "AccGc";
```
dlmarion, line 38: I think we want to be careful with modifying the metric names being emitted unless there is good reason. A change like this would cause existing consumers of the metrics to have to change all of their monitoring and alert definitions.

DomGarguilo: Good point. I changed it to match micrometers' dot separated naming convention.

dlmarion: If we have to change it to make it work, that is one thing. If we don't have to change it, then we probably shouldn't.

DomGarguilo: That makes sense. I'll have it changed back in the next commit.

DomGarguilo:
Superseded by #2188.

PR #1982, opened 2021-03-30
Cleanup Tests

milleruntime:
* Add Test annotations to unused methods that contain test code
* Drop unused helper methods and declarations in various Tests


Manno15 (approved):
Everything looks good to me from a quick look over.

PR #2005, opened 2021-04-07
Update lastLocation on assignment in 1.10 branch

### LINKED ISSUE #1169: balancing strategy

ivakegg:
* Provides for faster initial accumulo by ensuring the last location is always up to date.

This is a backport of #1453 (re #1169)

### DIFF
--- a/server/base/src/main/java/org/apache/accumulo/server/master/state/MetaDataStateStore.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/master/state/MetaDataStateStore.java
@@ -74,6 +74,7 @@ public class MetaDataStateStore extends TabletStateStore {
       for (Assignment assignment : assignments) {
         Mutation m = new Mutation(assignment.tablet.getMetadataEntry());
         assignment.server.putLocation(m);
+        assignment.server.putLastLocation(m);
         assignment.server.clearFutureLocation(m);
         SuspendingTServer.clearSuspension(m);
         writer.addMutation(m);


ctubbsii (changes requested):
This doesn't achieve what you are probably expecting it to achieve. The last location field is *not* the "last assigned" location. It is the "last location data was written locally" location. It should *not* be updated on assignment, but when files are written. The "location" field is already updated to reflect the last location assigned. If you need assignments based on last location assigned, I suggest using the normal "location" field in a custom balancer, rather than the "last location" field.

This misunderstanding actually resulted in ultimately reverting the issues you are referencing. It should not be backported, since it is not the correct fix.

Also, since most people do not have ticket numbers memorized, please ensure the subject line of your git commit messages and subject line of the pull request or issue has a brief description of the issue, rather than merely references to ticket numbers. I will update the github PR description for this, but ultimately, I think this issue should be closed, since it doesn't achieve what it appears to, and the issue it is based on was actually reverted.

PR #2008, opened 2021-04-09
Drop duplicate constructor in ServiceLock

milleruntime:
* Remove constructor in ServiceLock and make all services use the
constructor that takes a ZooKeeper object
* Get the authenticatedSession ZK from ServerContext instead of passing
in conf and parsing properties everytime.
* Simplify method in ServiceLockIT

### DIFF
--- a/core/src/main/java/org/apache/accumulo/fate/zookeeper/ServiceLock.java
+++ b/core/src/main/java/org/apache/accumulo/fate/zookeeper/ServiceLock.java
@@ -25,8 +25,6 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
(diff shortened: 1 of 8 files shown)


Manno15:
Was the reasoning behind this change discussed elsewhere? The changes look good, just looking for more insight.

milleruntime:
> Was the reasoning behind this change discussed elsewhere? The changes look good, just looking for more insight.

Nope. I was examining the code while trying to decipher another bug and noticed the two similar constructors. Only having one constructor makes the class easier to follow everywhere it's created (like using find usage in an IDE) and is less code. The other improvements were just bonuses.

ctubbsii (approved):

milleruntime:
The unit-tests pass locally. For some reason GitHub keeps timing out our second QA check.

PR #2014, opened 2021-04-13
Initialize hdfs connection with correct ugi for SetGoalState

sjyang18:
This is the same problem fixed for managers and tservers to use the keytab file for kerberos authentication previously (https://github.com/apache/accumulo/pull/1727/commits/b6a9ad000d261f201b6322031d34c60fcbbb9d5a).   **serverLogin** should be called before **ServerContext** is created in order to initialize hdfs connection with correct ugi for **SetGoalState**


ctubbsii (approved):
Same fix as in #1727

PR #2018, opened 2021-04-14
Create new ZKDroppingTServer

milleruntime:
Putting this class here because it extends TServer and the constructor is package private. Required to run 
https://github.com/apache/accumulo-testing/pull/139


dlmarion:
FWIW I'm working on a test right now that extends TServer. I opted to add the `protected` modifier to the constructor and placed the class in the test package.

milleruntime:
> FWIW I'm working on a test right now that extends TServer. I opted to add the `protected` modifier to the constructor and placed the class in the test package.

OK cool. I wasn't sure if it was appropriate it or not. But it looks like being able to extend it for testing is very useful.

milleruntime:
I will close this since it seems making the constructor protected is a better move.

Issue #2030, opened 2021-04-26
Make delete table operations cancel user compactions


keith-turner:
**Is your feature request related to a problem? Please describe.**
User compactions start a FATE op that gets a table lock.  Delete table operations wait to get the table lock.  They may wait on a user compaction.  If they do, it would be nice if they didn't.

**Describe the solution you'd like**
There is already an existing cancel user compaction API call.  It may be that a delete table operation could call this (client side) prior to initiating a delete table FATE op. 

**Describe alternatives you've considered**
A more general alternative may be to add canceling FATE ops as a high level FATE operation that could be called before delete table to cancel any FATE op running against that table (like merge).

milleruntime:
I thought this is already done? https://github.com/apache/accumulo/blob/e6cc640216e22075c205efeb6ba784a5af720c60/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Compactor.java#L391-L403

dlmarion:
So, in one of the external compaction tests that I wrote the delete command was waiting to get the FATE lock which was held by the user compaction.

milleruntime:
I did see the CompactionCanceledException thrown in 1.10 while testing #1872. But it looks like the Delete doesn't take any action to cancel the compaction. I guess it depends on where in the compaction process the Delete occurs as to whether the compaction will see the table deleted before it kicks off. I created an IT to test a typical internal majc and saw the compaction hang, timing out the test #2038

dlmarion:
I wonder if [CompactionCheck ](https://github.com/apache/accumulo/blob/main/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java#L524) should return `false` if `Tablet.isBeingDeleted()` is `true`. It already returns `false` if the tablet is closed.

dlmarion:
So, that didn't work. Looking at the Manager logs I see a constant stream of:
```
2021-06-01T14:35:54,613 [tableOps.Utils] INFO : namespace +default (55e4a47017a35a48) locked for read operation: DELETE
```
I noticed that this is the success case when the getLock method is successful. I added some logging to the `else` case, when the lock is not successful and the method returns `100`. It appears that the DELETE fate operation cannot get the table lock. The delete call in your test is waiting until the compaction is complete before executing.

dlmarion:
Here's the snippet from the manager log (with my log changes):
```
[...]

PR #2038, opened 2021-04-27
WIP test for #2030

milleruntime:


test/src/main/java/org/apache/accumulo/test/DeleteCancelsMajCIT.java
```diff
@@ -0,0 +1,112 @@
...
+      }
+      // delete the table, interrupts the compaction
+      log.info("Delete table: " + tableName);
+      c.tableOperations().delete(tableName);
+
+      while (!finished.get()) {
+        log.info("Wait for compaction to be cancelled");
+        sleepUninterruptibly(5, TimeUnit.SECONDS);
+      }
+      assertTrue(finished.get());
```
dlmarion, line 109: Shouldn't this be false and ex.get() != null if the compaction was interrupted by the delete?

milleruntime: Possibly. I am not sure what will happen with the client yet as we haven't implemented the cancel.

dlmarion:
@milleruntime - I think this can be closed. I resolved #2030 in a different PR.

PR #2039, opened 2021-04-27
Add bulk import option to ignore empty dirs

jmark99:
Update the new bulk import operations to allow empty source directories to be ignored rather than throwing an exception. There is now a '-i' option that will log an info message indicating that the source directory was empty and no files were imported rather than throwing an IllegalArgumentException.

This will support the use case of users who want to write idempotent bulk import code.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
@@ -802,6 +802,32 @@ public interface TableOperations {
    * @since 2.0.0
    */
   default ImportDestinationArguments importDirectory(String directory) {
+    return importDirectory(directory, false);
+  }
+
+  /**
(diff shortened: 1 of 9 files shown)


EdColeman (approved):
I didn't try to run this, but visually the changes look good.

milleruntime:
@jmark99 I think it might be better to add a method to the fluent API here: https://github.com/apache/accumulo/blob/5c365f984b13dec105ecb654cea486a06443d967/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java#L731
Perhaps a method called `ignoreEmptyDir()`. This will allow it to be an optional method and won't require a entry point method on `TableOperations`

keith-turner:
The ImportOptions interface may be a good place to put it.

jmark99:
I will look into updating to take advantage of the fluent api.

jmark99:
The update has been submitted as PR #2045

PR #2045, opened 2021-04-28
Move ignoreEmptyDir opt to ImportOptions interface

jmark99:
Update the importdirectory ignoreEmptyDir option to use the bulk import fluent API. The ignoreEmptyDir option is moved into the ImportOptions interface and can be set during the call to the load method.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
@@ -718,6 +718,8 @@ public interface TableOperations {
 
     /**
      * Ignores empty bulk import source directory, rather than throwing an IllegalArgumentException.
+     *
+     *  @since 2.1.0
      */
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/apache/accumulo/core/client/admin/TableOperations.java
```diff
@@ -716,6 +716,11 @@ void importDirectory(String tableName, String dir, String failureDir, boolean se
      */
     ImportMappingOptions tableTime(boolean value);
 
+    /**
+     * Ignores empty bulk import source directory, rather than throwing an IllegalArgumentException.
+     */
+    ImportMappingOptions ignoreEmptyDir(boolean ignore);
```
milleruntime, line 722: Should include `@since 2.1.0` java doc tag since this will be new.

jmark99: Tag was added.

milleruntime (approved):

PR #2067, opened 2021-05-01
Declaration of map in MapCounter as final

Amisha328:


EdColeman (changes requested):
The PR should not contain the .class files - if you build with maven, those files would go to ./target directory, which is excluded.  Please delete those two .class files from the PR.

EdColeman:
Superseded by #2071

PR #2074, opened 2021-05-03
Fix for Constraint API

milleruntime:
* Add method to deprecated API to prevent errors
* Add test to DeprecatedConstraintExtendTest

This should fix ShellServerIT


ctubbsii (approved):
LGTM if it fixes the test.

milleruntime:
@EdColeman I think this is all that needs to be done for the latest failures in ShellServerIT, it passes locally for me with this fix. But I am not sure if it covers everything you are looking at in #2070 and #2073

PR #2096, opened 2021-05-11
External Compactions

### LINKED ISSUE #1451: Support external compactions in containers

dlmarion:
closes #1451 by adding two new optional components to support running compactions outside of the TabletServer


milleruntime:
FYI I plan on taking a look at these changes some time this week. Instead of just looking at ALL the changes in the diff, my strategy is to checkout the branch, run some stuff using Uno and then take a look at some of the important files.

milleruntime:
@dlmarion it would help a lot for developers reviewing to split this PR up into 2 commits: 1 with your changes and the other with the generated thrift files. This is something we have done in the past with big PRs that have thrift changes.

dlmarion:
@milleruntime - what's the best way to approach that? Create a branch off of this branch, remove the thrift files from this branch and remove the non-thrift files from the other?

dlmarion:
Another option would be to click the "Viewed" checkbox on all of the thrift files, then they would be removed from your view. Not sure if that persists across browser sessions or not.

milleruntime:
> @milleruntime - what's the best way to approach that? Create a branch off of this branch, remove the thrift files from this branch and remove the non-thrift files from the other?

I was thinking remove the thrift files, do a typical build, commit your changes and then build again with the `-Pthrift` and then commit the generated files. Hopefully you still have the thrift executable!

milleruntime:
> Another option would be to click the "Viewed" checkbox on all of the thrift files, then they would be removed from your view. Not sure if that persists across browser sessions or not.

I think this will only hide the .thrift files (which I want to see). It won't hide the generated Java files.

dlmarion:
I was referring to [this](https://github.blog/2019-07-01-mark-files-as-viewed/). I think the default in the File Filter dropdown is to remove them from your view.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionExecutorIdImpl.java
```diff
@@ -0,0 +1,51 @@
...
+ * under the License.
+ */
+package org.apache.accumulo.core.util.compaction;
+
+import org.apache.accumulo.core.spi.compaction.CompactionExecutorId;
+import org.apache.accumulo.core.spi.compaction.CompactionServiceId;
+
+import com.google.common.base.Preconditions;
+
+public class CompactionExecutorIdImpl extends CompactionExecutorId {
```
milleruntime, line 26: I assume this class was created to be separate from the SPI but I am not seeing anything internal exposed. It seems like it could just be included in the super class.

milleruntime (commented):
Some small things I noticed just from starting to look at the code.

core/src/main/java/org/apache/accumulo/core/clientImpl/InstanceOperationsImpl.java
```diff
@@ -204,6 +210,54 @@ public boolean testClassLoad(final String className, final String asTypeName)
...
+
+      for (HostAndPort compactorAddr : compactors) {
+        futures
+            .add(
+                executorService
+                    .submit(
+                        () -> ExternalCompactionUtil.getActiveCompaction(compactorAddr, context)
+                            .stream().map(tac -> new ActiveCompactionImpl(context, tac,
+                                compactorAddr, CompactionHost.Type.COMPACTOR))
+                            .collect(Collectors.toList())));
```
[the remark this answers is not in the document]
keith-turner: oh yeah this is awful, i think I did this.  I suspect breaking it out to multiple lines would help a lot.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionExecutorIdImpl.java
keith-turner: I was trying to avoid putting all of the external/internal method in the SPI class.

milleruntime:
I was able to successfully run the example @dlmarion gave on the website and it is pretty cool! I was also able to catch an entry in the metadata:
<pre>
4< ecomp:ECID:04d6f4a3-3caa-46e6-9ea5-ce77a0ddc8dc []	{"inputs":["hdfs://localhost:8020/accumulo/tables/4/default_tablet/A000005v.rf"],"nextFiles":[],"tmp":"hdfs://localhost:8020/accumulo/tables/4/default_tablet/A000005y.rf_tmp","dest":"hdfs://localhost:8020/accumulo/tables/4/default_tablet/A000005y.rf","compactor":"ip-10-113-12-25:9101","kind":"USER","executorId":"DCQ1","priority":288230376151711745,"propDels":false,"selectedAll":true,"compactionId":2}
</pre>
I love how everything about the compaction is explicit AND in one place. I think this will be invaluable for debugging.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionExecutorIdImpl.java
milleruntime: I get that. Thinking ahead is good but why make it more awkward right now if we don't have to? I was also thinking in this case having those "i" and "e" prefixes in the SPI would be good and there isn't much to `CompactionExecutorId`

keith-turner:
> I was able to successfully run the example @dlmarion gave on the website and it is pretty cool! I was also able to catch an entry in the metadata:

@milleruntime looking at your comment I opened #2110. This PR is so large, I think we can use the external compaction project to open issue and not just rely on the comment in this PR.  That opens the possibility of merging this PR and doing some of those issues as follow on PRs.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionExecutorIdImpl.java
keith-turner: > why make it more awkward right now if we don't have to?

None of those methods are needed to write a compaction planner plugin, they are only used internally by the implementation outside of the realm of SPI interfaces.  So exposing them in the SPI could make changing the implementation harder in the future.

milleruntime:
> @milleruntime looking at your comment I opened #2110. This PR is so large, I think we can use the external compaction project to open issue and not just rely on the comment in this PR. That opens the possibility of merging this PR and doing some of those issues as follow on PRs.

That works since I don't anticipate seeing something that would prevent you guys from merging. I will just keep putting comments here against the code in the PR since that is easier for me. They could all be follow on work if that's easier for you guys.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionExecutorIdImpl.java
keith-turner: To give some context using the DefaultPlanner as an example.  It will make SPI calls that give it a CompactionExecutorId like [here](https://github.com/apache/accumulo/blob/129d06b66ca52307a77f12cee203b70e3604585a/core/src/main/java/org/apache/accumulo/core/spi/compaction/DefaultCompactionPlanner.java#L126).  In never instantiates a CompactionExecutorId.  Then later it will pass those IDs back when planning like [here](https://github.com/apache/accumulo/blob/129d06b66ca52307a77f12cee203b70e3604585a/core/src/main/java/org/apache/accumulo/core/spi/compaction/DefaultCompactionPlanner.java#L240-L242)

core/src/main/java/org/apache/accumulo/core/clientImpl/InstanceOperationsImpl.java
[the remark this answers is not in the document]
keith-turner: Improved this a bit in 162dd806b1754f890c4f139a3611c54024c072a2.  Tried to rename the var `tac` to `thriftActiveCompaction` but it caused the formatting of the update code to be unsightly.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1022,6 +1022,49 @@
...
+  COORDINATOR_THRIFTCLIENT_CLIENTPORT("coordinator.thrift.port.client", "9100", PropertyType.PORT,
+      "The port used for handling Thrift client connections on the compaction coordinator server"),
+  COORDINATOR_THRIFTCLIENT_MINTHREADS("coordinator.server.thrift.threads.minimum", "1",
+      PropertyType.COUNT, "The minimum number of threads to use to handle incoming requests."),
+  COORDINATOR_THRIFTCLIENT_MINTHREADS_TIMEOUT("coordinator.server.thrift.threads.timeout", "0s",
+      PropertyType.TIMEDURATION,
+      "The time after which incoming request threads terminate with no work available.  Zero (0) will keep the threads alive indefinitely."),
+  COORDINATOR_THRIFTCLIENT_THREADCHECK("coordinator.server.thrift.threadcheck.time", "1s",
+      PropertyType.TIMEDURATION, "The time between adjustments of the server thread pool."),
+  COORDINATOR_THRIFTCLIENT_MAX_MESSAGE_SIZE("coordinator.server.thrift.message.size.max", "10M",
```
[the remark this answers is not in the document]
keith-turner: Removing `thrift` from the property names would make them consistent w/ the compactor properties.

core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionFinalState.java
```diff
@@ -0,0 +1,131 @@
...
+
+    Extent(KeyExtent extent) {
+      this.tableId = extent.tableId().canonical();
+      if (extent.endRow() != null) {
+        er = Base64.getEncoder().encodeToString(TextUtil.getBytes(extent.endRow()));
+      }
+
+      if (extent.prevEndRow() != null) {
+        per = Base64.getEncoder().encodeToString(TextUtil.getBytes(extent.prevEndRow()));
+      }
```
milleruntime, line 91: I think these could be final:
```suggestion
    final String tableId;
    final String er;
    final String per;

    Extent(KeyExtent extent) {
      this.tableId = extent.tableId().canonical();
      if (extent.endRow() != null) {
        er = Base64.getEncoder().encodeToString(TextUtil.getBytes(extent.endRow()));
      } else {
        er = null;
      }

      if (extent.prevEndRow() != null) {
        per = Base64.getEncoder().encodeToString(TextUtil.getBytes(extent.prevEndRow()));
      } else {
        per = null;
      }
```

core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionFinalState.java
```diff
@@ -0,0 +1,131 @@
...
+
+  public enum FinalState {
+    FINISHED, FAILED
+  }
+
+  private ExternalCompactionId ecid;
+  private KeyExtent extent;
+  private FinalState state;
+  private long fileSize;
+  private long fileEntries;
```
milleruntime, line 44: These fields could be final
```suggestion
  private final ExternalCompactionId ecid;
  private final KeyExtent extent;
  private final FinalState state;
  private final long fileSize;
  private final long fileEntries;
```

milleruntime (commented):
OK I looked through all the code except the thrift stuff and the tests. I am hoping someone else can take a look at those.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
```diff
@@ -0,0 +1,766 @@
...
+              QueueAndPriority qp =
+                  QueueAndPriority.get(summary.getQueue().intern(), summary.getPriority());
+              synchronized (qp) {
+                TIME_COMPACTOR_LAST_CHECKED.computeIfAbsent(qp.getQueue(), k -> 0L);
+                QUEUE_SUMMARIES.update(tsi, summaries);
+              }
+            });
+          } finally {
+            ThriftUtil.returnClient(client);
+          }
```
keith-turner, line 391: Wondering if these RPCs should be done in a thread pool so that multiple tserver can be contacted in parallel.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
```diff
@@ -0,0 +1,766 @@
...
+
+  private static final Logger LOG = LoggerFactory.getLogger(CompactionCoordinator.class);
+  private static final long TIME_BETWEEN_GC_CHECKS = 5000;
+  private static final long FIFTEEN_MINUTES =
+      TimeUnit.MILLISECONDS.convert(Duration.of(15, TimeUnit.MINUTES.toChronoUnit()));
+
+  protected static final QueueSummaries QUEUE_SUMMARIES = new QueueSummaries();
+
+  /* Map of compactionId to RunningCompactions */
+  protected static final Map<ExternalCompactionId,RunningCompaction> RUNNING =
```
keith-turner, line 98: Maybe the dead compaction detector should remove things from here when it finds a dead external compaction.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/RunningCompaction.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public TExternalCompactionJob getJob() {
+    return job;
+  }
+
+  public String getCompactorAddress() {
+    return compactorAddress;
+  }
+
+  public TServerInstance getTserver() {
```
keith-turner, line 58: This may not be used, could possibly be removed.

keith-turner (commented):
I am looking back over the code since we worked on it over such a long period.  I am not finished looking back over everything yet.  Some things I saw I commented on and some I changed in 746748ac08994fdee4f72570b8feb04e6c50a75e

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionJobImpl.java
```diff
@@ -40,14 +42,25 @@
...
+    this.hasSelectedAll = true;
+  }
+
+  public CompactionJobImpl(long priority, CompactionExecutorId executor,
+      Collection<CompactableFile> files, CompactionKind kind) {
+    this.priority = priority;
+    this.executor = Objects.requireNonNull(executor);
+    this.files = Set.copyOf(files);
+    this.kind = Objects.requireNonNull(kind);
+    this.hasSelectedAll = false;
```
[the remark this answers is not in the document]
keith-turner: I will look into this.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -277,6 +378,130 @@ private void checkIfUserCompactionCanceled() {
     }
   }
 
+  private void initializeSelection(
```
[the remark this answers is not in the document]
keith-turner: I will add some comments documenting what is going on. Also I opened #2111 because I am not sure this code is tested.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -539,54 +766,58 @@ public boolean isCompactionEnabled() {
...
     Long checkCompactionId = null;
     boolean propogateDeletes = true;
     CompactionHelper localHelper;
     List<IteratorSetting> iters = List.of();
     CompactionConfig localCompactionCfg;
+    boolean selectedAll;
+    Set<StoredTabletFile> selectedFiles;
+  }
+
+  private CompactionInfo reserveFilesForCompaction(CompactionServiceId service, CompactionJob job) {
```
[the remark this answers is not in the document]
keith-turner: Added a comment and used optional

core/src/main/java/org/apache/accumulo/core/conf/Property.java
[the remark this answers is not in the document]
dlmarion: Working on this.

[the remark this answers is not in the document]
dlmarion: Resolved in 011a861832036fd22d0a8ee55e4daa2e11c90801

server/base/src/main/java/org/apache/accumulo/server/compaction/RetryableThriftCall.java
```diff
@@ -0,0 +1,123 @@
...
+   * maxNumRetries is 0, then this will retry forever. If maxNumRetries is non-zero, then a
+   * RuntimeException is thrown when it has exceeded he maxNumRetries parameter.
+   *
+   * @return T
+   * @throws RetriesExceededException
+   *           when maximum number of retries has been exceeded and the cause is set to the last
+   *           TException
+   */
+  public T run() throws RetriesExceededException {
+    long waitTime = start;
```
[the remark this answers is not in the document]
dlmarion: Resolved in 3983cd23629789c1ed0020d01fdd968eabfe590f

ctubbsii:
Based on the documentation in apache/accumulo-website#282 , it seems that the way the external compactions are exposed through to the user in the compaction configuration is through the specially named `externalQueue`, which seems to be a special feature of the default compaction service. Is that correct?

From the perspective of the user configuring things, it seems difficult to reason about how this configuration maps to pluggable services that they are deploying and administering, because the external compaction services aren't configured explicitly, but are an internal feature of a monolithic compaction service implementation. It seems like it will try to "automagically" use external services because of a specially named queue, and not because some plugin like `my.external.compaction.CompactionExecutorService` is configured explicitly as an executor (unless I'm mistaken here; I'm still going through the code).

I'm wondering if there are ways this could be made a bit more modular and the configuration a bit more explicit, so that the default compaction service isn't so monolithic, and users can very explicitly swap out the external executor implementation for another one. This specific implementation could be maintained in a side repository, deployed as a separate optional jar, and configured by users by some class name at an appropriate place in the configuration that specifies a plugin implementation. However, it should also be capable of being easily swapped out for a different external compaction implementation, and I'm not sure that's the case right now.

My main concern is that if the implementation is too monolithic, it creates a maintenance burden to continue to add marginal improvements to the monolithic implementation to satisfy every divergent use case, rather than encourage users to take advantage of the modular nature of our plugins to develop implementations to satisfy specific use cases. Ideally, we'd really want to make things pluggable/modular and the plugin interfaces relatively lightweight.

dlmarion:
> it seems that the way the external compactions are exposed through to the user in the compaction configuration is through the specially named externalQueue, which seems to be a special feature of the default compaction service. Is that correct?

Currently, yes. I believe the intention was to allow for users to configure compactions on their table(s) such that some could run internally (say, smaller ones) and other (larger) externally.

> I'm wondering if there are ways this could be made a bit more modular and the configuration a bit more explicit, so that the default compaction service isn't so monolithic, and users can very explicitly swap out the external executor implementation for another one.

I don't think that one of our goals was to create a pluggable external implementation. It's possible it was, I just don't remember that. Is that something we want?

keith-turner:
>  it seems that the way the external compactions are exposed through to the user in the compaction configuration is through the specially named externalQueue, which seems to be a special feature of the default compaction service. Is that correct?

Its a feature of the default compaction planner, not the default compaction service.  Each compaction service has a configured planner that decides what to compact where.  The [default compaction planner](https://github.com/apache/accumulo/blob/3983cd23629789c1ed0020d01fdd968eabfe590f/core/src/main/java/org/apache/accumulo/core/spi/compaction/DefaultCompactionPlanner.java#L73) accepts the config externalQueue.  It then takes this information and calls the SPI method [getExternalExecutor](https://github.com/apache/accumulo/blob/3983cd23629789c1ed0020d01fdd968eabfe590f/core/src/main/java/org/apache/accumulo/core/spi/compaction/ExecutorManager.java#L38).  Anyone could write a compaction planner that uses external queues by calling this same API method.  The planners configuration could work any way the author likes.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
```diff
@@ -0,0 +1,766 @@
...
+              } finally {
+                ThriftUtil.returnClient(client);
+              }
+            } else {
+              LOG.info("Tablet server {} is not currently in live tserver set",
+                  tabletMetadata.getLocation().getHostAndPort());
+            }
+          } else {
+            LOG.info("No current location for extent: {}", extent);
+          }
```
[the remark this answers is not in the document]
dlmarion: I guess the question is whether or not the CompactionCoordinator can get stale data from the metadata table. The code tries to find the TabletServer running this compaction. It's possible that the Tablet could be in the process of being re-hosted (in the case of a TServer crash). In the current implementation an entry would not be made in the internal running list because we didn't find the TServer. If we just use the Metadata entries, we could put incorrect information in the running list.

keith-turner:
> I'm wondering if there are ways this could be made a bit more modular and the configuration a bit more explicit, so that the default compaction service isn't so monolithic, and users can very explicitly swap out the external executor implementation for another one. 

In the current design tablets are authorities, within a distributed system, on what external compactions are running and queued.  There are RPCs to reserve, fail, and commit external compactions within the tserver AND there is an RPC to get summary information about external queues in the tserver.  These RPCs are structured specifically for the way the rest of the distributed system (coordinator and compactors) works.  Should these tserver RPCs be generalized?  If so, how?

ctubbsii:
@dlmarion wrote:
> I don't think that one of our goals was to create a pluggable external implementation. It's possible it was, I just don't remember that. Is that something we want?

My understanding of the original idea to create an "external" compactions, was to create them "external" to Accumulo... something Accumulo could "submit" work to do and not care about the implementation, and get back a new compacted file. The current implementation only makes them "external" to the tserver, not to Accumulo, and does so by growing Accumulo by 30,000 lines of code. This isn't necessarily bad, but it seems much more complex and tightly coupled to Accumulo's core than I think is maintainable long-term. My understanding of the original idea was to be able to submit compactions to some external service that the user could designate.... it could be a separate process, a bare metal server farm, EC2 nodes that scale on demand, or something else, depending on user needs. My understanding was that there would also be some flexibility in externalizing this functionality, and that implies making it pluggable in some way. Perhaps my understanding was incorrect, but that's what I had thought.

The bare minimum implementation that I was expecting to review at this point would be a pluggable service API for external compactions, and possibly a rudimentary in-process implementation. I wasn't quite expecting new worker processes, and a new coordination service, as well as new ZooKeeper nodes and RPCs. All of that seems to add a lot of complexity, and while that complexity seems necessary for a robust implementation as seems to be provided here (which is great, overall), I worry about maintainability in the future if it isn't modularized and pluggable, or flexibility based on divergent use cases.

Don't get me wrong... I have no issue with the basic design of this PR, as far as I can tell. I'm just wondering how we can improve upon the API/configuration to make it more modular so it's more maintainable in the future, and more flexible for users if they need to swap in alternate implementations of externalized workers to do the actual compacting. Right now, I don't know enough about the implementation to propose suggestions for how to do that... I'm only raising the issue to those who might know. As I try to dive in and learn the implementation, I may have more concrete suggestions. But, this is a lot of code, so it may take time to get that point.

keith-turner:
> My understanding of the original idea to create an "external" compactions, was to create them "external" to Accumulo...

That was never an explicit goal I had. To achieve this goal, thinking about the following.  Currently the implementation of the coordinator and compactors components use a lot of internal Accumulo code.  To move these to external repo would require refactoring them to use only API and SPI calls into Accumulo. To do this would require exposing a lot internal low level API/SPI calls related to compactions(not sure what these would be, that would need exploration).   For this exercise to be useful these newly exposed API/SPI calls would have to be generally useful and not overfit (meaning the new API/SPI calls can only be used to implement the current design).  To avoid overfitting, it seems we would need to have multiple other designs in mind that we could test against these new API/SPI calls to verify that we have not overfit.  Is there another path other than what I have described?  Does this path actually lower the complexity? I am not sure it would, I think the answer to that depends on what these new API/SPI calls are.

dlmarion:
In the current website PR I removed our original design document because @keith-turner and I had made several changes over the course of development. However, I still have the original documentation in another branch. I revisited our [original](https://github.com/dlmarion/accumulo-website/blob/external-compaction-design-capture-all-information/design/external-compaction.md) design and I think what we have today is closely aligned with what we envisioned at the beginning of this process. Having the benefit of already writing the external compaction code, I'm not sure that a separate and independent service  could be achieved without a significant set of new APIs being written for Accumulo.

I think @keith-turner pointed to the planner being the place where someone can write their own compaction implementation. But, I do agree that this initial implementation is tightly coupled to Accumulo internals. Maybe this is a stepping stone to a fully pluggable implementation. Looking at the coordinator should tell us what it needs in a public API to make it fully external.

Finally, my personal goal here was to move compactions out of the TServer for several reasons (below). The Compactor component can be run on different hardware than the TabletServers, and even run in Kubernetes using their dynamic pod scheduling feature to scale up/down the number of Compactors based on load.
 
  * Allow compactions to outlive a TabletServer
  * Allow compactions to occur concurrent to a Tablet being re-hosted
  * Reduce the load on the TabletServer, giving it more cycles to insert mutations and respond to scans
  * Allow compactions to be scaled differently than the number of TabletServers

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
```diff
@@ -0,0 +1,766 @@
...
+   *          trace info
+   * @param credentials
+   *          tcredentials object
+   * @param externalCompactionId
+   *          compaction id
+   * @throws TException
+   *           thrift error
+   */
+  @Override
+  public void cancelCompaction(TInfo tinfo, TCredentials credentials, String externalCompactionId)
```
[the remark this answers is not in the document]
dlmarion: Resolved in cdfec2afccd22ea344ad8fd3a3ce622ee5b87324

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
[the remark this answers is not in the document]
keith-turner: I may be missing something.  I was looking at the code saw a good bit of effort went into finding `tsi`, however it seems after the information was found that it was never used.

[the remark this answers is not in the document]
dlmarion: So, it looks like the CompactionFinalizer scans the Metadata table to determine which TabletServer to notify when a compaction fails or succeeds. We might be able to remove this after all.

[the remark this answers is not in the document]
keith-turner: To be more specific, my IDE indicated nothing calls [RunningCompaction.getTserver()](https://github.com/apache/accumulo/blob/cdfec2afccd22ea344ad8fd3a3ce622ee5b87324/server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/RunningCompaction.java#L58)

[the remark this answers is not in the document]
keith-turner: To be more specific, my IDE indicated nothing calls [RunningCompaction.getTserver()](https://github.com/apache/accumulo/blob/cdfec2afccd22ea344ad8fd3a3ce622ee5b87324/server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/RunningCompaction.java#L58)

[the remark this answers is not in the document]
dlmarion: Agreed, I'm going to push another commit shortly that will significantly reduce the lines of code for the re-start and remove the getTServer method from RunningCompaction.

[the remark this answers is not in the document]
dlmarion: resolved in 90e76cd59e5ed6e08a75020f0d50fe69c4bebee3

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/RunningCompaction.java
dlmarion: resolved in 90e76cd59e5ed6e08a75020f0d50fe69c4bebee3

ctubbsii:
@keith-turner wrote:
> These RPCs are structured specifically for the way the rest of the distributed system (coordinator and compactors) works. Should these tserver RPCs be generalized? If so, how?

Sorry, I don't know enough about the current implementation to understand the question yet, in order to answer it. The generalized interface I was thinking about was something like `BiFunction<Set<File>, CompactionConfig, Future<File>>`. I was thinking any necessary RPCs would be implementation-specific, and independent of Accumulo's RPCs. My preconceptions of what this would look like have been broken, so I would need to do some catch up understanding before I could have specific thoughts on how they could be improved.

> > My understanding of the original idea to create an "external" compactions, was to create them "external" to Accumulo... something Accumulo could "submit" work to do and not care about the implementation, and get back a new compacted file.
> 
> That was never an explicit goal I had.

Okay, I guess I just had a different idea in mind of what "external" meant.

> Is there another path other than what I have described? Does this path actually lower the complexity? I am not sure it would, I think the answer to that depends on what these new API/SPI calls are.

If we went with the `BiFunction<Set<File>, CompactionConfig, Future<File>>` concept, or similar, then nothing internal need be exposed to it. The RFile config settings (locality groups, etc.) and iterators are already "public API". A lot of the compaction complexity, with the services, planners, executors, queues, workers, etc. could sit inside an implementation. Or, all that could be left as-is, and just expect the pluggable part to be the "last stage" after going through that management pipeline, which I think is basically how it is now... but not expressed as simply as I have with the "BiFunction" expression.

I'm curious, do the external compactors support RFile encryption and other RFile-specific configs (like block sizes, locality groups, etc.) today? My guess is that they do, because it's still all the same Accumulo code, just running in a different process, but just curious how much you've exercised those while testing this.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
dlmarion: I wonder if the DeadCompactionDetector should have a reference to the CompactionCoordinator and call `compactionFailed` instead of calling the CompactionFinalizer directly, that way the behavior and logic are the same as a normal failure.

keith-turner:
Omitting a lot of detail, the following is basics of what the current system enable.  The first step uses a pluggable planner.

1. Plan compaction :  Take current tablets files, current running compactions for the tablet, and emit set of compactions jobs.  A compaction job is set of files to compact, a  priority, and a destination queue.
 2. Queue Jobs : An attempt to queue the jobs emitted from the planner is made by doing the following.
    1. Cancel anything that was previously queued by the planner for the tablet.  If this fails, go back to planning step and try again because things probably changed during planning.
    2. Queue the new job from the planner on the desired priority queues
 3. When an internal or external compactor thread finishes a task, it takes the next highest priority job from the queue and actually run the compaction.

I don' think interface `BiFunction<Set<File>, CompactionConfig, Future<File>>` would be sufficient to achieve this functionality.  It would not support the current functionality of priority queues and canceling queued work when things change (new files arrive, job start/finish during planning). In the current impl the priority queue for local compactions is very precise.  For external compaction there is essentially a global priority queue that is approximate (eventually consistent) that may not always start the highest priority job next, but usually will.

keith-turner:
> I'm curious, do the external compactors support RFile encryption and other RFile-specific configs (like block sizes, locality groups, etc.) today? My guess is that they do, because it's still all the same Accumulo code, just running in a different process, but just curious how much you've exercised those while testing this.

The intent is that they should, but not everything has been tested.  The [code that actually runs compactions of files](https://github.com/apache/accumulo/blob/90e76cd59e5ed6e08a75020f0d50fe69c4bebee3/server/base/src/main/java/org/apache/accumulo/server/compaction/FileCompactor.java#L73) was refactored and moved so that it could be used by tservers and external compactors.  So the same code is used, the question is do we always give it the same inputs.  Also if compactions were to be highly generalized, some analogue of this code would probably need to be exposed via SPI or API.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
keith-turner: That could work w/ some slight tweaks.  The compactionFailed method in the coordinator is kinda thrifty and only takes a single ecid.  Could have package private non thrifty compactionFailed method in the coordinator that takes a collections ecids and is called by the dead compaction detector and the thrifty compactionFailed method.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -0,0 +1,889 @@
...
+        var extent = KeyExtent.fromThrift(job.getExtent());
+        var ecid = ExternalCompactionId.of(job.getExternalCompactionId());
+
+        TabletMetadata tabletMeta =
+            getContext().getAmple().readTablet(extent, ColumnType.ECOMP, ColumnType.PREV_ROW);
+        if (tabletMeta == null || !tabletMeta.getExtent().equals(extent)
+            || !tabletMeta.getExternalCompactions().containsKey(ecid)) {
+          // table was deleted OR tablet was split or merged OR tablet no longer thinks compaction
+          // is running for some reason
+          LOG.info("Cancelling compaction {} that no longer has a metadata entry at {}", ecid,
```
keith-turner, line 194: @ctubbsii I Am currently working through a complete review of the code.  You had mentioned having a very generic per tserver plugin to generally handle external compactions.  Thinking about that when looking at this code, it cancels a running external compaction for a tablet that was deleted (split/merge/table deletion).  To handle this case in a highly pluggable environment, this functionality would need to be present in some form in the SPI.   This would be unrelated to any pluggable component on a tserver, its more related to running external compactions.  The tablet for a running external compaction could be unload and then the tablet deleted, no tserver would be aware of this.

server/tserver/src/main/java/org/apache/accumulo/tserver/ThriftClientHandler.java
```diff
@@ -1654,6 +1658,79 @@ public void compact(TInfo tinfo, TCredentials credentials, String lock, String t
...
+          SecurityErrorCode.PERMISSION_DENIED).asThriftException();
+    }
+
+    server.getCompactionManager().externalCompactionFailed(
+        ExternalCompactionId.of(externalCompactionId), KeyExtent.fromThrift(extent),
+        server.getOnlineTablets());
+  }
+
+  @Override
+  public boolean isRunningExternalCompaction(TInfo tinfo, TCredentials credentials,
```
keith-turner, line 1724: This thrift method may not be called by anything.

dlmarion: I'll look at this.

dlmarion: The code in the CompactionCoordinator that called this method was removed yesterday when I removed TServer from RunningCompaction. This was resolved in 937d7eeb8059380138ab384eacdfe3aa79b688fb. Removing this simple method from the code reduced the Thrift changes in this PR by about 100K LOC.

j/k.

dlmarion:
@milleruntime @ctubbsii - I'm looking for an existing IT that tests encryption so I can write something for the external compactions. I have found CryptoTest, which is a unit test, and WriteAheadLogEncryptedIT which doesn't have anything that might be relevant for me. Do you know of an existing test that I can look to for an example?

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
dlmarion: I'm going to look at making this change.

dlmarion: @keith-turner - I resolved this in f39707a492aea982b5127c7358e4a7f209a812f0

milleruntime:
> @milleruntime @ctubbsii - I'm looking for an existing IT that tests encryption so I can write something for the external compactions. I have found CryptoTest, which is a unit test, and WriteAheadLogEncryptedIT which doesn't have anything that might be relevant for me. Do you know of an existing test that I can look to for an example?

That is currently all we have for automated tests. The other testing we did was manual using Uno, which has a configurable plugin for crypto.

dlmarion:
@ctubbsii wrote:
> it seems that the way the external compactions are exposed through to the user in the compaction configuration is through the specially named externalQueue

The DefaultCompactionPlanner javadoc is out of date, it's doesn't describe `externalQueue`. Looking at `DefaultCompactionPlanner.ExecutorConfig` and `DefaultCompactionPlanner.init()` it seems it would be trivial to change the configuration so that it was more explicit. We could add a `type` field, for example:

```
[
 {
   "name": "small",
   "type": "internal",
   "maxSize": "16M",
   "numThreads": 4
 },
 {
   "name": "large",
   "type": "external",
   "queue": "LargeQ"
 }
]
```

keith-turner:
Thinking through plugin interfaces and alternative designs, I thought of a very different way compactions could work in Accumulo.  We could have a process (like the manager or compaction coordinator) that continually scans the files in the metadata table looking for compaction work that needs to be done per tablet and gives that to external compaction processes.  Under this model tablet server would have no compaction code at all and there would only be external compactions. The tserver would only need a mechanism to replace a set of files in a tablet to support this model.  I think this is an interesting model to think about, but I am not advocating for it.  It has pluses and minuses over the current model.  The reason I am sharing it is that I think it informs the general discussion about API and SPI interfaces around compaction and supporting alternative designs.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
```diff
@@ -313,6 +291,61 @@ public void run() {
     LOG.info("Shutting down");
   }
 
+  private void updateSummaries() {
+    ExecutorService executor =
+        ThreadPools.createFixedThreadPool(10, "Compaction Summary Gatherer", false);
```
dlmarion, line 296: Should the number of threads (10 here) be some factor (like 10%) of the # of tservers in tServerSet?

keith-turner: I Was thinking of % up to some max (was thinking a max of 10)... and then just decided on the max... that should get through even a large number of tsevers in short order, except when there are a good bit of unresponsive tservers... was also wondering if it should be configurable, but was thinking 10 should be good and we can make it configurable if ever needed

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/CompactionCoordinator.java
keith-turner: Done in b59c13ff80da6491a89757a00c0fa32166d038d2

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableUtils.java
```diff
@@ -534,10 +540,10 @@ public static AccumuloConfiguration getCompactionConfig(CompactionKind kind, Tab
   }
 
   static StoredTabletFile compact(Tablet tablet, CompactionJob job, Set<StoredTabletFile> jobFiles,
-      Long compactionId, boolean propogateDeletes, CompactableImpl.CompactionHelper helper,
-      List<IteratorSetting> iters, CompactionCheck compactionCheck, RateLimiter readLimiter,
-      RateLimiter writeLimiter, CompactionStats stats)
-      throws IOException, CompactionCanceledException {
+      Long compactionId, Set<StoredTabletFile> selectedFiles, boolean propogateDeletes,
```
DomGarguilo, line 543:
```suggestion
      Long compactionId, Set<StoredTabletFile> selectedFiles, boolean propagateDeletes,
```

DomGarguilo: Looks like there are lots of occurrences of this typo actually. propogate vs propagate

Manno15:
During one of my basic runs testing this code (two servers, no agitation) I ran into an error in the logs. I did not notice any other issue though and I am not sure if this even interrupted anything. Screen shot below:


image


<details>

```
java.lang.IllegalStateException
	at com.google.common.base.Preconditions.checkState(Preconditions.java:494)
	at org.apache.accumulo.tserver.tablet.CompactableImpl.commitExternalCompaction(CompactableImpl.java:1055)
	at org.apache.accumulo.tserver.compactions.CompactionManager.commitExternalCompaction(CompactionManager.java:474)
	at org.apache.accumulo.tserver.ThriftClientHandler.compactionJobFinished(ThriftClientHandler.java:1705)
	at org.apache.accumulo.core.trace.TraceUtil.lambda$wrapService$6(TraceUtil.java:235)
	at com.sun.proxy.$Proxy40.compactionJobFinished(Unknown Source)
	at org.apache.accumulo.core.tabletserver.thrift.TabletClientService$Processor$compactionJobFinished.getResult(TabletClientService.java:4001)
	at org.apache.accumulo.core.tabletserver.thrift.TabletClientService$Processor$compactionJobFinished.getResult(TabletClientService.java:3982)
	at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:38)
	at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39)
	at org.apache.accumulo.server.rpc.TimedProcessor.process(TimedProcessor.java:63)
	at org.apache.thrift.server.AbstractNonblockingServer$FrameBuffer.invoke(AbstractNonblockingServer.java:518)
	at org.apache.accumulo.server.rpc.CustomNonBlockingServer$CustomFrameBuffer.invoke(CustomNonBlockingServer.java:114)
	at org.apache.thrift.server.Invocation.run(Invocation.java:18)

```
</details>

dlmarion:
That's in a finally block, line 1055 is:
```
Preconditions.checkState(externalCompactionsCommitting.remove(extCompactionId));
```

dlmarion:
@Manno15  wrote:
> This fills up the log file and it will eventually get pretty large with it mostly repeating the same information. Screenshot of the particular logs that I am talking about. I am not necessarily sure if this is an issue or not but is something I noticed.

Yeah, we could probably turn that to debug. No issue there, just chatty

keith-turner:
@Manno15 re the exception you saw.  I am seeing slightly different code that @dlmarion  saw at line 1055.  What commit were you testing with?  Or can you look in the source you built from and see what is at CompactableImpl.java line 1055? I am seeing the following line in the latest commit.

```java
Preconditions.checkState(!closed);
```

keith-turner:
@Manno15 I think I may have fixed the issue you were seeing in 3095d3d, if line 1055 was the !closed check.  That was an incorrect sanity check, looking at how the close() method works, closed could be true and that would be ok.

keith-turner:
@Manno15 the feedback on logging was great.  I made some quick changes with the following goals.

 * At INFO when things are working normally there should only be  two INFO message per compaction. One message before and one after.
 * At DEBUG there are multiple messages per compaction giving more detailed information about compactions that are working as expected.
 * At TRACE there are only message that people would want to see when there is a suspected bug in the compactor code or performance problem.  These are messages that people would not want to see whenever the system is working as expected.

I Am not sure if I achieved the INFO goal of two message per compaction because I didn't try running it.

keith-turner:
>  I noticed with the CompactionCoordinator and CompactionFinalizer both have filesize and entries parameters but they are in the opposite order as each other.

Consistency is nice.  Looking the message you posted, thinking we should also minimize the INFO message per compaction in the coordinator like the Compactor.

keith-turner:
@Manno15 re the consistency of those fields, that data was coming from a thrift class and GSON generator json, so not sure they could reliably be ordered consistently.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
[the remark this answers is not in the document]
keith-turner: @milleruntime  I added a comment in e81e790.  Let me know if that looks good.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -95,6 +117,8 @@
 
   private Supplier<Set<CompactionServiceId>> servicesInUse;
 
+  private Set<CompactionServiceId> servicesUsed = new ConcurrentSkipListSet<>();
+
   // status of special compactions
   private enum SpecialStatus {
     NEW, SELECTING, SELECTED, NOT_ACTIVE, CANCELED
```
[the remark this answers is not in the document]
keith-turner: I like your suggestion, renamed in e81e790

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -277,6 +378,130 @@ private void checkIfUserCompactionCanceled() {
...
+    boolean unexpectedExternal = false;
+    Set<StoredTabletFile> tmpSelectedFiles = null;
+    Boolean selAll = null;
+    Long cid = null;
+    Boolean propDel = null;
+    int count = 0;
+
+    ArrayList<String> reasons = new ArrayList<>();
+
+    for (Entry<ExternalCompactionId,ExternalCompactionMetadata> entry : extCompactions.entrySet()) {
```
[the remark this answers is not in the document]
keith-turner: I made a comment on #2111 that this code should be refactored so that it could be unit tested.  Thinking it would be  best to wait until then to refactor it.

core/src/main/java/org/apache/accumulo/core/spi/compaction/ExecutorManager.java
```diff
@@ -31,4 +31,9 @@
    * Create a thread pool executor within a compaction service.
    */
   public CompactionExecutorId createExecutor(String name, int threads);
+
+  /**
+   * @return an id for a configured external execution queue.
+   */
+  public CompactionExecutorId getExternalExecutor(String name);
```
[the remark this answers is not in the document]
keith-turner: The interface was not there so that users could implement their own, I was just using it to logically group things that went together to avoid having lots of unrelated groups of methods interleaved in a single interface.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionJobImpl.java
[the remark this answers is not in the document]
keith-turner: @milleruntime that was confusing. I made some changes in 2ba605cd62ea4c1d2cc4733c7a562a8f0fb2499d for this, let me know what you think.

Manno15:
> @Manno15 I think I may have fixed the issue you were seeing in 3095d3d, if line 1055 was the !closed check. That was an incorrect sanity check, looking at how the close() method works, closed could be true there and that would be ok.

Yeah, that is the line I saw for 1055. I was running the latest commit on this branch at that time.

> @Manno15 the feedback on logging was great. I made some quick changes with the following goals.

Great, I like the logging changes you made. I will test them out today. 

> @Manno15 re the consistency of those fields, that data was coming from a thrift class and GSON generator json, so not sure they could reliably be ordered consistently.

Ah, that makes sense. It's not a huge deal, just something I noticed while reading through it.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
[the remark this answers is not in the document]
milleruntime: Looks good, thanks!

keith-turner:
> These changes help, thanks. I am still wondering why we need both constructors? If you make the jobSelectedAll an Optional type parameter wouldn't that suffice?

@milleruntime  I made the constructor take an optional and added a comment about the parameter in 28c84b6

keith-turner:
> Though, I am not quite sure how to change the log levels for the compactor and coordinator log files. 

@Manno15 you should be able to edit `<ACCUMULO DIR>/conf/log4j2-service.properties` and add something like the following.

```
logger.compactor.name = org.apache.accumulo.compactor
logger.compactor.level = trace
```

The above is from a file I had laying around for a bit so not completely sure its still correct.  Could add similar lines for the coordinator. Changes should be picked up while processes are running (which is really really nice).

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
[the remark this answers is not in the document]
keith-turner: Sure @milleruntime that would be helpful.  Was thinking those test could be done after this is merged but before release.  It would be easier to review on its own.

Manno15:
@keith-turner @dlmarion is there any particular aspect or feature that you think needs to be tested specifically? I've done multiple runs with the normal testing procedures (multiple tservers, agitation, etc) and haven't run into any issues so far.

Manno15:
I also noticed that this log message below can get chatty in the right scenario (one Tserver cluster that has the Tserver crash or killed). In that scenario, it will repeat this log message until one TServer is brought back up. From what I see in `TServerUtils`, this may only feasibly occur on a one TServer cluster as the logic to obtain the larger pool size will never be larger than the active pool count. 
2021-05-21T07:44:01,916 [rpc.TServerUtils] INFO : Increasing server thread pool size on CompactionCoordinator to 1

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -0,0 +1,898 @@
...
+        // ever fail, it means there is a bug elsewhere.
+        Preconditions.checkState(compactionRunning.compareAndSet(false, true));
+        try {
+          LOG.info("Starting up compaction runnable for job: {}", job);
+          updateCompactionState(job, TCompactionState.STARTED, "Compaction started");
+
+          final AccumuloConfiguration tConfig;
+
+          if (!job.getTableCompactionProperties().isEmpty()) {
+            tConfig = new ConfigurationCopy(DefaultConfiguration.getInstance());
```
keith-turner, line 546: Looking how this works here and in the tserver code, thinking this is brittle and error prone. What if a non table props is used, like a general prop?  That could cause silent failures in this code.  Think we need to rework this, but not sure what to do ATM.  The peer code in the tserver is in CompactableImpl.reserveExternalCompaction()

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/CompactionExecutorsMetrics.java
```diff
@@ -77,10 +92,47 @@ public synchronized AutoCloseable addExecutor(CompactionExecutorId ceid,
...
+
+      MetricsRegistry registry = super.getRegistry();
+
+      synchronized (exMetrics) {
+        externalMetricsSupplier.get().forEach(ecm -> {
+          seenIds.add(ecm.ceid);
+
+          ExMetrics exm = exMetrics.computeIfAbsent(ecm.ceid, id -> {
+            ExMetrics m = new ExMetrics();
+            m.queuedGauge = registry.newGauge(id.canonical().replace('.', '_') + "_queued",
```
keith-turner, line 108: Wondering if a queue being configured out of existence causes any problems for the gauges that may be left around.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
keith-turner: Resolved this in c2da878084c51d59acf6cbda9a92f59d629d6a2b

keith-turner (approved):
I went back and reviewed all of the code since we worked on this for so long.  It took a bit, but I am finished reviewing.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableUtils.java
keith-turner: @DomGarguilo I ran

```bash
find . -name "*.java" | xargs sed -i '' -e 's/propogate/propagate/g'
```

 and pushed the changes in d90f42178f76171e785372e0bf1a04359d24e816

keith-turner:
> Everything works as expected from my testing. Ran numerous tests with varying parameters (with and without agitation, multi-node cluster, and varying amounts of ingested data).

Nice, did you verify the data after testing?

Manno15:
> Nice, did you verify the data after testing?

I have after each test run. No issue with data loss. However, I just re-ran verify and received an interesting error. The job itself completed successfully but then ended up with a compactor error. Server logs below.

<details>

```
2021-05-25T10:13:22,937 [compactor.Compactor] ERROR: Unhandled error occurred in Compactor

java.lang.RuntimeException: org.apache.zookeeper.KeeperException$ConnectionLossException: KeeperErrorCode = ConnectionLoss for /accumulo/3737bcb7-377b-4ad9-be42-efede2692e16/coordinators/lock
	at org.apache.accumulo.core.util.compaction.ExternalCompactionUtil.findCompactionCoordinator(ExternalCompactionUtil.java:86)
	at org.apache.accumulo.compactor.Compactor.getCoordinatorClient(Compactor.java:501)
	at org.apache.accumulo.compactor.Compactor$5.execute(Compactor.java:473)
	at org.apache.accumulo.compactor.Compactor$5.execute(Compactor.java:470)
	at org.apache.accumulo.server.compaction.RetryableThriftCall.run(RetryableThriftCall.java:107)
	at org.apache.accumulo.compactor.Compactor.getNextJob(Compactor.java:490)
	at org.apache.accumulo.compactor.Compactor.run(Compactor.java:663)
Caused by: org.apache.zookeeper.KeeperException$ConnectionLossException: KeeperErrorCode = ConnectionLoss for /accumulo/3737bcb7-377b-4ad9-be42-efede2692e16/coordinators/lock
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:102)
	at org.apache.zookeeper.KeeperException.create(KeeperException.java:54)
	at org.apache.zookeeper.ZooKeeper.getChildren(ZooKeeper.java:2746)
	at org.apache.accumulo.fate.zookeeper.ServiceLock.getLockData(ServiceLock.java:653)
	at org.apache.accumulo.core.util.compaction.ExternalCompactionUtil.findCompactionCoordinator(ExternalCompactionUtil.java:79)
	... 7 more
```

```
ERROR: Unable to call shutdownNow
java.lang.NoSuchFieldException: executorService_
        at java.lang.Class.getDeclaredField(Class.java:2411) ~[?:?]
        at org.apache.accumulo.server.rpc.TServerUtils.stopTServer(TServerUtils.java:717) ~[accumulo-server-base-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.compactor.Compactor.run(Compactor.java:793) ~[accumulo-compactor-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
2021-05-25T10:13:22,939 [compactor.Compactor] INFO : stop requested. exiting ...
2021-05-25T10:13:30,118 [compactor.Compactor] ERROR: Compactor lost lock (reason = LOCK_DELETED), exiting.
```
</details>

Everything seems to be up in the monitor so I am not quite sure what happened there.

dlmarion:
There are two approvals on this. I'm planning on merging this early next week.

keith-turner:
On a test table in Uno w/ 2B entries and 128 tablets I was able to run 40 concurrent external compactions in AKS Kubernetes today.  I am continuing to poke at this, but I think its looking pretty good.  I hope to run test on cluster+Kubernetes soon instead of Uno+Kubernetes.

```
root@uno> listcompactions
 SERVER               | AGE       | TYPE  | REASON | READ  | WROTE | TABLE           | TABLET                                   | INPUT | OUTPUT                              | ITERATORS | ITERATOR OPTIONS
172.20.128.185:9101 (ext) |   26s80ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;9401a;92012                            |     1 |        /2/t-0000cj0/A0002ay3.rf_tmp |        [] | {}
172.20.128.188:9101 (ext) |  23s856ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;ba01;b801                              |     1 |        /2/t-00017pv/A0002ayb.rf_tmp |        [] | {}
172.20.128.41:9101 (ext) |  34s598ms |  FULL |   USER | 3.64M | 3.64M |       testTable | 2;0801b;0602                             |     1 |        /2/t-00004hf/A0002axd.rf_tmp |        [] | {}
172.20.128.53:9101 (ext) |  35s340ms |  FULL |   USER | 3.88M | 3.88M |       testTable | 2;02014<                                 |     1 |        /2/t-00017dz/A0002axb.rf_tmp |        [] | {}
172.20.129.116:9101 (ext) |  23s812ms |  FULL |   USER | 2.40M | 2.40M |       testTable | 2;b801;b601                              |     1 |        /2/t-00004ij/A0002ayc.rf_tmp |        [] | {}
172.20.129.13:9101 (ext) |   22s80ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;b400b;b200b                            |     1 |        /2/t-0000cjc/A0002aye.rf_tmp |        [] | {}
172.20.129.184:9101 (ext) |  25s695ms |  FULL |   USER | 2.79M | 2.79M |       testTable | 2;9801c;9602                             |     1 |        /2/t-00004gy/A0002ay7.rf_tmp |        [] | {}
172.20.129.195:9101 (ext) |  34s670ms |  FULL |   USER | 3.30M | 3.30M |       testTable | 2;0401a;02014                            |     1 |        /2/t-0000cl4/A0002axc.rf_tmp |        [] | {}
172.20.129.80:9101 (ext) |  28s262ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;8c015;8a01e                            |     1 |        /2/t-0000cq8/A0002axv.rf_tmp |        [] | {}
172.20.130.1:9101 (ext) |  24s517ms |  FULL |   USER | 2.51M | 2.51M |       testTable | 2;be00c;bc00e                            |     1 |        /2/t-00017xu/A0002ay9.rf_tmp |        [] | {}
172.20.130.104:9101 (ext) |  23s609ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;b601;b400b                             |     1 |        /2/t-00017s0/A0002ayd.rf_tmp |        [] | {}
172.20.130.113:9101 (ext) |  30s713ms |  FULL |   USER | 3.88M | 3.88M |       testTable | 2;4201;4                                 |     1 |        /2/t-00017kz/A0002axn.rf_tmp |        [] | {}
172.20.130.201:9101 (ext) |  26s766ms |  FULL |   USER | 2.91M | 2.91M |       testTable | 2;9;8e00d                                |     1 |        /2/t-000000d/A0002ay0.rf_tmp |        [] | {}
172.20.130.249:9101 (ext) |  28s350ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;46017;44014                            |     1 |        /2/t-00017xf/A0002axu.rf_tmp |        [] | {}
172.20.131.143:9101 (ext) |  24s460ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;bc00e;ba01                             |     1 |        /2/t-0000cnb/A0002aya.rf_tmp |        [] | {}
172.20.131.150:9101 (ext) |  25s101ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;9a01f;9801c                            |     1 |        /2/t-00017qv/A0002ay8.rf_tmp |        [] | {}
172.20.131.35:9101 (ext) |  26s969ms |  FULL |   USER | 2.77M | 2.77M |       testTable | 2;48016;46017                            |     1 |        /2/t-00004ho/A0002axz.rf_tmp |        [] | {}
172.20.131.82:9101 (ext) |  33s715ms |  FULL |   USER | 3.36M | 3.36M |       testTable | 2;0c01a;0a01d                            |     1 |        /2/t-0000ctq/A0002axg.rf_tmp |        [] | {}
172.20.132.14:9101 (ext) |  25s729ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;4c01;4a018                             |     1 |        /2/t-0000col/A0002ay6.rf_tmp |        [] | {}
172.20.132.141:9101 (ext) |  28s416ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;8a01e;8801d                            |     1 |        /2/t-00017so/A0002axt.rf_tmp |        [] | {}
172.20.132.187:9101 (ext) |  28s737ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;8801d;86021                            |     1 |        /2/t-00004gp/A0002axs.rf_tmp |        [] | {}
172.20.132.226:9101 (ext) |  30s934ms |  FULL |   USER | 4.30M | 4.30M |       testTable | 2;8;7e00d                                |     1 |        /2/t-0000000/A0002axl.rf_tmp |        [] | {}
172.20.132.26:9101 (ext) |  29s994ms |  FULL |   USER | 3.17M | 3.17M |       testTable | 2;22016;2                                |     1 |        /2/t-00017lc/A0002axq.rf_tmp |        [] | {}
172.20.133.113:9101 (ext) |  29s711ms |  FULL |   USER | 3.20M | 3.20M |       testTable | 2;44014;4201                             |     1 |        /2/t-0000clm/A0002axr.rf_tmp |        [] | {}
172.20.133.144:9101 (ext) |  32s707ms |  FULL |   USER | 4.10M | 4.10M |       testTable | 2;1;0e013                                |     1 |        /2/t-0000008/A0002axi.rf_tmp |        [] | {}
172.20.133.182:9101 (ext) |  28s313ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;8e00d;8c015                            |     1 |        /2/t-00017zq/A0002axw.rf_tmp |        [] | {}
172.20.133.255:9101 (ext) |  26s703ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;92012;9                                |     1 |        /2/t-00017kk/A0002ay1.rf_tmp |        [] | {}
172.20.133.55:9101 (ext) |  25s724ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;26024;2401d                            |     1 |        /2/t-00017us/A0002ay5.rf_tmp |        [] | {}
172.20.134.114:9101 (ext) |  26s191ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;4a018;48016                            |     1 |        /2/t-00017wj/A0002ay2.rf_tmp |        [] | {}
172.20.134.139:9101 (ext) |  25s993ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;9602;9401a                             |     1 |        /2/t-00017rl/A0002ay4.rf_tmp |        [] | {}
172.20.134.161:9101 (ext) |  30s202ms |  FULL |   USER | 3.70M | 3.70M |       testTable | 2;86021;84019                            |     1 |        /2/t-00017hh/A0002axp.rf_tmp |        [] | {}
172.20.134.250:9101 (ext) |  30s186ms |  FULL |   USER | 3.19M | 3.19M |       testTable | 2;84019;82013                            |     1 |        /2/t-0000co6/A0002axo.rf_tmp |        [] | {}
172.20.134.61:9101 (ext) |   34s46ms |  FULL |   USER | 4.52M | 4.52M |       testTable | 2;0602;0401a                             |     1 |        /2/t-00017mi/A0002axf.rf_tmp |        [] | {}
172.20.135.166:9101 (ext) |   33s46ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;0e013;0c01a                            |     1 |        /2/t-0001816/A0002axh.rf_tmp |        [] | {}
172.20.135.191:9101 (ext) |  32s423ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;2;1e00e                                |     1 |        /2/t-0000005/A0002axj.rf_tmp |        [] | {}
172.20.135.222:9101 (ext) |  34s165ms |  FULL |   USER | 4.43M | 4.43M |       testTable | 2;0a01d;0801b                            |     1 |        /2/t-00017gs/A0002axe.rf_tmp |        [] | {}
172.20.135.41:9101 (ext) |  26s997ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;2401d;22016                            |     1 |        /2/t-0000cle/A0002axy.rf_tmp |        [] | {}
172.20.136.109:9101 (ext) |  27s177ms |  FULL |   USER | 2.98M | 2.98M |       testTable | 2;1201;1                                 |     1 |        /2/t-00017v7/A0002axx.rf_tmp |        [] | {}
172.20.136.162:9101 (ext) |  31s305ms |  FULL |   USER | 3.07M | 3.07M |       testTable | 2;4;3e00f                                |     1 |        /2/t-0000002/A0002axk.rf_tmp |        [] | {}
172.20.136.52:9101 (ext) |  30s742ms |  FULL |   USER | 3.35M | 3.35M |       testTable | 2;82013;8                                |     1 |        /2/t-00017iu/A0002axm.rf_tmp |        [] | {}
```

PR #2113, opened 2021-05-19
Move logic in LogSorter outside sort method

milleruntime:
* Move the error handling in the LogSorter outside of the sort method.
This will allow errors being thrown to be seen in the unit test


milleruntime:
Errors are showing up in the output of the test but being swallowed by the error handling code in `LogSorter`. The test is currently passing even though errors are being thrown. This change will expose any errors being thrown, most likely causing a test failure.

I am letting it run once to see the errors. I can then address the errors thrown by tests, either here or in a separate change.

milleruntime:
Not sure what is going on with Github actions. Successfully ran the checks locally:

- mvn clean verify -PskipQA -DskipTests=false
- mvn clean verify javadoc:jar -Psec-bugs -DskipTests

milleruntime:
I updated the Test by adding an expect throws for the old WAL, creating a new test for the latest WAL and refactoring the tests to use try-with-resources. It cleaned up the test a lot and I think it is good to go now.

keith-turner (approved):

PR #2118, opened 2021-05-25
PopulateMetadataTable Improvements

### LINKED ISSUE #1091: Set currentRow in PopulateMetadataTable
While looking at #1060 I noticed two things about the following code.  First I think currentRow needs to be set when a new mutation is generated.  Second I think the two if stmts `if(m != null)` and `if (!currentRow.equals(metadataRow))` could be combined into one.

https://github.com/apache/accumulo/blob/903119dac631b36009b389193d9d0bc04ed7fb37/server/master/src/main/java/org/apache/accumulo/master/tableOps/tableImport/PopulateMetadataTable.java#L164

DomGarguilo:
Closes #1091 

Improvements include:
1. Setting `currentRow` after each mutation
2. Adding a try-with-resources block
3. Consolidating two if statements


server/manager/src/main/java/org/apache/accumulo/manager/tableOps/tableImport/PopulateMetadataTable.java
```diff
@@ -91,15 +91,11 @@ static void readMappingFile(VolumeManager fs, ImportedTableInfo tableInfo, Strin
...
-    try {
-      VolumeManager fs = manager.getVolumeManager();
-
-      mbw = manager.getContext().createBatchWriter(MetadataTable.NAME, new BatchWriterConfig());
-
-      zis = new ZipInputStream(fs.open(path));
+    try (
+        BatchWriter mbw =
+            manager.getContext().createBatchWriter(MetadataTable.NAME, new BatchWriterConfig());
+        VolumeManager fs = manager.getVolumeManager();
```
DomGarguilo, line 97: [Recent discussion in another ticket](https://github.com/apache/accumulo/pull/2117#discussion_r639859548) makes me think this probably should be moved outside of the try-with-resources block so it is not closed.

Manno15: Possible. How much have you looked into it? I noticed in the test `SortedLogRecoveryTest`, it has the VolumeManager creation in a try with resources block but that is only using the local fs for testing purposes. I do not see anywhere else in the code where that is done, however.

Manno15 (approved):
Everything else looks fine.

server/manager/src/main/java/org/apache/accumulo/manager/tableOps/tableImport/PopulateMetadataTable.java
DomGarguilo: Im looking into it now. In this case it looks like it gets the VolumeManager from Manager which is probably needed elsewhere. Also, VolumeManager.close() has a comment saying: 

> close the underlying FileSystems

Which leads  me to believe it will close more than intended.

keith-turner: I don't think this should be closed, I think its a shared object in the master that would be used by other threads and task.

Manno15: Yeah, after looking into it more, I agree. I see that the Tserver will call the close function for the filesystem when it gets shut down.

DomGarguilo: Agreed. Made this change in the most recent commit, 3aa1bdf.

server/manager/src/main/java/org/apache/accumulo/manager/tableOps/tableImport/PopulateMetadataTable.java
```diff
@@ -121,63 +114,56 @@ static void readMappingFile(VolumeManager fs, ImportedTableInfo tableInfo, Strin
           Text currentRow = null;
           int dirCount = 0;
 
-          while (true) {
-            key.readFields(in);
-            val.readFields(in);
+          try (DataInputStream in = new DataInputStream(new BufferedInputStream(zis))) {
```
keith-turner, line 117: Seems like this could close zis, which would not be desired inside the loop.    I am uncertain about the actual behavior though.

keith-turner (approved):

ctubbsii (approved):
@DomGarguilo We were both fixing the conflict at the same time. You beat me to it, though :) This can be merged once the QA checks pass.

PR #2119, opened 2021-05-26
Fixes bug with -a server option

keith-turner:
The -a option for server processes was being ignored because the impl read the
argument before arguments were parsed.


keith-turner:
Found this while testing #2096.  I think this may be a problem in 2.0

Manno15 (approved):

brianloss (approved):

dlmarion:
I'm going to merge this so that I can update the external compaction branch from `main`.

PR #2123, opened 2021-05-28
Improve SuspendedTabletsIT reliability

### LINKED ISSUE #2121: NoSuchMethodException using HostRegexTableLoadBalancer
The runtime class loading of the `HostRegexTableLoadBalancer` is broken. It is likely this happened when it was refactored during the move to SPI. This exception can be seen in the manager log when running `SuspendedTabletsIT`, which seems to timeout often for me locally but the test can still pass even with this exception.
<pre>
2021-05-27T10:39:11,473 [balancer.TableLoadBalancer] WARN : Failed to load table balancer class org.apache.accumulo.core.spi.balancer.HostRegexTableLoadBalancer for table !0
java.lang.NoSuchMethodException: org.apache.accumulo.core.spi.balancer.HostRegexTableLoadBalancer.<init>(org.apache.accumulo.core.data.TableId)
        at java.lang.Class.getConstructor0(Class.java:3349) ~[?:?]
        at java.lang.Class.getConstructor(Class.java:2151) ~[?:?]
        at org.apache.accumulo.core.spi.balancer.TableLoadBalancer.constructNewBalancerForTable(TableLoadBalancer.java:58) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.core.spi.balancer.TableLoadBalancer.getBalancerForTable(TableLoadBalancer.java:91) [accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.core.spi.balancer.TableLoadBalancer.getAssignments(TableLoadBalancer.java:115) [accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.manager.Manager.getAssignments(Manager.java:1735) [accumulo-manager-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.manager.TabletGroupWatcher.getAssignmentsFromBalancer(TabletGroupWatcher.java:880) [accumulo-manager-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.manager.TabletGroupWatcher.flushChanges(TabletGroupWatcher.java:911) [accumulo-manager-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.manager.TabletGroupWatcher.run(TabletGroupWatcher.java:318) [accumulo-manager-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
2021-05-27T10:39:11,474 [balancer.TableLoadBalancer] INFO : Using balancer org.apache.accumulo.core.spi.balancer.SimpleLoadBalancer for table !0
</pre>

brianloss:
The previous attempt to fix SuspendedTabletsIT attempted to use
HostRegexTableLoadBalancer as a per-table balancer. Unfortunately,
HostRegexTableLoadBalancer was not designed to be used that way but
instead was meant to replace TableLoadBalancer and use regular
expressions to divide the hosts into pools. Therefore, the balancer was
failing to initialize and wasn't actually doing any work.

* Performed minor refactoring on HostRegexTableLoadBalancer so that it
  can be extended to include the port when performing regex checks.
* Extend HostRegexTableLoadBalancer and configure it to use the port as
  well as host name for regex checks so that a single tserver on a host
  with multiple tservers running can be isolated and used for a table.
  In particular, this is used to send all metadata table tablets to a
  single tserver running in MiniAccumuloCluster.
* Start MiniAccumuloCluster with a single tserver so that we can use
  that information to configure the extended HostRegexTableLoadBalancer
  to send all metadata tablets to that first tablet server. With one
  tablet server running, we can also grab the process reference to it
  for when we later kill tablet servers (to ensure we don't kill the
  tablet server hosting the metadata table). Once this information is
  collected, then MiniAccumuloCluster is re-configured for 3 tablet
  servers and the additional tablet servers are launched.
* Fix the "Waiting on hosting and balance" conditions to wait for BOTH
  all tablets to be hosted and for those tablets to be hosted across all
  tablet servers (except the one reserved for the metadata table)
* Fix the "Waiting on suspended tablets" condition which was incorrectly
  using a conjunction when a disjunction was needed. The result was the
  loop could exit without any tablets actually reaching the suspended
  state.
* Refactor TabletLocations so we can choose to scan either the root or
  metadata table. This allows us to retrieve tablet locations for the
  metadata table.

Fixes #2121

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/manager/SuspendedTabletsIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/manager/SuspendedTabletsIT.java
@@ -23,8 +23,10 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 
+import java.net.UnknownHostException;
 import java.security.SecureRandom;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -40,9 +42,11 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.FutureTask;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
 
 import org.apache.accumulo.core.client.Accumulo;
 import org.apache.accumulo.core.client.AccumuloClient;
+import org.apache.accumulo.core.client.admin.InstanceOperations;
 import org.apache.accumulo.core.client.admin.NewTableConfiguration;
 import org.apache.accumulo.core.clientImpl.ClientContext;
 import org.apache.accumulo.core.clientImpl.ManagerClient;
@@ -52,9 +56,11 @@ import org.apache.accumulo.core.conf.Property;
 import org.apache.accumulo.core.data.Range;
 import org.apache.accumulo.core.dataImpl.KeyExtent;
 import org.apache.accumulo.core.metadata.MetadataTable;
+import org.apache.accumulo.core.metadata.RootTable;
 import org.apache.accumulo.core.metadata.TServerInstance;
 import org.apache.accumulo.core.metadata.TabletLocationState;
 import org.apache.accumulo.core.spi.balancer.HostRegexTableLoadBalancer;
+import org.apache.accumulo.core.spi.balancer.data.TabletServerId;
 import org.apache.accumulo.core.util.HostAndPort;
 import org.apache.accumulo.minicluster.ServerType;
 import org.apache.accumulo.miniclusterImpl.MiniAccumuloConfigImpl;
@@ -64,6 +70,7 @@ import org.apache.accumulo.test.functional.ConfigurableMacBase;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.Text;
 import org.junit.AfterClass;
+import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 import org.slf4j.Logger;
@@ -81,6 +88,8 @@ public class SuspendedTabletsIT extends ConfigurableMacBase {
   public static final long SUSPEND_DURATION = 80;
   public static final int TABLETS = 30;
 
+  private ProcessReference metadataTserverProcess;
+
   @Override
   protected int defaultTimeoutSeconds() {
     return 5 * 60;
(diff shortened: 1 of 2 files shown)


Manno15:
Tests are passing locally. I will take a closer look at the code now.

test/src/main/java/org/apache/accumulo/test/manager/SuspendedTabletsIT.java
```diff
@@ -140,6 +193,7 @@ public void shutdownAndResumeTserver() throws Exception {
       }
 
       // remove servers with metadata on them from the list of servers to be shutdown
+      assertEquals("Expecting a single tServer in metadataServerSet", 1, metadataServerSet.size());
       tserverSet.removeAll(metadataServerSet);
 
       assertEquals("Expecting two tServers in shutdown-list", 2, tserverSet.size());
```
DomGarguilo, line 199:
```suggestion
      assertEquals("Expecting " + (TSERVERS - 1) + " tServers in shutdown-list", TSERVERS - 1, tserverSet.size());
```

DomGarguilo, line 199: Could add another assert around this line to ensure `tserverSet.size() >= count`

test/src/main/java/org/apache/accumulo/test/manager/SuspendedTabletsIT.java
```diff
@@ -285,6 +341,48 @@ public static void cleanup() {
...
   }
 
+  /**
+   * A version of {@link HostRegexTableLoadBalancer} that includes the tablet server port in
+   * addition to the host name when checking regular expressions. This is useful for testing when
+   * multiple tablet servers are running on the same host and one wishes to make pools from the
+   * tablet servers on that host.
+   */
+  public static class HostAndPortRegexTableLoadBalancer extends HostRegexTableLoadBalancer {
+    private static Logger LOG =
```
DomGarguilo, line 351:
```suggestion
    private static final Logger LOG =
```

DomGarguilo (commented):
These look like some solid improvements. Tests pass locally. Just a few minor suggestions/discussion points and everything else looks good to me.

test/src/main/java/org/apache/accumulo/test/manager/SuspendedTabletsIT.java
brianloss: This isn't a line I changed, but I did notice that the value 2 is hard-coded a bunch of places in the test. If I'm going to fix one, then they should probably all be fixed.

test/src/main/java/org/apache/accumulo/test/manager/SuspendedTabletsIT.java
```diff
@@ -91,21 +100,65 @@ public void configure(MiniAccumuloConfigImpl cfg, Configuration fsConf) {
...
   public void crashAndResumeTserver() throws Exception {
     // Run the test body. When we get to the point where we need a tserver to go away, get rid of it
     // via crashing
     suspensionTestBody((ctx, locs, count) -> {
-      List<ProcessReference> procs =
-          new ArrayList<>(getCluster().getProcesses().get(ServerType.TABLET_SERVER));
+      // Exclude the tablet server hosting the metadata table from the list and only
+      // kill tablet servers that are not hosting the metadata table.
+      List<ProcessReference> procs = getCluster().getProcesses().get(ServerType.TABLET_SERVER)
+          .stream().filter(p -> !metadataTserverProcess.equals(p)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
brianloss: Done

DomGarguilo (approved):

Manno15 (approved):

PR #2132, opened 2021-06-02
Make DeadCompactionDetector handle network hiccups

### LINKED ISSUE #2125: Make detection of dead external compactions more resilient to network hiccups
The code that detects dead compactions [reaches out to each compactor to ask what its running](https://github.com/apache/accumulo/blob/8a636a3ba91f5dae1d8b09b095178889a7d79c1d/server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java#L83-L84).  If there are transient network issues and a compactor can not be reached, then this could result in marking running external compactions as failed resulting in lost work.

dlmarion:
Modified the DeadCompactionDetector to fail compactions if
they are dead for more than two cycles. This should handle
the case where there is a transient network issue talking
to another component.

Closes #2125

### DIFF
--- a/server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
+++ b/server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
@@ -21,6 +21,7 @@ package org.apache.accumulo.coordinator;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 
@@ -41,13 +42,16 @@ public class DeadCompactionDetector {
 
   private final ServerContext context;
   private final CompactionCoordinator coordinator;
-  private ScheduledThreadPoolExecutor schedExecutor;
+  private final ScheduledThreadPoolExecutor schedExecutor;
+  private final ConcurrentHashMap<ExternalCompactionId,Long> deadCompactions;
+  private long threshold;
 
   public DeadCompactionDetector(ServerContext context, CompactionCoordinator coordinator,
       ScheduledThreadPoolExecutor stpe) {
     this.context = context;
     this.coordinator = coordinator;
     this.schedExecutor = stpe;
+    this.deadCompactions = new ConcurrentHashMap<>();
   }
 
   private void detectDeadCompactions() {
@@ -67,6 +71,9 @@ public class DeadCompactionDetector {
         });
 
     if (tabletCompactions.isEmpty()) {
+      // Clear out dead compactions, tservers don't think anything is running
+      log.trace("Clearing the dead compaction map, no tablets have compactions running");
+      this.deadCompactions.clear();
       // no need to look for dead compactions when tablets don't have anything recorded as running
       return;
     }
@@ -74,6 +81,17 @@ public class DeadCompactionDetector {
     if (log.isTraceEnabled()) {
       tabletCompactions.forEach((ecid, extent) -> log.trace("Saw {} for {}", ecid, extent));
     }
+
+    // Remove from the dead map any compactions that the Tablet's
+    // do not think are running any more.
+    this.deadCompactions.keySet().forEach(eci -> {
+      if (!tabletCompactions.containsKey(eci)) {
+        if (this.deadCompactions.remove(eci) != null)
+          log.trace(
+              "Removed {} from the dead compaction map, no tablet thinks this compaction is running",
+              eci);
+      }
+    });
 
     // Determine what compactions are currently running and remove those.
     //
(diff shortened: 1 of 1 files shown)


server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
```diff
@@ -75,6 +82,17 @@ private void detectDeadCompactions() {
...
+    // Remove from the dead map any compactions that the Tablet's
+    // do not think are running any more.
+    this.deadCompactions.keySet().forEach(eci -> {
+      if (!tabletCompactions.containsKey(eci)) {
+        if (this.deadCompactions.remove(eci) != null)
+          log.trace(
+              "Removed {} from the dead compaction map, no tablet thinks this compaction is running",
+              eci);
+      }
+    });
```
keith-turner, line 94: I think the following does the same thing.

```suggestion
    this.deadCompactions.keySet().retailAll(tabletCompactions.keySet());
```

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
```diff
@@ -85,31 +103,55 @@ private void detectDeadCompactions() {
...
+          if (this.deadCompactions.remove(ecid) != null) {
+            log.trace("Removed {} from the dead compaction map, it's committing", ecid);
+          }
+        });
 
-    tabletCompactions
-        .forEach((ecid, extent) -> log.debug("Detected dead compaction {} {}", ecid, extent));
+    tabletCompactions.forEach((ecid, extent) -> {
+      log.debug("Possible dead compaction detected {} {}", ecid, extent);
+      this.deadCompactions.putIfAbsent(ecid, System.currentTimeMillis());
```
keith-turner, line 126: Using a count instead of time may simplify code.  Not sure if the following is correct, I think it may increment a count or set it to 1 if it does not exists.

```suggestion
      this.deadCompactions.merge(ecid, 1, Long::sum);
```

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
```diff
@@ -85,31 +103,55 @@ private void detectDeadCompactions() {
...
+          log.warn(
+              "Failing compaction {} which is believed to be dead. Last seen at {} and not seen since.",
+              eci, startTime);
+          coordinator.compactionFailed(tabletCompactions);
+          this.deadCompactions.remove(eci);
+        } catch (UnknownCompactionIdException e) {
+          // One or more Ids was not in the Running compaction list. This is ok to ignore.
+        }
+      }
+    });
```
keith-turner, line 146: It would be good to only call `coordinator.compactionFailed()` once because it uses a batch writer in the impl instead of per entry.  The following may do this and uses a count.

```suggestion
    var compactionsToFail = deadCompactions.entrySet().stream().filter(e -> e.getValue() >= 2).map(e -> e.getKey()).collect(toSet());
   tabletCompactions.keySet().retainAll(compactionsToFail);
   coordinator.compactionFailed(tabletCompactions);
   deadCompactions.keySet().removeAll(compactionsToFail);
```

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: thanks. That's much simpler.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: It looks like it will set it to 1 if  it doesn't exist, I'm not sure if it will increment or not based on the javadoc.

dlmarion: It works.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: I applied this change in 95e8fec149722ac3f205734103223e3f42b33438

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: I applied this change in 95e8fec149722ac3f205734103223e3f42b33438

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: I applied this change in 95e8fec149722ac3f205734103223e3f42b33438

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
```diff
@@ -84,21 +97,42 @@ private void detectDeadCompactions() {
...
 
     // Everything left in tabletCompactions is no longer running anywhere and should be failed.
     // Its possible that a compaction committed while going through the steps above, if so then
     // that is ok and marking it failed will end up being a no-op.
+    Set<ExternalCompactionId> toFail =
+        this.deadCompactions.entrySet().stream().filter(e -> e.getValue() > 2).map(e -> e.getKey())
+            .collect(Collectors.toCollection(TreeSet::new));
+    tabletCompactions.keySet().retainAll(toFail);
+    tabletCompactions.forEach((eci, v) -> {
+      log.warn("Compaction {} believed to be dead, failing it.", eci);
```
keith-turner, line 132: Maybe make this info.  Thinking that compactor processes dying will be somewhat normal/expected so may not want to warn.

```suggestion
      log.info("Compaction {} believed to be dead, failing it.", eci);
```

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
```diff
@@ -84,21 +97,42 @@ private void detectDeadCompactions() {
...
+          if (this.deadCompactions.remove(ecid) != null) {
+            log.trace("Removed {} from the dead compaction map, it's committing", ecid);
+          }
+        });
 
-    tabletCompactions
-        .forEach((ecid, extent) -> log.debug("Detected dead compaction {} {}", ecid, extent));
+    tabletCompactions.forEach((ecid, extent) -> {
+      log.debug("Possible dead compaction detected {} {}", ecid, extent);
+      this.deadCompactions.putIfAbsent(ecid, System.currentTimeMillis());
```
keith-turner, line 120: Think this line can be removed

```suggestion
```

keith-turner (approved):

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: I think compactor processes dying might only be normal in the situation where you are dynamically scaling and the scaling mechanism has no way to gracefully stop the compactor process. In a static deployment, compactor processes should not be dying. I think `warn` might be appropriate as it's highlighting a failure. Curious what others have to say here.

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
dlmarion: I thought I had removed that. Doh! removed in e95b6c46063bb5bfb9e1a252c8502a47818b2a49

server/compaction-coordinator/src/main/java/org/apache/accumulo/coordinator/DeadCompactionDetector.java
keith-turner: Yeah I can see that case for making it warn.  I feel it should be INFO or WARN, but not sure which. WARN is fine w/ me.

EdColeman: Why would that be expected?  If something is repeatedly dying I'd think that would indicate that some action / investigation might be warranted.  If I had a compaction that was taking a long time - and it was because these kept dying, how would I know?

Recently had a discussion  related to the trace level statements that are associated with this particular log message, wondering if they might not be better at debug (at least for some of them) - my impression is that dead / dying and removing the entries should not be common - and if they are then something should be fixed so they are not.   It might not be obvious that if you are having issues with external compaction you need to set logging to trace to obtain enough info to figure out what the issue might be.  I realize that we often log to much, but from looking at the code without understanding the frequency / how common / what a corrective approach might be - some of this code might benefit from more logging at debug or higher levels - just my initial impression, would probability need to run and then have to actually try to debug something before having something concrete to add to the discussion.

keith-turner: > Why would that be expected?

I was thinking that on a large cluster, host and processes dying is expected.  In general I think Accumulo has a bit too much logging at the warn level for things that occur as a routine part of operating a cluster.  If so, maybe this makes people ignore warnings.

>  It might not be obvious that if you are having issues with external compaction you need to set logging to trace to obtain enough info to figure out what the issue might be.

I was thinking that detection of a dead compaction should be at WARN or INFO, but was not sure which after @dlmarion comment.  The stuff at trace is related to the source information on how it finds dead compactions and will routinely log information even when it does not detect a dead compaction.

keith-turner: I like how the change logs at debug when it finds a candidate dead compaction and adds that to the map.

PR #2134, opened 2021-06-02
Converting to correct use of ClientContext.createBatchWriter()

DomGarguilo:
There are two createBatchWriter() methods in ClientContext.
```java
public BatchWriter createBatchWriter(String tableName, BatchWriterConfig config)
```
and
```java
public BatchWriter createBatchWriter(String tableName)
```
The latter just uses the former and creates a `new BatchWriterConfig` as its config param. This PR is to convert the cases where the simplified method can be used.

Additionally there were a few cases of `createBatchDeleter` that were converted for the same reason. I also moved a few cases into a try-with-resources block to make use of their autoclosable implementation.


dlmarion:
Looks good.

test/src/main/java/org/apache/accumulo/test/replication/GarbageCollectorCommunicatesWithTServersIT.java
```diff
@@ -196,18 +196,18 @@ public void testActiveWalPrecludesClosing() throws Exception {
...
+        bw.addMutation(m);
+      }
 
-    byte[] empty = new byte[0];
-    for (int i = 0; i < 5; i++) {
-      Mutation m = new Mutation(Integer.toString(i));
-      m.put(empty, empty, empty);
-      bw.addMutation(m);
+      log.info("Flushing mutations to the server");
+      bw.flush();
```
keith-turner, line 208: I think flushing the batch writer could be dropped now that its closed. I think closing the batch writer will flush it.

DomGarguilo: Included in the most recent commit.

ctubbsii (approved):

PR #2135, opened 2021-06-02
Test converting client props to ConditionalWriterConfig

### LINKED ISSUE #2131: Unit test converting client props to ConditionalWriterConfig
Issue #2115 requested that unit tests be added to test that BatchWriterConfig was correctly created from client props. Very similar tests could be added for ConditionalWriterConfig to test it as well. The changes in PR #2126 should be very similar to the tests for ConditionalWriterConfig.

jmark99:
Created two unit tests for the conversion of client properties to ConditionalWriterConfig. One using default values and another using modified values.

Refactored getConditionalWriterConfig into two methods to facilitate unit testing.

Closes #2131

### DIFF
--- a/core/src/test/java/org/apache/accumulo/core/clientImpl/ClientContextTest.java
+++ b/core/src/test/java/org/apache/accumulo/core/clientImpl/ClientContextTest.java
@@ -30,6 +30,7 @@ import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.accumulo.core.client.BatchWriterConfig;
+import org.apache.accumulo.core.client.ConditionalWriterConfig;
 import org.apache.accumulo.core.client.Durability;
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
 import org.apache.accumulo.core.conf.ClientProperty;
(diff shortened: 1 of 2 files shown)


core/src/test/java/org/apache/accumulo/core/clientImpl/ClientContextTest.java
```diff
@@ -170,4 +171,58 @@ public void testGetBatchWriterConfigNotUsingDefaults() {
...
   }
 
+  @Test
+  public void testGetConditionalWriterConfigUsingDefaults() {
+    Properties props = new Properties();
+    ConditionalWriterConfig conditionalWriterConfig =
+        ClientContext.getConditionalWriterConfig(props);
+    assertNotNull(conditionalWriterConfig);
+
+    // If the value of BATCH_WRITER_TIMEOUT_MAX is set to zero, Long.MAX_VALUE is returned.
```
DomGarguilo, line 181:
```suggestion
    // If the value of CONDITIONAL_WRITER_TIMEOUT_MAX is set to zero, Long.MAX_VALUE is returned.
```

DomGarguilo: There is a second occurrence in this comment that needs to be changed as well but there isn't a good way to suggest a multi-line change here because the extra length pushes words onto the next line.

jmark99:
@DomGarguilo thanks for the catch on the comment!

EdColeman (approved):

PR #2145, opened 2021-06-04
Fix crypto in SplitLarge utility

milleruntime:


Manno15 (approved):

ctubbsii (approved):
I don't see any problems with the changes, but it's not clear what is being done here. The description merely says "Fix crypto", but doesn't say what's broken or what's being done to fix it.

milleruntime:
The option that was added in 2.0 for the encryption only included the name of the service. You also need a way to pass in other options for it to work. This change just simplifies that by removing the option altogether and reading the config from classpath, similar to other utilities.

PR #145, opened 2016-09-01
ACCUMULO-4376 add KeyBuilder

melrief:
Adds the `KeyBuilder` as proposed in ACCUMULO-4376. For the design, I took inspiration from the Step Builder design pattern as described in [Vladimir Stankovic blog](http://www.svlada.com/step-builder-pattern/) to try to give some rules to the way `Key`s are built. I also added a new constructor to `Key` that accepts all the parameters, including `deleted` similar to to the `Key.init` method, that is used to build `Key`s from the fields.

PR #2170, opened 2021-06-16
Add test and fix for bug in KeyBuilder

milleruntime:
Found this weird bug in `KeyBuilder` while working on #2117. The Java UTF-8 Charset encoder puts an empty byte at the end of the string when there is an underscore present and I can't figure out why. Opening this PR to show the test failure, if anyone has any ideas.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/data/KeyBuilder.java
+++ b/core/src/main/java/org/apache/accumulo/core/data/KeyBuilder.java
@@ -20,11 +20,6 @@ package org.apache.accumulo.core.data;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-import java.nio.CharBuffer;
-import java.nio.charset.CharacterCodingException;
-import java.nio.charset.CharsetEncoder;
-import java.nio.charset.CodingErrorAction;
-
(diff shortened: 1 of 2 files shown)


Manno15:
I will help investigate the issue.

Manno15:
From my testing (using your test), the issue appears to be more related to the length of the CharSequence than what is in the CharSequence.

milleruntime:
> From my testing (using your test), the issue appears to be more related to the length of the CharSequence than what is in the CharSequence.

Yeah it looks like 10 or more characters comes back with an extra byte. While 9 characters or less is fine.

milleruntime:
There is a simpler way to convert from `CharSequence` to byte array that I implemented it in 9c99225. I am not sure why the Java `CharsetEncoder` was used in the first place.

Manno15 (approved):
Confirmed that this fixes the issue

milleruntime:
The original PR https://github.com/apache/accumulo/pull/145 for KeyBuilder used `CharsetEncoder` as a replacement for Text to avoid going from string->text->bytes. But I think this was an oversight as we can just go from String -> bytes and avoid the `CharBuffer` and `CharsetEncoder` expanding the byte array beyond the original size of the `CharSequence`.

PR #2185, opened 2021-06-30
Handle old sorted map files in Upgrade

### LINKED ISSUE #2179: Handle old sorted map files
The changes in #2117 will cause errors if users have old sorted map files left around from before an upgrade to 2.1. I think we can handle this by catching the error or detecting the old files. Then deleting the map files and forcing Accumulo to re-sort is probably the best approach.

milleruntime:
~~* Add upgradeFiles method for upgrading files to Upgrader~~
~~* Refactor status handling in UpgradeCoordinator to allow upgradeFiles
to run concurrently and wait for metadata upgrade to complete~~
~~* Implement upgradeFiles method in Upgrader9to10~~
* Add dropSortedMapWALFiles for resolving sorted map files that may
still be around during upgrade. Follow up for #2117
* Closes #2179

### DIFF
--- a/server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
+++ b/server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
@@ -272,14 +272,12 @@ public class Manager extends AbstractServer
             + " all logs and file a bug.");
       }
       upgradeMetadataFuture = upgradeCoordinator.upgradeMetadata(getContext(), nextEvent);
-      upgradeFilesFuture = upgradeCoordinator.upgradeFiles(getContext(), nextEvent);
     }
   }
 
   private final UpgradeCoordinator upgradeCoordinator = new UpgradeCoordinator();
 
   private Future<Void> upgradeMetadataFuture;
-  private Future<Void> upgradeFilesFuture;
 
   private ManagerClientServiceHandler clientHandler;
 
@@ -1120,9 +1118,6 @@ public class Manager extends AbstractServer
       // wait for metadata upgrade running in background to complete
       if (null != upgradeMetadataFuture) {
         upgradeMetadataFuture.get();
-      }
-      if (null != upgradeFilesFuture) {
-        upgradeFilesFuture.get();
       }
     } catch (ExecutionException | InterruptedException e) {
       throw new IllegalStateException("Metadata upgrade failed", e);
--- a/server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
+++ b/server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
@@ -272,12 +272,14 @@ public class Manager extends AbstractServer
             + " all logs and file a bug.");
       }
       upgradeMetadataFuture = upgradeCoordinator.upgradeMetadata(getContext(), nextEvent);
+      upgradeFilesFuture = upgradeCoordinator.upgradeFiles(getContext(), nextEvent);
     }
   }
 
   private final UpgradeCoordinator upgradeCoordinator = new UpgradeCoordinator();
 
   private Future<Void> upgradeMetadataFuture;
+  private Future<Void> upgradeFilesFuture;
 
   private ManagerClientServiceHandler clientHandler;
 
@@ -1118,6 +1120,9 @@ public class Manager extends AbstractServer
       // wait for metadata upgrade running in background to complete
       if (null != upgradeMetadataFuture) {
         upgradeMetadataFuture.get();
+      }
+      if (null != upgradeFilesFuture) {
+        upgradeFilesFuture.get();
       }
     } catch (ExecutionException | InterruptedException e) {
       throw new IllegalStateException("Metadata upgrade failed", e);
--- a/server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
+++ b/server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
@@ -105,6 +105,12 @@ import com.google.common.base.Preconditions;
  * <a href="https://github.com/apache/accumulo/issues/1642">#1642</a>, and
  * <a href="https://github.com/apache/accumulo/issues/1643">#1643</a> as well.</li>
  * </ul>
+ *
+ * Sorted recovery was updated to use RFiles instead of map files. So to prevent issues during
+ * tablet recovery, remove the old temporary map files and resort using RFiles. This is done in
+ * {@link #dropSortedMapWALFiles(VolumeManager)}. For more information see the following issues:
+ * <a href="https://github.com/apache/accumulo/issues/2117">#2117</a> and
+ * <a href="https://github.com/apache/accumulo/issues/2179">#2179</a>
  */
 public class Upgrader9to10 implements Upgrader {
 
@@ -142,6 +148,11 @@ public class Upgrader9to10 implements Upgrader {
     upgradeRelativePaths(ctx, Ample.DataLevel.USER);
     upgradeDirColumns(ctx, Ample.DataLevel.USER);
     upgradeFileDeletes(ctx, Ample.DataLevel.USER);
+  }
+
+  @Override
+  public void upgradeFiles(ServerContext ctx) {
+    dropSortedMapWALFiles(ctx.getVolumeManager());
   }
 
   private void setMetaTableProps(ServerContext ctx) {
@@ -725,5 +736,38 @@ public class Upgrader9to10 implements Upgrader {
           "Missing required property " + Property.INSTANCE_VOLUMES_UPGRADE_RELATIVE.getKey());
     }
     return new Path(upgradeProperty, VolumeManager.FileType.TABLE.getDirectory() + oldDelete);
+  }
+
+  /**
+   * Remove old temporary map files to prevent problems during recovery.
+   */
+  static void dropSortedMapWALFiles(VolumeManager vm) {
+    Path recoveryDir = new Path("/accumulo/recovery");
+    try {
+      if (!vm.exists(recoveryDir)) {
+        log.info("There are no recover files in /accumulo/recovery");
+        return;
+      }
+      List<Path> directoriesToDrop = new ArrayList<>();
+      for (FileStatus walDir : vm.listStatus(recoveryDir)) {
+        // map files will be in a directory starting with "part"
+        Path walDirPath = walDir.getPath();
+        for (FileStatus dirOrFile : vm.listStatus(walDirPath)) {
+          if (dirOrFile.isDirectory()) {
+            directoriesToDrop.add(walDirPath);
+            break;
+          }
+        }
+      }
+      if (!directoriesToDrop.isEmpty()) {
+        log.info("Found {} old sorted map directories to delete.", directoriesToDrop.size());
+        for (Path dir : directoriesToDrop) {
+          log.info("Deleting everything in old sorted map directory: {}", dir);
+          vm.deleteRecursively(dir);
+        }
+      }
+    } catch (IOException ioe) {
+      throw new UncheckedIOException(ioe);
+    }
   }
 }
--- a/server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
+++ b/server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
@@ -745,7 +745,7 @@ public class Upgrader9to10 implements Upgrader {
     Path recoveryDir = new Path("/accumulo/recovery");
     try {
       if (!vm.exists(recoveryDir)) {
-        log.info("There are no recover files in /accumulo/recovery");
+        log.info("There are no recovery files in /accumulo/recovery");
         return;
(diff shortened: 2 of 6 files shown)


Manno15 (approved):
Everything looks good. Noticed one small typo but that is about it.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -726,4 +737,37 @@ static Path resolveRelativeDelete(String oldDelete, String upgradeProperty) {
...
   }
+
+  /**
+   * Remove old temporary map files to prevent problems during recovery.
+   */
+  static void dropSortedMapWALFiles(VolumeManager vm) {
+    Path recoveryDir = new Path("/accumulo/recovery");
+    try {
+      if (!vm.exists(recoveryDir)) {
+        log.info("There are no recover files in /accumulo/recovery");
```
Manno15, line 748: Small typo. Recover should be recovery*

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -144,6 +150,11 @@ public void upgradeMetadata(ServerContext ctx) {
     upgradeFileDeletes(ctx, Ample.DataLevel.USER);
   }
 
+  @Override
+  public void upgradeFiles(ServerContext ctx) {
+    dropSortedMapWALFiles(ctx.getVolumeManager());
```
ctubbsii, line 155: What happens when we delete these files? Wouldn't sorted files have references in the metadata tables? And, if so, would their original WAL files have already been subject to deletion? Is it safe to just remove these and Accumulo will just recreate them using the new format? That behavior isn't clear to me.

server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
```diff
@@ -1120,6 +1122,9 @@ boolean canSuspendTablets() {
       if (null != upgradeMetadataFuture) {
         upgradeMetadataFuture.get();
       }
+      if (null != upgradeFilesFuture) {
+        upgradeFilesFuture.get();
+      }
```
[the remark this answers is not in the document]
milleruntime: Correct, it doesn't matter in this case. But I did have the same concerns about future problems with the upgrader. Since I didn't have a concrete example, I wasn't sure when `upgradeFiles()` should get called. I ended up settling with this concurrent upgrade since that works for the current changes. 

Any thoughts on where in the upgrade order the files would ideally get upgraded? I was originally thinking last, after the metadata. Maybe the files should be done first before anything (but after ZK?) so we know what paths to store? Like I said, I couldn't think of any specific upgrade example where files were dependent on other changes.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
milleruntime: > What happens when we delete these files? 

Accumulo will automatically re-sort the WALs based off of the original files in `/accumulo/wal`.

> Wouldn't sorted files have references in the metadata tables? 

No they do not. The references are stored in ZK and hdfs. The `DistributedWorkQueue` runs the `LogSorter` based on data in the ZK node. Each node is populated based on the WAL filename/ID taken from `/accumulo/wal`. This method only deletes the temporary sorted files in `/accumulo/recovery`.

> Is it safe to just remove these and Accumulo will just recreate them using the new format? 

Yes that is the intent. Without this change, Accumulo won't be able to recovery tablets in the old map files. If it happens that metadata tablets are in these old sorted map files, then Accumulo won't be able to start.

milleruntime: Here is some sample logging that might help:
<pre>
manager_ip-10-113-12-25.log:2021-07-08T09:03:05,079 [recovery.RecoveryManager] INFO : Created zookeeper entry /accumulo/5012a7fa-5daa-45fc-92ca-1c789e70d4f7/recovery/13315719-9dee-460e-92ab-7a3276d34a26 with data hdfs://localhost:8020/accumulo/wal/ip-10-113-12-25+9997/13315719-9dee-460e-92ab-7a3276d34a26|hdfs://localhost:8020/accumulo/recovery/13315719-9dee-460e-92ab-7a3276d34a26
manager_ip-10-113-12-25.log:2021-07-08T09:03:06,693 [recovery.RecoveryManager] INFO : Starting recovery of hdfs://localhost:8020/accumulo/wal/ip-10-113-12-25+9997/13315719-9dee-460e-92ab-7a3276d34a26 (in : 20s), tablet +r<< holds a reference
tserver_ip-10-113-12-25.err:2021-07-08T09:03:05,093 [log.LogSorter] INFO : Copying hdfs://localhost:8020/accumulo/wal/ip-10-113-12-25+9997/13315719-9dee-460e-92ab-7a3276d34a26 to hdfs://localhost:8020/accumulo/recovery/13315719-9dee-460e-92ab-7a3276d34a26
tserver_ip-10-113-12-25.err:2021-07-08T09:03:06,587 [log.LogSorter] INFO : Finished log sort 13315719-9dee-460e-92ab-7a3276d34a26 119262720 bytes 2 parts in 1501ms
</pre>

server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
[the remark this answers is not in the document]
ctubbsii: Well, usually we don't upgrade files at all, because the number of files tends to be excessive and not scalable. So, we support upgrading files over time in a distributed way, by reading multiple older versions, and writing only newer versions. See RFile code. I think we even still have some remaining code for reading MapFiles (which we used before RFile existed). In general, we don't want to upgrade files at all at this phase, certainly not isolated to a single machine, when we could have thousands and thousands of files across HDFS.

In this case, I was concerned that these files being dropped depend on metadata changes... since there could be references to these sorted WAL files in the metadata. That would be one example. In general, though, we just don't upgrade files at this point at all.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
ctubbsii: Oh, I thought we put references to them in the "log" column family in the metadata table. Looking at the code, it looks like we use that column family for storing references to unsorted logs instead, when a tablet is unassigned from a dead tserver. I knew sorted WALs were short-lived, but I didn't expect them to be not persisted in the metadata.

server/manager/src/main/java/org/apache/accumulo/manager/Manager.java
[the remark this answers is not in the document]
milleruntime: Since we don't upgrade files, should this fix be made somewhere else? I didn't want to clutter the recovery code for many versions to handle deleting the files one time.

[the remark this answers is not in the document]
ctubbsii: I think it's probably okay to upgrade the files at manager startup through this mechanism. The part I'm concerned about is changing the upgrade interface to provide a general mechanism to upgrade files, as that could encourage misuse. Maybe we can just consider these part of the "metadata" and perform the upgrade there instead of add the new "upgradeFiles" stuff?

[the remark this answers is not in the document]
milleruntime: I reverted the changes I made adding a new `upgradeFiles` method to the Upgrade interface and now this change is much simpler.

Manno15 (approved):
Everything looks good from my perspective. Have you done any testing to verify that Accumulo will automatically re-sort the WALs once these are deleted? Any way to create a test for that?

milleruntime:
> Everything looks good from my perspective. Have you done any testing to verify that Accumulo will automatically re-sort the WALs once these are deleted? Any way to create a test for that?

Yeah I did some manual testing with Uno. It may be possible to create an IT with the old files saved off but I imagine it would be quite involved. I am not sure its worth the work for a one time thing that will only ever happen between versions.

milleruntime:
@ctubbsii I reverted the changes I made adding a new upgradeFiles method to the Upgrade interface and now this change is much simpler. I believe I got rid of the changes you had concern over. This PR was just the new code I added to fix old sorted logs.

milleruntime:
So I did so more testing and while it successfully deleted the old data, I am not 100% sure that it won't delete WALs that still have referenced data in them. This is a little tricky to test manually, as you need to make sure the table is recovering but also have unloaded tablets (before the data is flushed to disk). I am going to open a ticket to test this using Continuous Ingest.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -726,4 +734,37 @@ static Path resolveRelativeDelete(String oldDelete, String upgradeProperty) {
     }
     return new Path(upgradeProperty, VolumeManager.FileType.TABLE.getDirectory() + oldDelete);
   }
+
+  /**
+   * Remove old temporary map files to prevent problems during recovery.
+   */
+  static void dropSortedMapWALFiles(VolumeManager vm) {
+    Path recoveryDir = new Path("/accumulo/recovery");
```
keith-turner, line 742: Seems like this should loop over the set of configured volumes instead of this.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -142,6 +148,8 @@ public void upgradeMetadata(ServerContext ctx) {
     upgradeRelativePaths(ctx, Ample.DataLevel.USER);
     upgradeDirColumns(ctx, Ample.DataLevel.USER);
     upgradeFileDeletes(ctx, Ample.DataLevel.USER);
+    // special case where old files need to be deleted
```
keith-turner, line 151: Did you consider calling this new method in upgradeZookeeper instead?  I think that is called before the root tablet is loaded, which would allow deleting any old sorted logs the root tablet may reference.

milleruntime: No, I didn't think any tables were loaded until after the Upgrader was finished. If that is the case, then moving it to `upgradeZookeeper` would be better.

keith-turner: I think the upgrade methods are run as follows.

 * `upgradeZookeeper()` is run before the root tablet is loaded.
 * `upgradeRoot()` is run after the root tablet is loaded and before the metadata table is loaded.  
 * `upgradeMetadata` is run after the metadata table is loaded and before loading user tablets.

milleruntime: Ah OK. I will move it to `upgradeZookeeper()`.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
milleruntime: Do you think using the volumes configured in `instance.volumes` will be enough?

keith-turner: Yeah, could look through all of those.  Could narrow it by calling : https://github.com/apache/accumulo/blob/a9100ac35f3d72677b767d7670dfb032c7e7d733/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java#L180

But I don't think it hurts to just look through all volumes.  Maybe could use https://github.com/apache/accumulo/blob/a9100ac35f3d72677b767d7670dfb032c7e7d733/server/base/src/main/java/org/apache/accumulo/server/fs/VolumeManager.java#L196 to get the list of volumes

keith-turner: Would probably be best to avoid choosable() as that will give the volumes configured for new WALs.  The config could change and an old WAL could be on a volume that choosable() no longer returns.  So probably best to inspect all volumes looking for old sorted wals to nuke.

ctubbsii: > Would probably be best to avoid choosable() as that will give the volumes configured for new WALs. The config could change and an old WAL could be on a volume that choosable() no longer returns. So probably best to inspect all volumes looking for old sorted wals to nuke.

Definitely avoid this. Consider RandomVolumeChooser :smiley_cat:

milleruntime: I thought the VolumeChooser was for selecting volumes for writes? I want to loop through every volume configured to find the old sorted WALs to remove. Will the VolumeChooser allow for this?

PR #2186, opened 2021-07-01
Refactor common table property validation to reduce duplication

EdColeman:
- Move duplcated code for table property validation to one place
- Minor check-style fixes.
- fix possible nulll pointer warning (FateServiceHandler)

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/conf/Property.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/Property.java
@@ -1352,10 +1352,10 @@ public enum Property {
     return false;
   }
 
-  private static final HashSet<String> validTableProperties;
-  private static final HashSet<String> validProperties;
-  private static final HashSet<String> validPrefixes;
-  private static final HashMap<String,Property> propertiesByKey;
+  private static final HashSet<String> validTableProperties = new HashSet<>();
+  private static final HashSet<String> validProperties = new HashSet<>();
+  private static final HashSet<String> validPrefixes = new HashSet<>();
+  private static final HashMap<String,Property> propertiesByKey = new HashMap<>();
 
   private static boolean isKeyValidlyPrefixed(String key) {
     for (String prefix : validPrefixes) {
@@ -1541,25 +1541,23 @@ public enum Property {
   }
 
   static {
+
     // Precomputing information here avoids :
     // * Computing it each time a method is called
     // * Using synch to compute the first time a method is called
-    propertiesByKey = new HashMap<>();
-    validPrefixes = new HashSet<>();
-    validProperties = new HashSet<>();
-    validTableProperties = new HashSet<>();
 
     for (Property p : Property.values()) {
       propertiesByKey.put(p.getKey(), p);
       if (p.getType().equals(PropertyType.PREFIX)) {
         validPrefixes.add(p.getKey());
       } else {
-
         validProperties.add(p.getKey());
-
-        if (p.getKey().startsWith(Property.TABLE_PREFIX.getKey())) {
-          validTableProperties.add(p.getKey());
-        }
+      }
+      // exclude prefix types (avoids being able to set things like table.custom or
+      // table.constraint)
+      if (!p.getType().equals(PropertyType.PREFIX)
+          && p.getKey().startsWith(Property.TABLE_PREFIX.getKey())) {
+        validTableProperties.add(p.getKey());
       }
     }
 
--- a/core/src/main/java/org/apache/accumulo/core/conf/Property.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/Property.java
@@ -163,7 +163,8 @@ public enum Property {
   INSTANCE_RPC_SASL_ALLOWED_HOST_IMPERSONATION("instance.rpc.sasl.allowed.host.impersonation", "",
       PropertyType.STRING,
       "One-line configuration property controlling the network locations "
-          + "(hostnames) that are allowed to impersonate other users"), // Crypto-related properties
+          + "(hostnames) that are allowed to impersonate other users"),
+  // Crypto-related properties
   @Experimental
   INSTANCE_CRYPTO_PREFIX("instance.crypto.opts.", null, PropertyType.PREFIX,
       "Properties related to on-disk file encryption."),
@@ -220,22 +221,8 @@ public enum Property {
   GENERAL_MAX_MESSAGE_SIZE("general.server.message.size.max", "1G", PropertyType.BYTES,
       "The maximum size of a message that can be sent to a server."),
   GENERAL_SIMPLETIMER_THREADPOOL_SIZE("general.server.simpletimer.threadpool.size", "1",
-      PropertyType.COUNT, "The number of threads to use for " + "server-internal scheduled tasks"), // If
-                                                                                                    // you
-                                                                                                    // update
-                                                                                                    // the
-                                                                                                    // default
-                                                                                                    // type,
-                                                                                                    // be
-                                                                                                    // sure
-                                                                                                    // to
-                                                                                                    // update
-                                                                                                    // the
-                                                                                                    // default
-                                                                                                    // used
-                                                                                                    // for
-                                                                                                    // initialization
-                                                                                                    // failures
+      PropertyType.COUNT, "The number of threads to use for " + "server-internal scheduled tasks"),
+  // If you update the default type, be sure to update the default used for initialization failures
   // in VolumeManagerImpl
   @Experimental
   GENERAL_VOLUME_CHOOSER("general.volume.chooser", RandomVolumeChooser.class.getName(),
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -221,8 +220,22 @@
...
+                                                                                                    // be
+                                                                                                    // sure
+                                                                                                    // to
+                                                                                                    // update
+                                                                                                    // the
+                                                                                                    // default
+                                                                                                    // used
+                                                                                                    // for
+                                                                                                    // initialization
+                                                                                                    // failures
```
dlmarion, line 238: I'm not sure that this comment change is an improvement. Was this intended, or a formatting issue?

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -333,8 +346,9 @@
       "Maximum time manager will wait for tserver available threshold "
           + "to be reached before continuing. When set to 0 or less, will block "
           + "indefinitely. Default is 0 to block indefinitely. Only valid when tserver available "
-          + "threshold is set greater than 0. Added with version 1.10"),
-  // properties that are specific to tablet server behavior
+          + "threshold is set greater than 0. Added with version 1.10"), // properties that are
+                                                                         // specific to tablet
+                                                                         // server behavior
```
dlmarion, line 351: It seems this comment may have been meant as a separator between MANAGER and the TSERV properties below. It seems that just moving it down one line would be appropriate

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1171,120 +1201,70 @@
...
+    // * Using synch to compute the first time a method is called
+    propertiesByKey = new HashMap<>();
+    validPrefixes = new HashSet<>();
+    validProperties = new HashSet<>();
+
+    for (Property p : Property.values()) {
+      if (p.getType().equals(PropertyType.PREFIX)) {
+        validPrefixes.add(p.getKey());
+      } else {
+        validProperties.add(p.getKey());
```
dlmarion, line 1231: I think you can add the following here and remove the loop at line 1237:

```
if (p.getKey().startsWith(Property.TABLE_PREFIX.getKey())) {
    validTableProperties.add(p.getKey());
}
```

dlmarion:
I think there are some issues with the comments being reformatted in Property.java. Did the formatter do this?

EdColeman:
I’ll see if I can fix it – but I didn’t do it.  Not sure where its coming from at this point.

dlmarion (approved):

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1531,21 +1547,19 @@ public static boolean isClassProperty(String key) {
...
-    }
 
-    validTableProperties = new HashSet<>();
-    for (Property p : Property.values()) {
-      if (!p.getType().equals(PropertyType.PREFIX)
-          && p.getKey().startsWith(Property.TABLE_PREFIX.getKey())) {
-        validTableProperties.add(p.getKey());
+        if (p.getKey().startsWith(Property.TABLE_PREFIX.getKey())) {
+          validTableProperties.add(p.getKey());
+        }
```
ctubbsii, line 1562: I like that this logic was moved into the previous loop to avoid looping twice... although I don't think it will make much of a difference since it's a static initializer that will only run once and there aren't that many to loop through.

However, it's a problem that the logic has changed to no longer exclude PREFIX types that start with "table." This structure should not contain prefix types, but only actual keys. We don't want users to be able to set `table.custom.` or `table.constraint.` by themselves because these are incorrectly deemed to be valid properties when they aren't.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1531,21 +1547,19 @@ public static boolean isClassProperty(String key) {
     propertiesByKey = new HashMap<>();
     validPrefixes = new HashSet<>();
     validProperties = new HashSet<>();
+    validTableProperties = new HashSet<>();
```
ctubbsii, line 1550: These don't even need to be assigned here. These 4 lines could be deleted, if the `= new HashMap<>();` part were appended to their respective `private static final` lines instead of just removing the `= null;` and assigning them here.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1531,21 +1547,19 @@ public static boolean isClassProperty(String key) {
...
 
     for (Property p : Property.values()) {
+      propertiesByKey.put(p.getKey(), p);
       if (p.getType().equals(PropertyType.PREFIX)) {
         validPrefixes.add(p.getKey());
       } else {
+
```
ctubbsii, line 1557: Not sure how this made it in. It seems a few blank lines made their way in. It's not a big deal, but the changeset is smaller if these are omitted when not needed.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1531,21 +1547,19 @@ public static boolean isClassProperty(String key) {
     propertiesByKey = new HashMap<>();
     validPrefixes = new HashSet<>();
     validProperties = new HashSet<>();
+    validTableProperties = new HashSet<>();
 
     for (Property p : Property.values()) {
+      propertiesByKey.put(p.getKey(), p);
```
ctubbsii, line 1553: I can't tell if this got moved for aesthetics, or some other reason. Does it behave the same here as it did before, when it was below the if/else?

core/src/test/java/org/apache/accumulo/core/conf/PropertyTest.java
```diff
@@ -186,4 +188,61 @@ public void testIsValidTablePropertyKey() {
...
     assertFalse(Property.isValidTablePropertyKey("abc.def"));
   }
+
+  /**
+   * PR #2186 eliminated a loop in the computation - this test verifies the original methods and the
+   * refactor method produce the same result.
+   */
+  @Test
+  public void verifyRefactor() {
+
```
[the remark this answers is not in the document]
EdColeman: The test compares that that current code and the proposed code maps contain the same contents.  It was basically a one-off to show that the proposed code functions the same as what is there now.  No issues deleting it if the code modifications are accepted.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
EdColeman: Should be fixed in update

core/src/main/java/org/apache/accumulo/core/conf/Property.java
EdColeman: Yes - I believe that it was always being assigned - so I moved it to the top so it was easier to see that was happening.

core/src/test/java/org/apache/accumulo/core/conf/PropertyTest.java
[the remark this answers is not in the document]
ctubbsii: Okay. Just thinking that after this is merged, "proposed code" doesn't make sense anymore, because the proposed code becomes the current code. So, after this is merged, it will look like a test comparing the current implementation to a previous one that no longer exists.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1525,24 +1541,20 @@ public static boolean isClassProperty(String key) {
...
       } else {
         validProperties.add(p.getKey());
       }
-      propertiesByKey.put(p.getKey(), p);
-    }
-
-    validTableProperties = new HashSet<>();
-    for (Property p : Property.values()) {
+      // exclude prefix types (avoids being able to set things like table.custom or
+      // table.constraint)
```
ctubbsii, line 1557: Minor wording clarification (especially using "prevents" rather than "avoids" to make it clear it's important for blocking the behavior we don't want, and why).
```suggestion
      // exclude prefix types (prevents setting a prefix type, like table.custom or
      // table.constraint, directly, since they aren't valid properties on their own)
```

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -1525,24 +1541,20 @@ public static boolean isClassProperty(String key) {
   }
 
   static {
+
```
ctubbsii, line 1544: LOL, I think you removed the other blank line, but then added this one. It must be the IDE. It's fine either way. I just think it's funny.

ctubbsii (commented):
The implementation looks okay to me, but the test case should be deleted (or changed to check for desired behavior, rather than compare to non-existent prior implementation).

ctubbsii (approved):

PR #2188, opened 2021-07-02
WIP: Add micrometer metrics implementation

### LINKED ISSUE #1134: Look into using abstraction for metrics.

DomGarguilo:
This WIP PR serves to track changes made toward implementing micrometer as an alternative metrics system.

Fixes #1134


DomGarguilo:
I marked this as ready for review so I could start getting feedback on these changes. At this point I could use some suggestions on next steps or other improvements.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -243,7 +243,13 @@
...
       PropertyType.TIMEDURATION,
       "The maximum amount of time that a Scanner should wait before retrying a failed RPC"),
-
+  // experimental - micrometer properties
+  @Experimental
+  GENERAL_METRICS_CONFIGURATION_PROPERTIES_FILE("general.metrics.configuration.properties",
+      "accumulo.metrics.configuration.properties", PropertyType.STRING,
+      "The Accumulo metrics configuration properties filename. The file is used to"
+          + " configure micrometer MeterRegistries that loaded at runtime using classes that"
+          + " extend MetricsServiceLoader.class"),
```
foster33, line 252: After the most recent changes in #2190, properties will need `since` information. I figured I would leave a comment regarding it.

```suggestion
  @Experimental
  GENERAL_METRICS_CONFIGURATION_PROPERTIES_FILE("general.metrics.configuration.properties",
      "accumulo.metrics.configuration.properties", PropertyType.STRING,
      "The Accumulo metrics configuration properties filename. The file is used to"
          + " configure micrometer MeterRegistries that loaded at runtime using classes that"
          + " extend MetricsServiceLoader.class", "2.1.0"),
```

DomGarguilo: thanks!

EdColeman:
A follow task may be to implement a bridge between micrometer and the hadoop metrics system so that, if a hadoop exporter was present, it would read metrics from micrometer and then publish to hadoop configured sinks.

The basic concept would be to find / extend a micrometer registry (see https://micrometer.io/docs/guide/customMeterRegistry) that would instantiate a hadoop metrics system (see System design - https://blog.cloudera.com/what-is-hadoop-metrics2/ ) The custom micrometer registry would function as the "metrics sources" and the hadoop metrics system would configure the sinks by reading the hadoop.metrics.xxx.properties file.  So when the hadoop metrics system "read" metrics from the sources, the values would be obtained via the micrometer metrics registry.  The goal would be that the Accumulo would only have micrometer Meters (gauges / counters,...) and if a hadoop metrics exporter was configured, it would read from the micrometer registry to get those values.  

I'm not sure that the hadoop metrics would be compatible with a StepMeterRegistry - it might require the more general MeterRegistry.  Using a StepMeterRegistry might be enough to figure out some of the concepts - and then find an existing registry (etsy statsd, datadog,...) that matches the hadoop metrics system dimensionality (hierarchical / dimensional), rate aggregation (client / server) and publication (pull / push)  My impression is that hadoop would be hierarchical, client and pull) but that's making some guesses and needs to be confirmed.

DomGarguilo:
> A follow-on task may be to implement a bridge between micrometer and the hadoop metrics system

@EdColeman, do you think this should wait until after these changes are made or incorporated in some way into these changes?

EdColeman:
These changes would have more impact if the hadoop metrics could be removed from the code base - and to do that it, it would ease transitions if there was a way to do the translation for users.  If the translation happens where the metrics are collected / reported, then that would be the only place where the hadoop metrics classes are used / needed.

server/base/src/main/java/org/apache/accumulo/server/metrics/service/MicrometerMetricsFactory.java
```diff
@@ -0,0 +1,182 @@
...
+          DistributionStatisticConfig config) {
+        if (id.getName().equals("replicationQueue")) {
+          return DistributionStatisticConfig.builder().percentiles(0.5, 0.75, 0.9, 0.95, 0.99)
+              .expiry(Duration.ofMinutes(10)).build().merge(config);
+        }
+        return config;
+      }
+    });
+
+    var propKey = METRICS_CONFIG_PROPERTIES.getKey();
```
dlmarion, line 72: Is there a reason that we have to have a configuration file in Accumulo for these pluggable Registries? Can't implementations of the MetricsRegistrationService configure the Registry in its own way?

pom.xml
```diff
@@ -243,6 +244,16 @@
         <artifactId>commons-logging</artifactId>
         <version>1.2</version>
       </dependency>
+      <dependency>
+        <groupId>io.micrometer</groupId>
+        <artifactId>micrometer-core</artifactId>
+        <version>${micrometer.metrics.version}</version>
+      </dependency>
+      <dependency>
```
dlmarion, line 252: I'm thinking that the default metrics implementation would do nothing. If someone wants to use Prometheus (or logging for that matter), then they would drop the appropriate jar file (and configuration information) on the classpath. These implementations could then be moved to another repo (accumulo-metrics?)

server/base/src/main/java/org/apache/accumulo/server/metrics/Metrics.java
```diff
@@ -18,6 +18,9 @@
  */
 package org.apache.accumulo.server.metrics;
 
+import java.util.ServiceLoader;
+
+import org.apache.accumulo.server.metrics.service.MetricsRegistrationService;
 import org.apache.hadoop.metrics2.MetricsCollector;
```
dlmarion, line 24: I'm of the opinion that the Hadoop Metrics2 stuff should be ripped out. If I'm understanding this correctly, we are publishing metrics via Micrometer and Hadoop Metrics concurrently. If both are configured to send metrics to the same destination, then we will be double reporting the numbers. IMO, Hadoop can send it's metrics via mechanism M and Micrometer can send it's metrics via mechanism M if the user so chooses.

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/CompactionExecutorsMetrics.java
```diff
@@ -24,115 +24,231 @@
...
   }
 
-  public CompactionExecutorsMetrics() {
+  public CompactionExecutorsMetrics(MeterRegistry registryMicrometer) {
     super("compactionExecutors");
+    this.micrometerRegistry = registryMicrometer;
+
+    ScheduledExecutorService scheduler =
+        ThreadPools.createScheduledExecutorService(1, "compactionExecutorsMetricsPoller", false);
+    Runtime.getRuntime().addShutdownHook(new Thread(scheduler::shutdownNow));
```
dlmarion, line 73: Just like the `ThreadPools` class is for creating pools, there is a `Threads` class for a common way to create a Thread.

server/base/src/main/java/org/apache/accumulo/server/metrics/service/MicrometerMetricsFactory.java
EdColeman: The intent was to show that we could provide a standard way to allow metrics configuration. This keeps us inline with what is currently provided with the current metrics.  

Also, it seems desirable to allow some configuration flexibility - it may be easier for some users to deploy with a standard set of metrics available in the metrics registration and then configure them through a configuration file.

That was the intent - not sure if this meets those goals or not...

PR #2197, opened 2021-07-09
Per table crypto + other crypto improvements

### LINKED ISSUE #2001: Per table encryption
I want to be able to configure the CryptoService to encrypt on a per table basis.

This will probably require new properties and new information to be passed into the CryptoEnvironment. This also might require changing the CryptoService configuration from `instance` properties to use `general` properties instead.

milleruntime:
* Create table prefix properties for encrypting tables
* Replace instance.crypto.service with instance.crypto.opts.factory for
configuring a crypto factory per instance that returns services
* Change scope in CryptoEnvironment to WAL, Table and Recovery
* Replace CryptoServiceFactory class with a new interface and implement
three types of factories: NoCryptoServiceFactory,
GenericCryptoServiceFactory, and PerTableCryptoServiceFactory
* Utilize general.custom prefix for shared crypto configuration
* Use TableConfiguration to get CryptoService from table properties
* Create CryptoServiceFactory.none() method to return no crypto
* Make CryptoEnvironment return Optional for getDecryptionParams
* Update CryptoTest with new configuration and added tests
* Add new method getAllCryptoProperties() to AccumuloConfiguration
* Fixes for WriteAheadLogEncryptedIT
* Create PerTableCryptoIT
* Refactor bcfile class PrintInfo and rename to PrintBCInfo to fix rfile-info command
* Add test to PerTableCryptoIT and Fix OfflineIterator
* Closes #2001


Manno15:
Depending on which PR is merged first, the properties changed here should match the design set in #2190.

ctubbsii:
> Depending on which PR is merged first, the properties changed here should match the design set in #2190.

I'll finish reviewing #2190 first, since that's much more narrowly scoped (and fewer than 70 files to review!)

milleruntime:
> > Depending on which PR is merged first, the properties changed here should match the design set in #2190.
> 
> I'll finish reviewing #2190 so we can get that in first, since that's much more narrowly scoped (and fewer than 70 files to review!)

Sounds good. I don't think the complexity of this change is that bad though. And don't let the number intimidate you, more than half of those are Tests.

server/tserver/src/test/java/org/apache/accumulo/tserver/log/SortedLogRecoveryTest.java
```diff
@@ -79,6 +80,7 @@
   static final Text cq = new Text("cq");
   static final Value value = new Value("value");
   static ServerContext context;
+  static CryptoService cs = CryptoServiceFactory.none();
```
DomGarguilo, line 83: It does not appear this is used anywhere. Can this be removed or am I overlooking something?

server/tserver/src/main/java/org/apache/accumulo/tserver/InMemoryMap.java
```diff
@@ -754,9 +758,11 @@ public void delete(long waitTime) {
           aconf = createSampleConfig(aconf);
         }
 
+        // TODO can we drop the crateSampleConfig and just use tableConf?
```
DomGarguilo, line 761:
```suggestion
        // TODO can we drop the createSampleConfig and just use tableConf?
```

server/base/src/test/java/org/apache/accumulo/server/client/BulkImporterTest.java
```diff
@@ -112,11 +113,15 @@ public void testFindOverlappingTablets() throws Exception {
     MockTabletLocator locator = new MockTabletLocator();
     FileSystem fs = FileSystem.getLocal(new Configuration());
     ServerContext context = MockServerContext.get();
+    CryptoService cs = CryptoServiceFactory.none();
+    // TableConfiguration tableConf = EasyMock.createMock(TableConfiguration.class);
+    // expect(context.getTableConfiguration(tableId)).andReturn(tableConf).anyTimes();
+    // expect(tableConf.getCryptoService()).andReturn(CryptoServiceFactory.none()).anyTimes();
```
DomGarguilo, line 119: Are these 3 comments intentionally present?

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -19,8 +19,10 @@
 package org.apache.accumulo.server;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static org.apache.accumulo.core.spi.crypto.CryptoEnvironment.Scope.WAL;
 
 import java.io.IOException;
+import java.util.Map;
```
DomGarguilo, line 25:
```suggestion

```
If tableCryptoMap is removed I think this can be as well.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -788,6 +788,20 @@
...
+  @Experimental
+  @Sensitive
+  TABLE_CRYPTO_SENSITIVE_PREFIX("table.crypto.opts.sensitive.", null, PropertyType.PREFIX,
+      "Sensitive properties related to on-disk file encryption."),
+  @Experimental
+  TABLE_CRYPTO_SERVICE("table.crypto.service",
+      "org.apache.accumulo.core.spi.crypto.NoCryptoService", PropertyType.CLASSNAME,
+      "The class which executes on-disk table encryption/decryption. The default does "
+          + "nothing. This property must be a classname with an implementation of the "
+          + "org.apache.accumulo.core.spi.crypto.CryptoService interface."),
```
DomGarguilo, line 804: All added properties will have to have the since added after changes from #2190 are merged in

server/tserver/src/test/java/org/apache/accumulo/tserver/log/SortedLogRecoveryTest.java
milleruntime: Yup good catch. This was leftover from a previous change.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -65,7 +67,8 @@
   private DefaultConfiguration defaultConfig = null;
   private AccumuloConfiguration systemConfig = null;
   private AuthenticationTokenSecretManager secretManager;
-  private CryptoService cryptoService = null;
+  private CryptoService walCryptoService = null;
+  private Map<TableId,CryptoService> tableCryptoMap;
```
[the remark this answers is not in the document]
milleruntime: This was an unfinished idea I had to cache the crypto Services in ServerContext so they don't need to be created every time. Unless someone thinks this needs to be done for this change to work properly, I  can remove this and create a follow on task.

server/tserver/src/main/java/org/apache/accumulo/tserver/InMemoryMap.java
milleruntime: I added this comment to look at this again and I don't think we can. Removed comment in 515e963.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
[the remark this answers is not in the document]
milleruntime: Also, we had previously done work in ServerContext to make sure only one Crypto Service was created per tserver. If we want to continue this behavior on the per table level, we will need to do something like a cache to make sure each table only creates one Crypto Service. I am not sure how this could be done though as tablets could be spread across different tablet servers. We may have to do something in the Manager where the tservers ask for the CryptoService per table.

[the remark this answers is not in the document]
keith-turner: > This was an unfinished idea I had to cache the crypto Services in ServerContext so they don't need to be created every time. 

@milleruntime I am looking at this comment in isolation, so my response may not be applicable.  Its possible you could use a configuration Deriver to create this instead of the Map.  Some examples

https://github.com/apache/accumulo/blob/548b037c5917dc1e693733dc3d89599369d79103/server/base/src/main/java/org/apache/accumulo/server/conf/TableConfiguration.java#L269-L275

https://github.com/apache/accumulo/blob/548b037c5917dc1e693733dc3d89599369d79103/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Tablet.java#L341-L349

https://github.com/apache/accumulo/blob/548b037c5917dc1e693733dc3d89599369d79103/core/src/main/java/org/apache/accumulo/core/conf/AccumuloConfiguration.java#L525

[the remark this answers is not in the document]
milleruntime: I already did on this PR! Check out the change to make sure I did it correctly: 
https://github.com/apache/accumulo/pull/2197/files#diff-c8804c4993551d0fc6f293963d664d4e35151f7639955f4c7e2b3a7bf2a694b1

milleruntime:
I tested this branch on Uno locally using continuous ingest and was able to ingest 160 million entries in 20 minutes. The data verified correctly in about 15 minutes.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -176,22 +176,6 @@
...
-  INSTANCE_CRYPTO_SENSITIVE_PREFIX("instance.crypto.opts.sensitive.", null, PropertyType.PREFIX,
-      "Sensitive properties related to on-disk file encryption.", "2.0.0"),
-  @Experimental
-  INSTANCE_CRYPTO_SERVICE("instance.crypto.service",
-      "org.apache.accumulo.core.spi.crypto.NoCryptoService", PropertyType.CLASSNAME,
-      "The class which executes on-disk file encryption. The default does nothing. To enable "
-          + "encryption, replace this classname with an implementation of the"
-          + "org.apache.accumulo.core.spi.crypto.CryptoService interface.",
-      "2.0.0"),
-
```
ctubbsii, line 194: Part of the benefit of having crypto properties as `instance.` properties ensures that all tservers are using the same config. The potential for different tservers to have different crypto services means that data written by one tserver might not be able to be read by another, as they could have wildly different behaviors. This is one reason why I think we should keep an `instance.` property to refer to the crypto service factory, which is the overall module that supports all the different crypto that will be needed, and that per-tserver and per-table properties should only modify the behavior of the crypto services produced by this factory with some options.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -474,6 +458,21 @@
...
+  @Sensitive
+  TSERV_WAL_CRYPTO_SENSITIVE_PREFIX("tserver.wal.crypto.opts.sensitive.", null, PropertyType.PREFIX,
+      "Sensitive properties related to on-disk file encryption.", "2.1.0"),
+  @Experimental
+  TSERV_WAL_CRYPTO_SERVICE("tserver.wal.crypto.service",
+      "org.apache.accumulo.core.spi.crypto.NoCryptoService", PropertyType.CLASSNAME,
+      "The class which executes on-disk write ahead log encryption/decryption. The default does "
+          + "nothing. This property must be a classname with an implementation of the "
+          + "org.apache.accumulo.core.spi.crypto.CryptoService interface.",
+      "2.1.0"),
```
ctubbsii, line 475: I would expect all tservers to use the same WAL options, so that files written by one tserver will be able to be read by another. This implies the use of `instance.` properties for WAL crypto options, not `tserver.` Also, if a crypto service factory is specified in an `instance.` property, then there's no need to specify a specific crypto service here, as the factory would make that decision based on the options and scope.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -894,6 +893,21 @@
...
+  @Sensitive
+  TABLE_CRYPTO_SENSITIVE_PREFIX("table.crypto.opts.sensitive.", null, PropertyType.PREFIX,
+      "Sensitive properties related to on-disk file encryption.", "2.1.0"),
+  @Experimental
+  TABLE_CRYPTO_SERVICE("table.crypto.service",
+      "org.apache.accumulo.core.spi.crypto.NoCryptoService", PropertyType.CLASSNAME,
+      "The class which executes on-disk table encryption/decryption. The default does "
+          + "nothing. This property must be a classname with an implementation of the "
+          + "org.apache.accumulo.core.spi.crypto.CryptoService interface.",
+      "2.1.0"),
```
ctubbsii, line 910: If there were a crypto service factory specified as an `instance.` option, then there's no need to specify a specific crypto service here, as that would be determined by the factory, given the TABLE scope and the options (either the options in the config, in the case of writing a new file, or the options read from the file in the case of reading a previously written file).

core/src/main/java/org/apache/accumulo/core/conf/Property.java
milleruntime: That makes sense. I had only gone with "tserver" for "wal" properties since that is where most of the write ahead log configuration lives.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
milleruntime: I will try out both types of configuration using `instance` properties.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
ctubbsii: Yeah, I can see it both ways: "wal" is a tserver feature, but "crypto" support is an instance feature. I see this as somewhat more closely aligned with the way volume chooser does things, where it's specified at the system scope (using `general.` prefix) but a specific implementation could delegate to per-table config to act in a per-table way.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
milleruntime: Just so we are on the same page, you are saying we should just have one `instance.crypto.service` for the whole instance? Then if you want to specify what tables to use it, say just have a list the tables to encrypt?
`instance.crypto.tables=table1,table2,table3`
`instance.crypto.wal.enabled=true`

I think that would be easier then having something like:
`instance.crypto.service.table1=true`
`instance.crypto.service.table2=true`

ctubbsii: No. I'm saying we should have one `instance.crypto.service.factory`, and that should spit out crypto services based on the per-table config. This is essentially how the VolumeManager per-table feature works.

milleruntime: We would use the `table.custom` the same way? The volume chooser uses the same class type for `general` properties and for `tables.custom` properties. And what about the wals? Would they have their own factory? Maybe if you give an example of your idea of a config that would help.

ctubbsii: I'm not a fan of the `.custom` property prefix naming convention. It implies that we have a bag of properties segregated between "official" properties in Accumulo property namespaces, and "unofficial" properties in a designated namespace for user-provided plugins. But, our reference implementations use the `.custom` property namespace in some cases but not in others. We're not consistent. I think we should just treat it as one big bag of well-named and well-scoped properties and relax the constraint that requires users to use the `.custom` namespace for their properties.

So, that said, I'd be fine leaving the `table.crypto.`, rather than require they be in `table.custom.crytpo.` or something similar. I'm not suggesting copying the property naming convention from VolumeChooser, just the pattern of letting the single factory instance decide whether and how to use these properties, based on its implementation and the provided scope.

So, imagine a `MyCryptoServiceFactoryThatSupportsPerTableConfig` being set as `instance.crypto.service.factory`. The factory has an API that is called, something like `factory.getCryptoService(scopeConfig)`. So, depending on scope, you'd provide different configs:

```java
var walCryptoService = factory.getCryptoService(systemConf.getAllPropertiesWithPrefixStripped("instance.crypto.wal."));
var tableCryptoServiceForTableA = factory.getCryptoService(tableAConf.getAllPropertiesWithPrefixStripped("table.crypto."));
```

Now, it's up to the implementation of the factory to decide what CryptoService to return, based on the scope and scope config. (My example only shows passing in config properties, but you could pass in the scope as a type as well, or some other object performing a similar function of communicating context to the factory.)

milleruntime: OK. Yeah I like `table.crypto.` better as well. 

If we are not specifying the crypto service in the table config, do we use the same crypto service for all tables? If not, what in the table config will determine which crypto service to use. Based on what you are saying, I am imagining a config like this:
<pre>
instance.crypto.factory=MyCryptoServiceFactoryThatSupportsPerTableConfig
instance.wal.enabled=true
instance.crypto.wal.opts.key.uri=/home/mike/workspace/uno/CryptoTest-testkeyfile
secureTable:
table.crypto.enabled=true
table.crypto.type=strong
table.crypto.opts.key=/something/more/secure/than/a/file
lessSecureTable:
table.crypto.enabled=true
table.crypto.type=weak
table.crypto.opts.key.uri=/home/mike/workspace/uno/CryptoTest-testkeyfile2
</pre>

With the impl of MyCryptoServiceFactoryThatSupportsPerTableConfig looking like this:
<pre>
if scope == wal && enabled, return AESCryptoService
if scope == table && enabled && type=strong, return StrongerCryptoService
if scope == table && enabled && type=weak, return WeakerCryptoService
</pre>

ctubbsii: > If we are not specifying the crypto service in the table config, do we use the same crypto service for all tables? If not, what in the table config will determine which crypto service to use.

I think that would be implementation-dependent.

> Based on what you are saying, I am imagining a config like this:
> 
> instance.crypto.factory=MyCryptoServiceFactoryThatSupportsPerTableConfig
> instance.wal.enabled=true
> instance.crypto.wal.opts.key.uri=/home/mike/workspace/uno/CryptoTest-testkeyfile
> secureTable:
> table.crypto.enabled=true
> table.crypto.type=strong
> table.crypto.opts.key=/something/more/secure/than/a/file
> lessSecureTable:
> table.crypto.enabled=true
> table.crypto.type=weak
> table.crypto.opts.key.uri=/home/mike/workspace/uno/CryptoTest-testkeyfile2
> With the impl of MyCryptoServiceFactoryThatSupportsPerTableConfig looking like this:
> 
> if scope == wal && enabled, return AESCryptoService
> if scope == table && enabled && type=strong, return StrongerCryptoService
> if scope == table && enabled && type=weak, return WeakerCryptoService

Something like that. I would probably just put everything under `table.crypto.opts.`, though. So, something closer to:

```ini
instance.crypto.service.factory=PerTableAESCryptoFactory
instance.crypto.wal.opts.algorithm=AES256+CBC
instance.crypto.wal.opts.key.uri=/path/to/keyfile
; no crypto table (still provide the key, in case it needs to read old data):
table.crypto.opts.algorithm=NONE
table.crypto.opts.key.uri=/path/to/keyfile
; table A:
table.crypto.opts.algorithm=AES128+ECB
table.crypto.opts.key.uri=/path/to/keyfile
; table B:
table.crypto.opts.algorithm=AES512+ECB
table.crypto.opts.key.uri=/path/to/keyfile
```

inside PerTableAESCryptoFactory
```java
  if (algorithm == NONE) return new CryptoService(new NoEncrypter(), new AESDecrypter(key));
  else return new CryptoService(new AESEncrypter(algorithm, key), new AESDecrypter(key));
```

However, exactly how/where the algorithm/key/other opts get passed depends on what the interfaces look like exactly, and at what point are they called.

ctubbsii: > So it looks like these are the only crypto properties we need to bake into Accumulo's Property.java:

Looks mostly right to me. But, I'm not sure about the "sensitive" stuff. I think maybe we could just assume all crypto parameters are "sensitive". This really only has to do with returning the config via the public API for printing in the shell. We could require admin permission to retrieve/view any of those without creating two maps for all the opts.

> instance.crypto.factory (don't think we need service in the name)

Agreed on the parenthetical.

> Do you think we need to keep the two we have now `instance.crypto.opts.` and `instance.crypto.opts.sensitive.`? The only thing I can think of is if a user needs generic properties for the factory itself.

I'm not sure. If you use them, then the interface would probably have to require an `init(Map<String,String> opts, Map<String,String> sensitiveOpts)` method, in addition to a no-arg constructor to instantiate the class. (or just a single map parameter, if we drop the separate set of sensitive props).

milleruntime:
@ctubbsii mind taking a quick high level look at this branch before I redo the tests? I think I implemented the changes we discussed and got everything compiling (ignore tests for now). Changes include:

- I dropped the `init` method from CryptoService
- Added TableId to CryptoEnvironment
- Created NoCryptoFactory for no crypto
- Used NoCryptoFactory in RFile API, until we figure out how to support crypto in the RFile API
- Implemented CryptoFactoryLoader for loading the factory and services
- Created a basic factory called PerTableCryptoFactory

Classes to take a quick look at:
[Property](https://github.com/apache/accumulo/pull/2197/files#diff-6c2d215eebb1a6157a2b7d95f49496834bd91a150c8aedd97cca6b01aedc05cb)
[CryptoFactoryLoader](https://github.com/apache/accumulo/pull/2197/files#diff-6bc04f7fde1143ced2a2ef3097baf6429467690b93557e44c0c9c2542b6d32e9)
[PerTableCryptoFactory](https://github.com/apache/accumulo/pull/2197/files#diff-8a243847ef39a9dedc8daceaa1629aca85a4a550b2ff20ec3ec249c0bb917625)
[CryptoFactory interface](https://github.com/apache/accumulo/pull/2197/files#diff-3ab2b08082155f62ed88ac1bb83d9456ece437b562425e255bc76170f3004e7f)
[CryptoEnvironment interface](https://github.com/apache/accumulo/pull/2197/files#diff-e9b7d2e72ebd5675e60efb4646ae6cbe868199c557225ec18db0392d2368b300)
[TableConfiguration](https://github.com/apache/accumulo/pull/2197/files#diff-c8804c4993551d0fc6f293963d664d4e35151f7639955f4c7e2b3a7bf2a694b1)
[ServerContext](https://github.com/apache/accumulo/pull/2197/files#diff-44b9172cc16cdb9b671926c7aed1b0d57372715bc181dd79ef0be76828b708f3)

ctubbsii:
@milleruntime I'd like to take another look at this soon, but it seems to be a little out of date right now. Want to rebase it?

core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java
```diff
@@ -132,11 +134,11 @@ public FileEncrypter getFileEncrypter(CryptoEnvironment environment) {
...
-    byte[] decryptionParams = environment.getDecryptionParams();
-    if (decryptionParams == null || checkNoCrypto(decryptionParams))
+    var decryptionParams = environment.getDecryptionParams();
+    if (decryptionParams.isEmpty() || checkNoCrypto(decryptionParams.get()))
       return new NoFileDecrypter();
 
-    ParsedCryptoParameters parsed = parseCryptoParameters(decryptionParams);
+    ParsedCryptoParameters parsed = parseCryptoParameters(decryptionParams.get());
     Key kek = loadDecryptionKek(parsed);
     Key fek = unwrapKey(parsed.getEncFek(), kek);
```
dlmarion, line 143: What does `Fek` mean in this context?

core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java
```diff
@@ -132,11 +134,11 @@ public FileEncrypter getFileEncrypter(CryptoEnvironment environment) {
...
     CryptoModule cm;
-    byte[] decryptionParams = environment.getDecryptionParams();
-    if (decryptionParams == null || checkNoCrypto(decryptionParams))
+    var decryptionParams = environment.getDecryptionParams();
+    if (decryptionParams.isEmpty() || checkNoCrypto(decryptionParams.get()))
       return new NoFileDecrypter();
 
-    ParsedCryptoParameters parsed = parseCryptoParameters(decryptionParams);
+    ParsedCryptoParameters parsed = parseCryptoParameters(decryptionParams.get());
     Key kek = loadDecryptionKek(parsed);
```
dlmarion, line 142: From what I can tell, loadDecryptionKek is only called once, and inside of the method it only parses and stores one key for decryption in the variable `decryptingKeys`. However, `decryptingKeys` is a Map, which makes sense because IIRC you will only use 1 current un-expired key for encrypting, but you may need the current and all historical keys (the expired or rotated ones) for decrypting.

core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java
milleruntime: File encrypting key? I am not sure its been a while.

dlmarion: Yeah, some of the abbreviations are foreign to me. I think you are right that `fek` is [File Encryption Key](https://csrc.nist.gov/glossary/term/fek). However, it's not used for the entire file, right? It's used for the block? I have seen it referenced as a [Secret Key](https://csrc.nist.gov/glossary/term/secret_key_cryptographic_algorithm) in the [JCA](https://docs.oracle.com/javase/8/docs/technotes/guides/security/crypto/CryptoSpec.html#SecretKeyFactory).

milleruntime: It is used for both, the WALs or block encryption for R-Files, depending on which Scope is provided in the `CryptoEnvironment`.
https://github.com/apache/accumulo/blob/41310c57dbf840ab0ed5faea70f6c3b8311a9697/core/src/main/java/org/apache/accumulo/core/spi/crypto/CryptoService.java#L44-L48

milleruntime: To clarify... We use "AES/CBC/NoPadding" transformation for WALs and "AES/GCM/NoPadding" transformation for RFiles. So different modes per type of file.

dlmarion: I take that back, it looks like it might be used for the entire file and a new IV is generated for each block. I do have a concern though that we are leaving unencrypted secret keys (fek) in the heap after using them. Reference [here](https://github.com/apache/accumulo/blob/d454afd39c91ee0a1ab267e0caafc52254d209b0/core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java#L316), [here](https://github.com/apache/accumulo/blob/d454afd39c91ee0a1ab267e0caafc52254d209b0/core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java#L449) and [here](https://github.com/apache/accumulo/blob/d454afd39c91ee0a1ab267e0caafc52254d209b0/core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java#L141). I believe, looking at the class javadoc for [SecretKey](https://github.com/openjdk/jdk/blob/jdk-11%2B27/src/java.base/share/classes/javax/crypto/SecretKey.java), we should be calling `destroy()` when we are done with it. I'm not sure we are doing that.

milleruntime: Are you talking about these changes or the current code?

ctubbsii: FEK = file encrypting key, KEK = key encrypting key. A Google search resulted in: https://en.wikipedia.org/wiki/Glossary_of_cryptographic_keys
I've heard "FEK" more commonly than "DEK", but "DEK" seems to be a more generic term for basically the same thing... as far as I can tell from that glossary anyway. "FEK" is used in https://en.wikipedia.org/wiki/Encrypting_File_System and https://www.kernel.org/doc/html/latest/security/keys/ecryptfs.html, so I think it's more common in terms of file system or file-based encryption.

In any case, the idea is that one key that's relatively persistent somewhere (KEK) encrypts a random key (FEK) that's used to actually encrypt the file's blocks. I think Dave is correct the FEK is used for the whole file, with each block using a new IV. The KEK should be per-table.

I agree that we should call destroy on them after we're done with them. I'm not sure it's worth zeroing out the byte arrays manually, though, because that's making a lot of assumptions about how Java handles memory internally, and I think it's not likely to work the way we expect. Calling destroy on them should be easy, and we would be deferring to the implementation to do it correctly.

milleruntime: I can try to integrate the call to destroy into this PR, pulling some of Dave's changes from #2616 . Should I wait until you take a look at my latest changes?

core/src/main/java/org/apache/accumulo/core/spi/crypto/CryptoEnvironment.java
```diff
@@ -28,10 +32,18 @@ public interface CryptoEnvironment {
    * Where in Accumulo the on-disk file encryption takes place.
    */
   enum Scope {
-    WAL, RFILE
+    WAL, TABLE
```
milleruntime, line 35: I was looking at these changes again and wondering if we should add another scope: `RECOVERY`. With my changes to the recovery code, we now have the ability to encrypt the R-files during recovery...

ctubbsii: Another type probably makes sense. These files use the same file type as per-table files, but contain data for multiple tables, like WAL files.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -135,13 +140,16 @@ public InstanceId getInstanceID() {
...
+      throw new CryptoService.CryptoException("Crypto Factory " + cryptoFactory.getClass().getName()
           + " already exists and cannot be setup again");
     }
 
     AccumuloConfiguration acuConf = getConfiguration();
-    cryptoService = CryptoServiceFactory.newInstance(acuConf, ClassloaderType.ACCUMULO);
+    cryptoFactory = CryptoFactoryLoader.newInstance(acuConf, ACCUMULO);
+    var props = acuConf.getAllPropertiesWithPrefixStripped(Property.INSTANCE_CRYPTO_PREFIX);
+    CryptoEnvironment env = new CryptoEnvironmentImpl(CryptoEnvironment.Scope.WAL, null, null);
+    walCryptoService = cryptoFactory.getService(env, props);
```
milleruntime, line 152: @ctubbsii recently in #2658 you changed ServerContext to use `Supplier` to provide a lot of the resources. What do you think about the 2 CryptoServices? I thought I had created `setupCrypto()` to take care of this but I guess that is not in main yet. I am currently working on merging conflicts for this branch.

milleruntime: I think i figured it out. I saw that you dropped the method in #2658.

milleruntime:
@ctubbsii I re-base'd my per-table crypto changes again with latest for 2.1. Did you want to take another look at this PR before I rewrite the tests? There are a lot of changes but I think the design is solid. Here is a summary of the big updates for this feature:

1. New `CryptoFactoryLoader` class. This class replaces `CryptoServiceFactory`. It can create a new instance of CryptoFactory and has convenience methods to get the CryptoService directly.
- `public static CryptoFactory newInstance(AccumuloConfiguration conf, ClassloaderType ct)`
- `public static CryptoFactory none() `
- `public static CryptoService getCryptoJavaTable(AccumuloConfiguration conf)` (Used by server code)
- `public static CryptoService getCryptoFromTableProps(InstanceOperations instOps, Map<String,String> tableProps, TableId TableId)` (Used by client code)

2. New `CryptoFactory` interface. Single method: `  CryptoService getService(CryptoEnvironment environment, Map<String,String> properties)`
- Create 2 basic factory implementations: `PerTableCryptoFactory` and `NoCryptoFactory`

3. Modifications to the current Crypto.
- Drop the init method from `CryptoService`
- Add TableId to `CryptoEnvironment`

4. Disable Crypto in the `RFile` API. This is a stop gap until we figure out how to support crypto in the RFile API


Classes with major changes:
[Property](https://github.com/apache/accumulo/pull/2197/files#diff-6c2d215eebb1a6157a2b7d95f49496834bd91a150c8aedd97cca6b01aedc05cb)
[CryptoFactoryLoader](https://github.com/apache/accumulo/pull/2197/files#diff-6bc04f7fde1143ced2a2ef3097baf6429467690b93557e44c0c9c2542b6d32e9)
[PerTableCryptoFactory](https://github.com/apache/accumulo/pull/2197/files#diff-8a243847ef39a9dedc8daceaa1629aca85a4a550b2ff20ec3ec249c0bb917625)
[CryptoFactory interface](https://github.com/apache/accumulo/pull/2197/files#diff-3ab2b08082155f62ed88ac1bb83d9456ece437b562425e255bc76170f3004e7f)
[CryptoEnvironment interface](https://github.com/apache/accumulo/pull/2197/files#diff-e9b7d2e72ebd5675e60efb4646ae6cbe868199c557225ec18db0392d2368b300)
[TableConfiguration](https://github.com/apache/accumulo/pull/2197/files#diff-c8804c4993551d0fc6f293963d664d4e35151f7639955f4c7e2b3a7bf2a694b1)
[ServerContext](https://github.com/apache/accumulo/pull/2197/files#diff-44b9172cc16cdb9b671926c7aed1b0d57372715bc181dd79ef0be76828b708f3)

core/src/test/java/org/apache/accumulo/core/crypto/CryptoTest.java
```diff
@@ -289,10 +319,20 @@ public void testRFileEncrypted() throws Exception {
...
 
-    Scanner iter =
-        RFile.newScanner().from(file).withFileSystem(fs).withTableProperties(cryptoOnConf).build();
+    // test to make sure the RFile is encrypted
     ArrayList<Key> keysRead = new ArrayList<>();
-    iter.forEach(e -> keysRead.add(e.getKey()));
+    try (Scanner iter = RFile.newScanner().from(file).withFileSystem(fs).build()) {
+      iter.forEach(e -> keysRead.add(e.getKey()));
+      fail("File was not encrypted.");
+    } catch (UncheckedIOException e) {
```
DomGarguilo, line 327:
```suggestion
      assertThrows(UncheckedIOException.class, () -> iter.forEach(e -> keysRead.add(e.getKey())),
          "The file was expected to be encrypted but was not");
```
This might be a better place for `assertThrows` than try-catch

milleruntime: Yep, thanks!

test/src/main/java/org/apache/accumulo/test/functional/PerTableCryptoIT.java
```diff
@@ -0,0 +1,213 @@
...
+      TableId tableId0 = TableId.of(c.tableOperations().tableIdMap().get(tables[0]));
+      VerifyIngest.VerifyParams params = new VerifyIngest.VerifyParams(getClientProps(), tables[0]);
+      TestIngest.ingest(c, params);
+      VerifyIngest.verifyIngest(c, params);
+
+      // verify data was not encrypted
+      System.out.println("-------------" + checkTableEncryption(tableId0, false));
+
+      // encrypt table 1
+      props.clear();
```
DomGarguilo, line 114: It looks like `props` is not used before it is cleared. Is this intended?

test/src/main/java/org/apache/accumulo/test/functional/PerTableCryptoIT.java
```diff
@@ -0,0 +1,213 @@
...
+      TestIngest.ingest(c, params);
+      VerifyIngest.verifyIngest(c, params);
+      TableId tableId = TableId.of(c.tableOperations().tableIdMap().get(tableName));
+      c.tableOperations().offline(tableName, true);
+
+      var oScanner = new OfflineScanner((ClientContext) c, tableId, Authorizations.EMPTY);
+      int count = 0;
+      for (var e : oScanner) {
+        count++;
+      }
```
DomGarguilo, line 164:
```suggestion
      long count = oScanner.stream().count();
```
Thought this was a nice use of Scanners stream

test/src/main/java/org/apache/accumulo/test/functional/PerTableCryptoIT.java
milleruntime: That was holdover from some experimentation I was doing to try and detect encryption in the WALs. I think I can drop it now.

milleruntime:
This PR is ready for review. I still have a few tasks that can be follow on work but I want to keep this PR to the minimum changes required. I have done some manual testing using Uno and all the ITs are passing.

Here are the follow on tasks I have in mind:

1. Modify or drop the AESCryptoService enabled property
2. Create an option for wal-info and rfile-info to just print headers
3. Add test for recovery scope
4. Test utils: CreateEmpty & GenerateSplits

dlmarion (commented):
I think I found an issue...

core/src/main/java/org/apache/accumulo/core/spi/crypto/AESCryptoService.java
```diff
@@ -381,14 +396,21 @@ void incrementIV(byte[] iv, int i) {
...
+
+      private final Cipher cipher;
       private final Key fek;
 
       AESGCMFileDecrypter(Key fek) {
+        try {
+          cipher = Cipher.getInstance(transformation);
+        } catch (NoSuchAlgorithmException | NoSuchPaddingException e) {
+          throw new CryptoException("Error obtaining cipher for transform " + transformation, e);
+        }
```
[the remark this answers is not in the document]
dlmarion: [This](https://bugs.openjdk.org/browse/JDK-8012900) leads me to believe that the issue is resolved and that the javadoc is wrong.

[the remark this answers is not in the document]
dlmarion: [This](https://blog.heckel.io/2014/03/01/cipherinputstream-for-aead-modes-is-broken-in-jdk7-gcm/) blog post explains the issue and has related tests [here](https://github.com/binwiederhier/cipherinputstream-aes-gcm/blob/master/src/CipherInputStreamIssuesTests.java). According to the last comment on the blog post, running the `testB_JavaxCipherInputStreamWithAesGcm` test, which fails in the blog post, passes in JDK11. I did this locally and the test passes.

core/src/main/java/org/apache/accumulo/core/spi/crypto/CryptoService.java
```diff
@@ -21,7 +21,7 @@
 import java.util.Map;
 
 /**
- * Self contained cryptographic service. All on disk encryption and decryption will take place
+ * A self-contained cryptographic service. All on disk encryption and decryption will take place
```
dlmarion, line 24: This comment is outside the scope of the changes here, but I wonder if `CryptoException` should extend `GeneralSecurityException` like the other Exceptions in the crypto code. I'm also wondering if `CryptoException` should also be declared on all the methods and the implementations changed to *not* throw RuntimeExceptions.

milleruntime: Yeah, this would be a good follow on task.

server/base/src/main/java/org/apache/accumulo/server/util/FileUtil.java
```diff
@@ -522,15 +506,15 @@ public static Map<TabletFile,FileInfo> tryToGetFirstAndLastRows(ServerContext co
...
 
     Key lastKey = null;
 
     for (TabletFile file : mapFiles) {
       FileSystem ns = context.getVolumeManager().getFileSystemByPath(file.getPath());
       FileSKVIterator reader = FileOperations.getInstance().newReaderBuilder()
-          .forFile(file.getPathStr(), ns, ns.getConf(), context.getCryptoService())
-          .withTableConfiguration(context.getConfiguration()).seekToBeginning().build();
+          .forFile(file.getPathStr(), ns, ns.getConf(), tableConf.getCryptoService())
+          .withTableConfiguration(tableConf).seekToBeginning().build();
```
dlmarion, line 517: I wonder if `FileOperations` should change, something like:

```
FileOperations.getInstance().newReaderBuilder().table(TableConfiguration)
     .file(String, FileSystem, Configuration);
```

core/src/main/java/org/apache/accumulo/core/spi/crypto/PerTableCryptoServiceFactory.java
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Loads a crypto service based on the name provided.
+   */
+  private CryptoService newCryptoService(String cryptoServiceName) {
+    CryptoService cs;
+    try {
+      cs = PerTableCryptoServiceFactory.class.getClassLoader().loadClass(cryptoServiceName)
+          .asSubclass(CryptoService.class).getDeclaredConstructor().newInstance();
+    } catch (ClassNotFoundException | IllegalAccessException | InvocationTargetException
```
dlmarion, line 108: can replace all of these with `ReflectiveOperationException`

core/src/main/java/org/apache/accumulo/core/spi/crypto/PerTableCryptoServiceFactory.java
```diff
@@ -0,0 +1,118 @@
...
+    return cs;
+  }
+
+  /**
+   * Loads a crypto service based on the name provided.
+   */
+  private CryptoService newCryptoService(String cryptoServiceName) {
+    CryptoService cs;
+    try {
+      cs = PerTableCryptoServiceFactory.class.getClassLoader().loadClass(cryptoServiceName)
```
dlmarion, line 106:
```suggestion
      return ClassLoaderUtil.loadClass(null, cryptoServiceName, CryptoService.class).getDeclaredConstructor().newInstance();
```

core/src/main/java/org/apache/accumulo/core/spi/crypto/GenericCryptoServiceFactory.java
```diff
@@ -0,0 +1,70 @@
...
+    return cs;
+  }
+
+  /**
+   * Loads a crypto service based on the name provided
+   */
+  private CryptoService newCryptoService(String cryptoServiceName) {
+    CryptoService cs;
+    try {
+      cs = GenericCryptoServiceFactory.class.getClassLoader().loadClass(cryptoServiceName)
```
dlmarion, line 61: See PerTableCryptoServiceFactory for comments, they apply here too. Maybe a default method is better on the interface?

test/src/main/java/org/apache/accumulo/test/performance/scan/CollectTabletStats.java
```diff
@@ -497,7 +497,7 @@ private static int readFilesUsingIterStack(VolumeManager fs, ServerContext conte
     for (TabletFile file : files) {
       FileSystem ns = fs.getFileSystemByPath(file.getPath());
       readers.add(FileOperations.getInstance().newReaderBuilder()
-          .forFile(file.getPathStr(), ns, ns.getConf(), CryptoServiceFactory.newDefaultInstance())
+          .forFile(file.getPathStr(), ns, ns.getConf(), NoCryptoServiceFactory.NONE)
```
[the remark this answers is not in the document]
milleruntime: This shouldn't matter for the test.

core/src/main/java/org/apache/accumulo/core/spi/crypto/NoCryptoServiceFactory.java
```diff
@@ -0,0 +1,30 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.accumulo.core.spi.crypto;
+
+import java.util.Map;
+
+public class NoCryptoServiceFactory implements CryptoServiceFactory {
+  public static final CryptoService NONE = new NoCryptoService();
```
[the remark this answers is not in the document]
milleruntime: Could make the class private and name the method `none()` or `get()`.

server/base/src/main/java/org/apache/accumulo/server/util/FileUtil.java
milleruntime: FileOperations is a mess. There is definitely room for improvement.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -384,11 +384,11 @@ MetadataTime computeRootTabletTime(ServerContext context, Collection<String> goo
         Path path = new Path(good);
 
         FileSystem ns = context.getVolumeManager().getFileSystemByPath(path);
+        var tableConf = context.getTableConfiguration(RootTable.ID);
         long maxTime = -1;
         try (FileSKVIterator reader = FileOperations.getInstance().newReaderBuilder()
-            .forFile(path.toString(), ns, ns.getConf(), context.getCryptoService())
-            .withTableConfiguration(context.getTableConfiguration(RootTable.ID)).seekToBeginning()
-            .build()) {
+            .forFile(path.toString(), ns, ns.getConf(), tableConf.getCryptoService())
```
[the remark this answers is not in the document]
milleruntime: Good question. I am not sure. I think it would be simpler to assume no encryption. But if the user wants the files to be encrypted during the upgrade, I think they would need the service here.

core/src/main/java/org/apache/accumulo/core/spi/crypto/GenericCryptoServiceFactory.java
milleruntime: Which interface?

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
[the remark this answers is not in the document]
dlmarion: And, I think I forgot when I commented that encryption existed in the prior version. So, the file could be encrypted already using the prior implementation?

core/src/main/java/org/apache/accumulo/core/spi/crypto/GenericCryptoServiceFactory.java
dlmarion: `GenericCryptoServiceFactory implements CryptoServiceFactory`

CryptoServiceFactory

test/src/main/java/org/apache/accumulo/test/performance/scan/CollectTabletStats.java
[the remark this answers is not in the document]
dlmarion: ah, good catch. I didn't realize it was in a test package. No Test in the name.

core/src/main/java/org/apache/accumulo/core/spi/crypto/GenericCryptoServiceFactory.java
milleruntime: Ah OK. Wasn't sure which one. Yeah that might be nice.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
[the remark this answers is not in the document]
milleruntime: Good point. We could add a method to the Upgrader to check for encryption. But that could be a follow on and maybe just a nice to have.

core/src/main/java/org/apache/accumulo/core/spi/crypto/GenericCryptoServiceFactory.java
milleruntime: I pulled the method up to the interface. I was thinking it would be a nice method for users writing their own factory so they don't have to figure out the Accumulo class loading and having it on the interface makes it highly visible.

milleruntime: This does expose `ClassLoaderUtil` to the SPI. I don't know if that is a problem or not.

dlmarion: It's used internally in a method though, not as an input parameter or result.

milleruntime:
I am going to merge this PR so we can do follow up tasks and more testing.

PR #2228, opened 2021-08-10
Fixes for #2203

harjitdotsingh:
Fixes for #2203

### DIFF
--- a/server/base/pom.xml
+++ b/server/base/pom.xml
@@ -116,11 +116,6 @@
       <artifactId>easymock</artifactId>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.jline</groupId>
-      <artifactId>jline</artifactId>
-      <scope>test</scope>
-    </dependency>
   </dependencies>
   <build>
     <testResources>
--- a/server/base/pom.xml
+++ b/server/base/pom.xml
@@ -93,10 +93,6 @@
       <artifactId>zookeeper-jute</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.jline</groupId>
-      <artifactId>jline</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
@@ -118,6 +114,11 @@
     <dependency>
       <groupId>org.easymock</groupId>
       <artifactId>easymock</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jline</groupId>
+      <artifactId>jline</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
@@ -23,7 +23,6 @@ import static org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSec
 import static org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSection.ServerColumnFamily.TIME_COLUMN;
 import static org.apache.accumulo.core.metadata.schema.MetadataSchema.TabletsSection.TabletColumnFamily.PREV_ROW_COLUMN;
 
-import java.io.Console;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.util.Arrays;
@@ -662,8 +661,7 @@ public class Initialize implements KeywordExecutable {
     return Constants.ZROOT + Constants.ZINSTANCES + "/";
   }
 
-  private String getInstanceNamePath(Opts opts)
-      throws IOException, KeeperException, InterruptedException {
+  private String getInstanceNamePath(Opts opts) throws KeeperException, InterruptedException {
     // setup the instance name
     String instanceName, instanceNamePath = null;
     boolean exists = true;
@@ -702,14 +700,13 @@ public class Initialize implements KeywordExecutable {
     return instanceNamePath;
   }
 
-  private String getRootUserName(SiteConfiguration siteConfig, Opts opts) throws IOException {
+  private String getRootUserName(SiteConfiguration siteConfig, Opts opts) {
     final String keytab = siteConfig.get(Property.GENERAL_KERBEROS_KEYTAB);
     if (keytab.equals(Property.GENERAL_KERBEROS_KEYTAB.getDefaultValue())
         || !siteConfig.getBoolean(Property.INSTANCE_RPC_SASL_ENABLED)) {
       return DEFAULT_ROOT_USER;
     }
 
-    Console c = System.console();
     System.out.println("Running against secured HDFS");
 
     if (opts.rootUser != null) {
@@ -717,7 +714,8 @@ public class Initialize implements KeywordExecutable {
     }
 
     do {
-      String user = c.readLine("Principal (user) to grant administrative privileges to : ");
+      String user =
+          System.console().readLine("Principal (user) to grant administrative privileges to : ");
       if (user == null) {
         // should not happen
         System.exit(1);
@@ -728,22 +726,20 @@ public class Initialize implements KeywordExecutable {
     } while (true);
   }
 
-  private byte[] getRootPassword(SiteConfiguration siteConfig, Opts opts, String rootUser)
-      throws IOException {
+  private byte[] getRootPassword(SiteConfiguration siteConfig, Opts opts, String rootUser) {
     if (opts.cliPassword != null) {
       return opts.cliPassword.getBytes(UTF_8);
     }
-    char[] rootpass;
-    char[] confirmpass;
     String strrootpass;
     String strconfirmpass;
     do {
-      rootpass = System.console().readPassword(
+      var rootpass = System.console().readPassword(
           "Enter initial password for " + rootUser + getInitialPasswordWarning(siteConfig));
       if (rootpass == null) {
         System.exit(0);
       }
-      confirmpass = System.console().readPassword("Confirm initial password for " + rootUser + ":");
+      var confirmpass =
+          System.console().readPassword("Confirm initial password for " + rootUser + ":");
       if (confirmpass == null) {
         System.exit(0);
       }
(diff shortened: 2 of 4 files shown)


Manno15:
It looks like JLine is still listed as a dependency in the core package. With these changes, JLine can be removed from the core's pom file.

harjitdotsingh:
oops . Let me fix that...

Manno15:
It seems you still have some references to JLine inside `Initialize.java` that should be removed. Then the pom changes can be made to server/base.

harjitdotsingh:
We still have this lines. I don't know if there is a equivalent of these in
System.console().

if (terminal == null)
  terminal = TerminalBuilder.builder().jansi(false).build();
if (reader == null)
  reader = LineReaderBuilder.builder().terminal(terminal).build();
return reader;

Manno15:
Those just build the terminal objects. You shouldn't need those anymore since you are using the console to print instead of the reader object from JLine

server/base/pom.xml
```diff
@@ -120,6 +116,11 @@
       <artifactId>easymock</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jline</groupId>
+      <artifactId>jline</artifactId>
+      <scope>test</scope>
+    </dependency>
```
Manno15, line 123:
```suggestion

```

Manno15:
It appears System.console() will always return null for our tests. Especially in an IDE. https://stackoverflow.com/questions/4644415/java-how-to-get-input-from-system-console has some reasoning into why and has alternative solutions. Though I wonder for our use case, a simple System.out.println() would suffice.

harjitdotsingh:
Thanks, that what I wasn't able to understand why it was failing with a
nullpointerexception. That helps

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
```diff
@@ -279,9 +267,8 @@ static boolean checkInit(VolumeManager fs, SiteConfiguration sconf, Configuratio
       return false;
     }
     if (sconf.get(Property.INSTANCE_SECRET).equals(Property.INSTANCE_SECRET.getDefaultValue())) {
-      LineReader c = getLineReader();
-      var w = c.getTerminal().writer();
-      c.getTerminal().puts(InfoCmp.Capability.bell);
+      Console c = System.console();
```
Manno15, line 270: I was more thinking of something like this: 

```java
      System.out.println();
      System.out.println();
      System.out.println("Warning!!! Your instance secret is still set to the default,"
          + " this is not secure. We highly recommend you change it.");
      System.out.println();
      System.out.println();
      System.out.println("You can change the instance secret in accumulo by using:");
      System.out.println("   bin/accumulo " + org.apache.accumulo.server.util.ChangeSecret.class.getName());
      System.out.println("You will also need to edit your secret in your configuration"
          + " file by adding the property instance.secret to your"
          + " accumulo.properties. Without this accumulo will not operate" + " correctly");


```

since all this is trying to do is print those messages to screen.

DomGarguilo: Would we also be able to do `System.out.println("\n");` instead of two seprate println's? Or even appending two newlines (\n) to the preceding statements?

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
```diff
@@ -748,24 +735,27 @@ private String getRootUserName(SiteConfiguration siteConfig, Opts opts) throws I
...
       if (confirmpass == null) {
         System.exit(0);
       }
-      if (!rootpass.equals(confirmpass)) {
+      strrootpass = new String(rootpass);
+      strconfirmpass = new String(confirmpass);
+      if (!strrootpass.equals(strconfirmpass)) {
         log.error("Passwords do not match");
       }
     } while (!rootpass.equals(confirmpass));
```
Manno15, line 757: This should probably be the comparison of the string version otherwise, you would need Arrays.,equals

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
```diff
@@ -144,17 +141,8 @@
...
   private static ZooReaderWriter zoo = null;
-
-  private static LineReader getLineReader() throws IOException {
-    if (terminal == null)
-      terminal = TerminalBuilder.builder().jansi(false).build();
-    if (reader == null)
-      reader = LineReaderBuilder.builder().terminal(terminal).build();
-    return reader;
-  }
+  private static Console reader = null;
```
Manno15, line 145:
```suggestion

```

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
```diff
@@ -748,24 +735,27 @@ private String getRootUserName(SiteConfiguration siteConfig, Opts opts) throws I
...
       }
-      if (!rootpass.equals(confirmpass)) {
+      strrootpass = new String(rootpass);
+      strconfirmpass = new String(confirmpass);
+      if (!strrootpass.equals(strconfirmpass)) {
         log.error("Passwords do not match");
       }
     } while (!rootpass.equals(confirmpass));
-    return rootpass.getBytes(UTF_8);
+    return strrootpass.getBytes(Charset.forName("UTF-8"));
```
Manno15, line 758:
```suggestion
    return strrootpass.getBytes(UTF_8);
```

Manno15 (changes requested):

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
Manno15: Probably could do some cleanup here, yes.

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
harjitdotsingh: I have already fixed that

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
harjitdotsingh: If I do strrootpass.getBytes(UTF_8) It fails the quality check. Recommend way is the way I have it  strrootpass.getBytes(Charset.forName("UTF-8"));

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
harjitdotsingh: \n is unix if we do it the way  it is, its compatible with any other systems you run on. I suggest we live it like that

DomGarguilo:
Looks like the files are not formatted. Try running `mvn package -DskipTests` and including the changes made.

harjitdotsingh:
I did that and pushed. I don't know why it didn't format it the first time

- Harjit

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
DomGarguilo: The tests pass for me with this change included, perhaps something else caused the test failures.

Manno15:
There appears to be an issue with GitHub actions at the moment. I apologise for the delay in the build @harjitdotsingh

harjitdotsingh:
I hope I didn't break it :-)

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
Manno15: I recommended this change since that was there previously and IntelliJ actually gives a warning with the way you have it. We are already importing the StandardCharset for UTF-8 and using it elsewhere so this part should follow the precedent already set.

Manno15:
I did run the tests locally and they all pass.

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
```diff
@@ -724,8 +709,8 @@ private String getRootUserName(SiteConfiguration siteConfig, Opts opts) throws I
       return DEFAULT_ROOT_USER;
     }
 
-    LineReader c = getLineReader();
-    c.getTerminal().writer().println("Running against secured HDFS");
+    Console c = System.console();
```
Manno15, line 712:
```suggestion
    
```

Can drop this and instead do 

```
System.console().readLine("Principal (user) to grant administrative privileges to : "); 
```
below where c is used (it is used only once).

Manno15 (changes requested):
These changes look good, just a few small things left. There are various IOExceptions that can be removed now.

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
```diff
@@ -748,24 +733,27 @@ private String getRootUserName(SiteConfiguration siteConfig, Opts opts) throws I
...
-    String confirmpass;
+    char[] rootpass;
+    char[] confirmpass;
+    String strrootpass;
+    String strconfirmpass;
     do {
-      rootpass = getLineReader().readLine(
-          "Enter initial password for " + rootUser + getInitialPasswordWarning(siteConfig), '*');
+      rootpass = System.console().readPassword(
+          "Enter initial password for " + rootUser + getInitialPasswordWarning(siteConfig));
```
Manno15, line 742:
```suggestion
     var rootpass = System.console().readPassword(
          "Enter initial password for " + rootUser + getInitialPasswordWarning(siteConfig));
```

Could do something like this for both passwords and drop the char[] objects above.

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
harjitdotsingh: Cleaned it up.

server/base/src/main/java/org/apache/accumulo/server/init/Initialize.java
harjitdotsingh: Done.

Manno15 (approved):
These changes look good. Thanks for the contribution @harjitdotsingh!

harjitdotsingh:
Thanks  For all your help and feedback.  Do I have to close this to be merged or somebody from the team would do that ?

Manno15:
I will give more time for others to give any feedback then either I or someone else will merge it in. You do not have to do anything at the moment.

PR #2231, opened 2021-08-12
Rework sorted WAL Upgrade code

milleruntime:
* Follow on work for #2185
* Add support for multiple volumes to upgrade code in 
Upgrader9to10 dropSortedMapWALFiles() and move
method call to upgradeZookeeper

### DIFF
--- a/server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
+++ b/server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
@@ -71,6 +71,7 @@ import org.apache.accumulo.core.security.Authorizations;
 import org.apache.accumulo.core.spi.compaction.SimpleCompactionDispatcher;
 import org.apache.accumulo.core.tabletserver.log.LogEntry;
 import org.apache.accumulo.core.util.HostAndPort;
+import org.apache.accumulo.core.volume.Volume;
 import org.apache.accumulo.fate.zookeeper.ZooReaderWriter;
 import org.apache.accumulo.fate.zookeeper.ZooUtil.NodeExistsPolicy;
 import org.apache.accumulo.fate.zookeeper.ZooUtil.NodeMissingPolicy;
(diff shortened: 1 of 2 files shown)


server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -738,33 +739,36 @@ static Path resolveRelativeDelete(String oldDelete, String upgradeProperty) {
...
-        // map files will be in a directory starting with "part"
-        Path walDirPath = walDir.getPath();
-        for (FileStatus dirOrFile : vm.listStatus(walDirPath)) {
-          if (dirOrFile.isDirectory()) {
-            directoriesToDrop.add(walDirPath);
-            break;
+  static void dropSortedMapWALFiles(ServerContext context) {
+    VolumeManager vm = context.getVolumeManager();
+    for (Volume volume : vm.getVolumes()) {
+      Path recoveryDir = volume.prefixChild("/accumulo/recovery");
```
keith-turner, line 745: I suspect `/accumulo` is not needed as it will be [part of the volume](https://github.com/apache/accumulo/blob/7db0561cac3d42ba09d1fb038dad1ff1a03afc46/assemble/conf/accumulo.properties#L24).  Also there is constant `RECOVERY_DIR` in ServerConstants that could be used.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
```diff
@@ -738,33 +739,36 @@ static Path resolveRelativeDelete(String oldDelete, String upgradeProperty) {
...
+      try {
+        if (!vm.exists(recoveryDir)) {
+          log.info("There are no recovery files in /accumulo/recovery");
```
keith-turner, line 748: Could include the volume in the log message.

server/manager/src/main/java/org/apache/accumulo/manager/upgrade/Upgrader9to10.java
milleruntime: Right, forgot about that. I think an IT would have caught that but unfortunately I think it's too involved to try and recreate. I will test with Uno one more time after we think the changes are solid.

milleruntime: Even better, I think I can use the method in `ServerConstants` to get the directories. Check it out: https://github.com/apache/accumulo/blob/618ba7a9f88342ea7c8d692ce19df34d88bf07c8/server/base/src/main/java/org/apache/accumulo/server/ServerConstants.java#L169

milleruntime:
I tested an upgrade from 2.0.1 to Snapshot still works using Uno just using a single volume and it removed the old directories properly. @keith-turner I think this is ready to merge if you want to take another quick look at it.

PR #2233, opened 2021-08-16
Drop ServerUtil and static state

milleruntime:
* Move static state from ServerUtil in to ServerConstants object methods
* Move creation of ServerConstants object into ServerInfo
* Move upgrade code from ServerUtil into UpgradeCoordinator
* Move methods dealing with server stuff to ServerContext
* Create getter for ServerConstants in ServerContext
* Move AccumuloTest to manager to avoid adding a dependency
* Drop ServerUtilTest and move testing of method to ServerContextTest


server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -279,4 +300,139 @@ public Ample getAmple() {
...
+    log.info("Data Version {}", dataVersion);
+    waitForZookeeperAndHdfs();
+
+    ensureDataVersionCompatible(dataVersion);
+
+    TreeMap<String,String> sortedProps = new TreeMap<>();
+    for (Map.Entry<String,String> entry : conf)
+      sortedProps.put(entry.getKey(), entry.getValue());
+
+    for (Map.Entry<String,String> entry : sortedProps.entrySet()) {
```
EdColeman, line 385: Is the sort for readability so that the messages reported are consistently ordered? Otherwise, it looks like the check could be done on iterating through the conf entries themselves.

Also, if the sort is desired, would
```
TreeMap<String,String> sortedProps = new TreeMap(conf);
```
work?

EdColeman (approved):
LGTM - my comments can be deferred if this code is being moved and additional changes are not wanted with this PR.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
milleruntime: > Is the sort for readability so that the messages reported are consistently ordered? 

I think so. That `sortedProps` is only used in that for loop to log information about the properties.

> Also, if the sort is desired, would
TreeMap<String,String> sortedProps = new TreeMap(conf);
work?

No, it doesn't. TreeMap constructor can take a map so we would have to get the properties map, which we can only do using `getAllPropertiesWithPrefix()`. It is odd we don't have a method to get all properties.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -279,4 +300,139 @@ public Ample getAmple() {
...
+
+  public void waitForZookeeperAndHdfs() {
+    log.info("Attempting to talk to zookeeper");
+    while (true) {
+      try {
+        getZooReaderWriter().getChildren(Constants.ZROOT);
+        break;
+      } catch (InterruptedException e) {
+        // ignored
+      } catch (KeeperException ex) {
```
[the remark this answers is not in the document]
milleruntime: I am not sure. This code was just moved. This would be good to look into thought.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
```diff
@@ -279,4 +300,137 @@ public Ample getAmple() {
...
+  public static void ensureDataVersionCompatible(int dataVersion) {
+    if (!(ServerConstants.CAN_RUN.contains(dataVersion))) {
+      throw new IllegalStateException("This version of accumulo (" + Constants.VERSION
+          + ") is not compatible with files stored using data version " + dataVersion);
+    }
+  }
+
+  public void waitForZookeeperAndHdfs() {
+    log.info("Attempting to talk to zookeeper");
+    while (true) {
```
EdColeman, line 316: This might be clearer if the try was lifted above the while and a local boolean used to terminate the while.  The code is fine as is - I just don't use break much and it gives me a pause when I see it - more of a personal preference.

EdColeman (approved):
LGTM - open comments are more of my personnel preference and can be ignored if not desired.

server/base/src/main/java/org/apache/accumulo/server/ServerContext.java
milleruntime: I agree. I looked at this and I am not even sure if getChildren() will return null or empty list so I am going to just leave it.

PR #2234, opened 2021-08-17
Create AccumuloDataVersion

milleruntime:
* Move data version from ServerConstants to AccumuloDataVersion and added
javadoc description
* Create AccumuloDataVersion.get() for getting data version
* Move other constants from ServerConstants to Constants
* Rename ServerConstants to ServerDirs

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/Constants.java
+++ b/core/src/main/java/org/apache/accumulo/core/Constants.java
@@ -21,9 +21,11 @@ package org.apache.accumulo.core;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 public class Constants {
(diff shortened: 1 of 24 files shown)


core/src/main/java/org/apache/accumulo/core/Constants.java
```diff
@@ -23,6 +23,11 @@
 public class Constants {
```
EdColeman, line 24:
```suggestion
   // defines Accumulo data version constants
```

EdColeman (approved):
LGTM - I really think this helps distinguishes what version means in context.

EdColeman (approved):

PR #2235, opened 2021-08-19
Correct chop behavior

### LINKED ISSUE #2222: Tablet may write chop information to the metadata table after its closed.
**Describe the bug**

When CompactableImpl writes chop information [here](https://github.com/apache/accumulo/blob/a9100ac35f3d72677b767d7670dfb032c7e7d733/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java#L691) it could happen after [close()](https://github.com/apache/accumulo/blob/a9100ac35f3d72677b767d7670dfb032c7e7d733/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java#L1384) was called.  When a tablet calls `CompactableImpl.close()` it expects the call to block until all ongoing metadata writes related to compactions are complete AND prevent any new writes from occurring after the `close()` call returns.  This expectation is not being met when writing chop information to the metadata table.

**Expected behavior**

`CompactableImpl.close()` will block if there are any ongoing metadata chop writes and prevent any new ones from starting after it returns.

DomGarguilo:
Fixes #2222 

There were two bugs in the chop behavior that these changes attempt to correct. The first being chops could be written after close is called. The second being new chops may be started after close returns.

To account for these, a check is added to wait until the chop status becomes `NOT_ACTIVE` before close returns as well as a check to ensure close has not returned before a chop is written.


server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -1388,11 +1388,12 @@ public synchronized void close() {
...
 
-      // wait while internal jobs are running or external compactions are committing, but do not
-      // wait on external compactions that are running
+      // wait while internal jobs are running, external compactions are committing or the status of
+      // chops is active, but do not wait on external compactions that are running
       while (runningJobs.stream()
           .anyMatch(job -> !((CompactionExecutorIdImpl) job.getExecutor()).isExternalId())
-          || !externalCompactionsCommitting.isEmpty()) {
+          || !externalCompactionsCommitting.isEmpty()
+          || fileMgr.chopStatus != FileSelectionStatus.NOT_ACTIVE) {
```
keith-turner, line 1396: I think in the current code when a chop is actually marking the state will actually be NOT_ACTIVE.  Thinking we should introduce a new explicit state for MARKING and then we could do something like the following.

```suggestion
          || fileMgr.chopStatus == FileSelectionStatus.MARKING) {
```

To go with the above change could do something like the following in checkIfChopComplete()

```java
  private void checkifChopComplete(Set<StoredTabletFile> allFiles) {

    boolean completed = false;

    synchronized (this) {
      if (closed) {
       // if closed do not attempt transition to marking state
        return;
     }
      // when this returns true it means we transitioned to the MARKING state
      completed = fileMgr.finishChop(allFiles);
    }

    if (completed) {
      try {
        markChopped();
      } finally {
        synchronized (this) {
          // this will transition from MARKING to NOT_ACTIVE
          fileMgr.finishMarkingChop();
          this.notifyAll();
        }
      }

      TabletLogger.selected(getExtent(), CompactionKind.CHOP, Set.of());
    }
  }
```

In a [branch](https://github.com/keith-turner/accumulo/commits/accumulo-2199) I created for #2199 I have created a separate enum for chop status since chop only uses a subset of the file status states.  Adding an extra state that is only used for chop would make this separate enum even more useful.  For this PR could add the new state to the existing enum and I can fix it up later.

DomGarguilo: > > For this PR could add the new state to the existing enum and I can fix it up later.
> 
> Okay I'll go ahead and do that here

@keith-turner, Actually do you think that the new state should just be added with the changes that will use it instead of adding it here before it is used anywhere? It seems like this PR could just be closed if there are plans to account for this later on.

keith-turner: For the work on #2199 I would not be adding a new MARKING state, only a new enum specific to chop states.  Still need to make these changes. These changes could wait until the new enum is added and add the new state to that enum as that will be cleaner.

DomGarguilo: @keith-turner, I tried incorporating your suggestions in a9c0f9b. Not too sure if they are what you had in mind.

keith-turner: The changes look good. Need to see if the new state affects existing code.

DomGarguilo: I wasn't too sure how to best test to see if these changes affect existing code so I ended up running all the ITs on this branch. Everything looks good (except a failing ConcurrentDeleteTableIT which is known flaky and also fails on main so is probably unrelated. Might create a ticket for this.).

I think these changes are ready for review @keith-turner

keith-turner:
@DomGarguilo in there is a method named `getCandidates()` that has the following bit of code in it.

```java
        case CHOP: {
          switch (chopStatus) {
            case NOT_ACTIVE:
            case SELECTING:
              return Set.of();
```

need to add the new enum to this switch like the following.

```java
        case CHOP: {
          switch (chopStatus) {
            case NOT_ACTIVE:
            case SELECTING:
            case MARKING:
              return Set.of();
```

This is the code that detemines which files are candidates for a chop compaction.  Unless chopStatus is SELECTED it should return the empty set.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -1447,11 +1470,12 @@ public synchronized void close() {
...
 
-      // wait while internal jobs are running or external compactions are committing, but do not
-      // wait on external compactions that are running
+      // wait while internal jobs are running, external compactions are committing or the status of
+      // chops is active, but do not wait on external compactions that are running
       while (runningJobs.stream()
           .anyMatch(job -> !((CompactionExecutorIdImpl) job.getExecutor()).isExternalId())
-          || !externalCompactionsCommitting.isEmpty()) {
+          || !externalCompactionsCommitting.isEmpty()
+          || fileMgr.chopStatus != ChopSelectionStatus.NOT_ACTIVE) {
```
keith-turner, line 1478: Need to make this check more narrow.  After closed is set to true above, its not possible to enter the marking state. So we only want to wait on things that were in the marking state before closed was set to true.
  
```suggestion
          || fileMgr.chopStatus == ChopSelectionStatus.MARKING) {
```

DomGarguilo: Oops, I had thought I included this change earlier. Just included in 2e3d97e

keith-turner (approved):

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -739,11 +749,25 @@ private void checkifChopComplete(Set<StoredTabletFile> allFiles) {
...
     }
 
     if (completed) {
-      markChopped();
+      try {
+        markChopped();
+      } finally {
+        synchronized (this) {
+          // transition the state from MARKING to NOT_ACTIVE
+          fileMgr.finishMarkingChop();
```
milleruntime, line 766: Could you do something with the boolean that is returned from this method? It might not matter but since you are checking the state and returning an outcome, maybe a false should be handled.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -111,7 +111,7 @@
   private Set<CompactionServiceId> servicesUsed = new ConcurrentSkipListSet<>();
 
   enum ChopSelectionStatus {
-    SELECTING, SELECTED, NOT_ACTIVE
+    SELECTING, SELECTED, NOT_ACTIVE, MARKING
```
milleruntime, line 114: It would be nice to have a description of the different states. Or if that isn't helpful, at least a description of the state to state transitions.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
keith-turner: Could remove the boolean and change the method to have a precondition that expects the current chop state to be MARKING.  I made a suggested change for the method.  The way the code is currently written, it would be unexpected for the chop state to be anything else when this method is called.

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -347,6 +347,15 @@ boolean finishChop(Set<StoredTabletFile> allFiles) {
...
     }
 
+    boolean finishMarkingChop() {
+      if (chopStatus == ChopSelectionStatus.MARKING) {
+        chopStatus = ChopSelectionStatus.NOT_ACTIVE;
+        return true;
+      } else {
+        return false;
+      }
+    }
```
keith-turner, line 357:
```suggestion
    void finishMarkingChop() {
        Preconditions.checkState(chopStatus == ChopSelectionStatus.MARKING);
        chopStatus = ChopSelectionStatus.NOT_ACTIVE;
    }
```

milleruntime (changes requested):
@DomGarguilo it might make the test tricker but I think what @keith-turner suggested for the finishMarkingChop() method would be better.

server/tserver/src/test/java/org/apache/accumulo/tserver/tablet/CompactableImplFileManagerTest.java
```diff
@@ -379,11 +379,15 @@ public void testChop() {
     tabletFiles = newFiles("C00004.rf", "C00006.rf", "F00003.rf", "F00004.rf");
     assertFalse(fileMgr.finishChop(tabletFiles));
     assertEquals(ChopSelectionStatus.SELECTED, fileMgr.getChopStatus());
-    assertFalse(fileMgr.finishMarkingChop());
+    try {
+      fileMgr.finishMarkingChop();
+    } catch (IllegalStateException e) {
+      // expected
+    }
```
keith-turner, line 386: Could use the assertThrows functions from junit here

keith-turner (approved):

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
```diff
@@ -739,11 +745,25 @@ private void checkifChopComplete(Set<StoredTabletFile> allFiles) {
     boolean completed;
 
     synchronized (this) {
+      if (closed) {
+        // if closed, do not attempt to transition to the MARKING state
+        return;
+      }
+      // when this returns true it means we transitioned to the MARKING state
```
milleruntime, line 752: This comment is outdated now.

milleruntime: Sorry, nevermind this is calling the other method.

milleruntime (approved):

PR #2236, opened 2021-08-19
Add new property to configure rfile sorted recovery

### LINKED ISSUE #2187: Sorted Log Recovery configuration design

milleruntime:
* Closes #2187 
* Create new property prefix "tserver.wal.sort.file." to configure the
rfiles written during sorted recovery
* Add method to LogSorter to convert the sort file properties to table
files properties
* Create new tests in SortedLogRecoveryTest
* Make method public in Compression to use in test
* Add property to MultiTableRecoveryIT to allow testing in an IT

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/conf/Property.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/Property.java
@@ -675,6 +675,11 @@ public enum Property {
       "The maximum number of threads to use to sort logs during" + " recovery", "1.5.0"),
   TSERV_SORT_BUFFER_SIZE("tserver.sort.buffer.size", "10%", PropertyType.MEMORY,
       "The amount of memory to use when sorting logs during recovery.", "1.5.0"),
+  TSERV_WAL_SORT_FILE_PREFIX("tserver.wal.sort.file.", null, PropertyType.PREFIX,
+      "The rfile properties to use when sorting logs during recovery. Most of the properties"
+          + " that begin with 'table.file' can be used here. For example, to set the compression"
+          + " of the sorted recovery files to snappy use 'tserver.sort.file.compress.type=snappy'",
+      "2.1.0"),
   TSERV_WORKQ_THREADS("tserver.workq.threads", "2", PropertyType.COUNT,
       "The number of threads for the distributed work queue. These threads are"
           + " used for copying failed bulk import RFiles.",
--- a/core/src/main/java/org/apache/accumulo/core/conf/Property.java
+++ b/core/src/main/java/org/apache/accumulo/core/conf/Property.java
@@ -678,7 +678,7 @@ public enum Property {
   TSERV_WAL_SORT_FILE_PREFIX("tserver.wal.sort.file.", null, PropertyType.PREFIX,
       "The rfile properties to use when sorting logs during recovery. Most of the properties"
           + " that begin with 'table.file' can be used here. For example, to set the compression"
-          + " of the sorted recovery files to snappy use 'tserver.sort.file.compress.type=snappy'",
+          + " of the sorted recovery files to snappy use 'tserver.wal.sort.file.compress.type=snappy'",
       "2.1.0"),
   TSERV_WORKQ_THREADS("tserver.workq.threads", "2", PropertyType.COUNT,
       "The number of threads for the distributed work queue. These threads are"
--- a/server/tserver/src/test/java/org/apache/accumulo/tserver/log/SortedLogRecoveryTest.java
+++ b/server/tserver/src/test/java/org/apache/accumulo/tserver/log/SortedLogRecoveryTest.java
@@ -1118,7 +1118,7 @@ public class SortedLogRecoveryTest {
               createKeyValue(COMPACTION_FINISH, 2, 1, null),
               createKeyValue(COMPACTION_START, 4, 1, "/t1/f1"),
               createKeyValue(COMPACTION_FINISH, 5, 1, null),
-              createKeyValue(MUTATION, 3, 1, ignored), createKeyValue(MUTATION, 5, 1, m),};
+              createKeyValue(MUTATION, 3, 1, ignored), createKeyValue(MUTATION, 5, 1, m)};
       String dest = workdir + "/testLogSortedProperties";
 
       List<Pair<LogFileKey,LogFileValue>> buffer = new ArrayList<>();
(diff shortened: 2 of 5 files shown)


core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -675,6 +675,11 @@
       "The maximum number of threads to use to sort logs during" + " recovery", "1.5.0"),
   TSERV_SORT_BUFFER_SIZE("tserver.sort.buffer.size", "10%", PropertyType.MEMORY,
       "The amount of memory to use when sorting logs during recovery.", "1.5.0"),
+  TSERV_WAL_SORT_FILE_PREFIX("tserver.wal.sort.file.", null, PropertyType.PREFIX,
+      "The rfile properties to use when sorting logs during recovery. Most of the properties"
+          + " that begin with 'table.file' can be used here. For example, to set the compression"
```
Manno15, line 680: Do you know of 'table.file' props that can't be used here?

milleruntime: Yeah but I think they will just get ignored. Properties like "table.file.max" or "table.file.summary.maxSize" that don't apply to these files.

dlmarion (approved):

server/tserver/src/test/java/org/apache/accumulo/tserver/log/SortedLogRecoveryTest.java
```diff
@@ -1064,4 +1070,110 @@ public void testFileWithoutOpen() throws IOException {
...
+      replay(context);
+      final Path workdirPath = new Path("file://" + workdir);
+      vm.deleteRecursively(workdirPath);
+
+      KeyValue[] events =
+          {createKeyValue(OPEN, 0, -1, "1"), createKeyValue(DEFINE_TABLET, 1, 1, extent),
+              createKeyValue(COMPACTION_FINISH, 2, 1, null),
+              createKeyValue(COMPACTION_START, 4, 1, "/t1/f1"),
+              createKeyValue(COMPACTION_FINISH, 5, 1, null),
+              createKeyValue(MUTATION, 3, 1, ignored), createKeyValue(MUTATION, 5, 1, m),};
```
DomGarguilo, line 1121:
```suggestion
              createKeyValue(MUTATION, 3, 1, ignored), createKeyValue(MUTATION, 5, 1, m)};
```
Not sure if the extra comma was intentional

milleruntime: Copy and paste error. Thanks!

DomGarguilo: Yea it looks like most if not all of the `KeyValue[]` have it as well for some reason :confused:

milleruntime:
Tested the most useful option, setting the compression on the recovery file in Uno. It worked well but I noticed that the sort recovery options are named differently. I think it would be good to name them all the same. I can do that as a follow on.
<pre>
09:35:49 {main} ~/workspace/uno/install/logs/accumulo$ hdfs dfs -ls -R /accumulo/recovery
drwx------   - mike supergroup          0 2021-08-20 09:35 /accumulo/recovery/5f821693-2a9e-4b1f-ab35-9b4542764ff0
-rw-r--r--   3 mike supergroup          0 2021-08-20 09:35 /accumulo/recovery/5f821693-2a9e-4b1f-ab35-9b4542764ff0/finished
-rw-r--r--   3 mike supergroup        334 2021-08-20 09:35 /accumulo/recovery/5f821693-2a9e-4b1f-ab35-9b4542764ff0/part-r-00000.rf
drwx------   - mike supergroup          0 2021-08-20 09:35 /accumulo/recovery/cdbf6848-11c4-4864-b149-da02fd4543ed
-rw-r--r--   3 mike supergroup          0 2021-08-20 09:35 /accumulo/recovery/cdbf6848-11c4-4864-b149-da02fd4543ed/finished
-rw-r--r--   3 mike supergroup      20182 2021-08-20 09:35 /accumulo/recovery/cdbf6848-11c4-4864-b149-da02fd4543ed/part-r-00000.rf
09:36:16 {main} ~/workspace/uno/install/logs/accumulo$ accumulo rfile-info /accumulo/recovery/cdbf6848-11c4-4864-b149-da02fd4543ed/part-r-00000.rf
Reading file: hdfs://localhost:8020/accumulo/recovery/cdbf6848-11c4-4864-b149-da02fd4543ed/part-r-00000.rf
No on disk encryption detected.
RFile Version            : 8
Locality group           : <DEFAULT>
	Num   blocks           : 1
	Index level 0          : 56 bytes  1 blocks
	First key              : %00;%80;%00;%00;%00;%00;%00;%00;%00;%00;%00;%00;%00; OPEN:cdbf6848-11c4-4864-b149-da02fd4543ed [] 9223372036854775807 false
	Last key               : %03;%80;%00;%00;%0f;%00;%00;%00;%00;%00;%00;%00;%03; MANY_MUTATIONS: [] 9223372036854775807 false
	Num entries            : 54
	Column families        : [MANY_MUTATIONS, COMPACTION_FINISH, DEFINE_TABLET, COMPACTION_START, OPEN, MUTATION]

Meta block     : BCFile.index
      Raw size             : 8 bytes
      Compressed size      : 18 bytes
      Compression type     : snappy

Meta block     : RFile.index
      Raw size             : 307 bytes
      Compressed size      : 236 bytes
      Compression type     : snappy
</pre>

PR #2241, opened 2021-08-24
Rename all ServerContext variables to context

milleruntime:


EdColeman (commented):
Would it be better if it was something more like srvContext ?  It looks like we have other contexts (ClientContext, ClassLoaderContext) as well as M/R contexts. It may be that ServerContext and ClientContext are not used together - but with using 'context' anyone should not assume that all instances of context are server contexts.

'context' does read better, but wondering if there are benefits to being more specific to add clarity or if it just makes it harder to read for something that is already obvious.

EdColeman (approved):
Changes look fine - still wonder if something besides context would improve comprehension, but not a blocker.

milleruntime:
> Would it be better if it was something more like srvContext ? It looks like we have other contexts (ClientContext, ClassLoaderContext) as well as M/R contexts. It may be that ServerContext and ClientContext are not used together - but with using 'context' anyone should not assume that all instances of context are server contexts.
> 
> 'context' does read better, but wondering if there are benefits to being more specific to add clarity or if it just makes it harder to read for something that is already obvious.

The other context objects are only used in specific classes. I went with the name that was most popular but I am not opposed to the more specific name `serverContext`. It would just require more changes.

milleruntime:
Just comparing what is already in the code... To change the name to `serverContext` would require changing 147 files vs `context`  in this PR only changes 23. The name `context` is used in MR classes for `JobContext`, for a private Context object in ContextManager, both of which are deprecated. But it is used for ClientContext in about 100 files and for a String type in `ContextClassLoaderFactory`, which is new. 

The way that the String type is used in `ContextClassLoaderFactory`, makes me think that it is the only object that should use the `context` variable name. As opposed to everywhere else we have specific types of context objects that we can be more specific (ClientContext, ServerContext, JobContext, etc).

milleruntime:
I started renaming `context` to `serverContext` but it is a lot more work. It would be nice to have both consistency and specificity, but I am going to go with this PR so we at least have consistency.

PR #2251, opened 2021-08-31
Create inner class in CompactionService

milleruntime:
* Improve readability of CompactionService by creating inner class
for the implementation of PlanningParameters


Manno15 (approved):
Changes seem reasonable to me.

PR #2255, opened 2021-09-01
Added Fix for #2227

### LINKED ISSUE #2227: Add support for basic escaping of characters inside Shell commands
Original Jira ticket: https://issues.apache.org/jira/browse/ACCUMULO-2929

Excerpt from ticket: 
```
For commands that have a -c option (e.g. InsertCommand, GrepCommand, ScanCommand),
 there is currently no way to specify a column family that contains a colon as it is
 interpreted as the delimiter between the column family and column qualifier 'cf[:cq]'.

It would be very simple to support a basic escape '\' character that would allow 
users to use column families which contain colons in the shell.
```

It would be a useful feature to allow escaping of special characters, specifically a ' : ' in this instance. Alternatively, extra options such as -cf and -cq could be added instead.

**Additional context**
The specific function inside `ScanCommand` (utilized by `GrepCommand` and others) that fetches the Columns. 
https://github.com/apache/accumulo/blob/3da639ac6c49ad1b9671230e3443632d27887f75/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java#L274-L275

harjitdotsingh:
Fixes #2227

### DIFF
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/QuotedStringTokenizer.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/QuotedStringTokenizer.java
@@ -72,7 +72,7 @@ public class QuotedStringTokenizer implements Iterable<String> {
         inEscapeSequence = false;
         if (ch == 'x') {
           hexChars = "";
-        } else if (ch == ' ' || ch == '\'' || ch == '"' || ch == '\\' || ch == ':') {
+        } else if (ch == ' ' || ch == '\'' || ch == '"' || ch == '\\') {
           token[tokenLength++] = inputBytes[i];
         } else {
           throw new BadArgumentException("can only escape single quotes, double"
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/QuotedStringTokenizer.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/QuotedStringTokenizer.java
@@ -72,7 +72,7 @@ public class QuotedStringTokenizer implements Iterable<String> {
         inEscapeSequence = false;
         if (ch == 'x') {
           hexChars = "";
-        } else if (ch == ' ' || ch == '\'' || ch == '"' || ch == '\\') {
+        } else if (ch == ' ' || ch == '\'' || ch == '"' || ch == '\\' || ch == ':') {
           token[tokenLength++] = inputBytes[i];
         } else {
           throw new BadArgumentException("can only escape single quotes, double"
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
@@ -290,7 +290,7 @@ public class ScanCommand extends Command {
 
   protected String[] extractColumnFamily(final String columnString) {
     String[] columnFamily = new String[2];
-    if (CharMatcher.is(':').countIn(columnString) == 2) {
+    if (CharMatcher.is(':').countIn(columnString) >= 2) {
       columnFamily[0] = columnString.substring(0, columnString.lastIndexOf(":"));
       columnFamily[1] = columnString.substring(columnString.lastIndexOf(":") + 1);
       return columnFamily;
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
@@ -56,12 +56,11 @@ import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 import org.apache.hadoop.io.Text;
 
-import com.google.common.base.CharMatcher;
-
 public class ScanCommand extends Command {
 
   private Option scanOptAuths, scanOptRow, scanOptColumns, disablePaginationOpt, showFewOpt,
-      formatterOpt, interpreterOpt, formatterInterpeterOpt, outputFileOpt;
+      formatterOpt, interpreterOpt, formatterInterpeterOpt, outputFileOpt, scanOptCfOptions,
+      scanOptColQualifier;
 
   protected Option timestampOpt;
   protected Option profileOpt;
@@ -116,6 +115,7 @@ public class ScanCommand extends Command {
 
       // handle columns
       fetchColumns(cl, scanner, interpeter);
+      fetchColumsWithCFAndCQ(cl, scanner, interpeter);
 
       // set timeout
       scanner.setTimeout(getTimeout(cl), TimeUnit.MILLISECONDS);
@@ -273,9 +273,18 @@ public class ScanCommand extends Command {
   protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
       final ScanInterpreter formatter) throws UnsupportedEncodingException {
 
+    if ((cl.hasOption(scanOptCfOptions.getOpt()) || cl.hasOption(scanOptColQualifier.getOpt()))
+        && cl.hasOption(scanOptColumns.getOpt())) {
+
+      String formattedString =
+          String.format("Options - %s AND (- %s" + "OR - %s are mutually exclusive ",
+              scanOptColumns.getOpt(), scanOptCfOptions.getOpt(), scanOptColQualifier.getOpt());
+      throw new IllegalArgumentException(formattedString);
+    }
+
     if (cl.hasOption(scanOptColumns.getOpt())) {
       for (String a : cl.getOptionValue(scanOptColumns.getOpt()).split(",")) {
-        final String[] sa = extractColumnFamily(a);
+        final String[] sa = a.split(":", 2);
         if (sa.length == 1) {
           scanner.fetchColumnFamily(
               formatter.interpretColumnFamily(new Text(a.getBytes(Shell.CHARSET))));
@@ -288,14 +297,28 @@ public class ScanCommand extends Command {
     }
   }
 
-  protected String[] extractColumnFamily(final String columnString) {
-    String[] columnFamily = new String[2];
-    if (CharMatcher.is(':').countIn(columnString) >= 2) {
-      columnFamily[0] = columnString.substring(0, columnString.lastIndexOf(":"));
-      columnFamily[1] = columnString.substring(columnString.lastIndexOf(":") + 1);
-      return columnFamily;
-    } else {
-      return columnString.split(":", 2);
+  private void fetchColumsWithCFAndCQ(CommandLine cl, Scanner scanner, ScanInterpreter interpeter) {
+    String cf = "";
+    String cq = "";
+    if (cl.hasOption(scanOptCfOptions.getOpt())) {
+      cf = cl.getOptionValue(scanOptCfOptions.getOpt());
+    }
+    if (cl.hasOption(scanOptColQualifier.getOpt())) {
+      cq = cl.getOptionValue(scanOptColQualifier.getOpt());
+    }
+
+    if (cf.isEmpty() && !cq.isEmpty()) {
+      String formattedString = String.format(
+          "Option - %s when used with (- %s" + "cannot be empty ", scanOptCfOptions.getOpt(),
+          scanOptColQualifier.getOpt(), scanOptColQualifier.getOpt());
+      throw new IllegalArgumentException(formattedString);
+    } else if (!cf.isEmpty() && cq.isEmpty()) {
+      scanner.fetchColumnFamily(
+          interpeter.interpretColumnFamily(new Text(cf.getBytes(Shell.CHARSET))));
+
+    } else if (!cf.isEmpty() && !cq.isEmpty()) {
+      scanner.fetchColumn(interpeter.interpretColumnFamily(new Text(cf.getBytes(Shell.CHARSET))),
+          interpeter.interpretColumnQualifier(new Text(cq.getBytes(Shell.CHARSET))));
 
     }
 
@@ -364,6 +387,9 @@ public class ScanCommand extends Command {
     optEndRowExclusive.setArgName("end-exclusive");
     scanOptRow = new Option("r", "row", true, "row to scan");
     scanOptColumns = new Option("c", "columns", true, "comma-separated columns");
+    scanOptCfOptions = new Option("cf", "column-family", true, "Column Family");
+    scanOptColQualifier = new Option("cq", "column-qualifier", true, "Column Qualifier");
+
     timestampOpt = new Option("st", "show-timestamps", false, "display timestamps");
     disablePaginationOpt = new Option("np", "no-pagination", false, "disable pagination of output");
     showFewOpt = new Option("f", "show-few", true, "show only a specified number of characters");
@@ -385,6 +411,8 @@ public class ScanCommand extends Command {
     scanOptRow.setArgName("row");
     scanOptColumns
         .setArgName("<columnfamily>[:<columnqualifier>]{,<columnfamily>[:<columnqualifier>]}");
+    scanOptCfOptions.setArgName("column-family");
+    scanOptColQualifier.setArgName("column-qualifier");
     showFewOpt.setRequired(false);
     showFewOpt.setArgName("int");
     formatterOpt.setArgName("className");
@@ -406,6 +434,8 @@ public class ScanCommand extends Command {
     o.addOption(optStartRowExclusive);
     o.addOption(optEndRowExclusive);
     o.addOption(scanOptColumns);
+    o.addOption(scanOptCfOptions);
+    o.addOption(scanOptColQualifier);
     o.addOption(timestampOpt);
     o.addOption(disablePaginationOpt);
     o.addOption(OptUtil.tableOpt("table to be scanned"));
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
@@ -307,8 +307,8 @@ public class ScanCommand extends Command {
     }
 
     if (cf.isEmpty() && !cq.isEmpty()) {
-          String formattedString = String.format("Option -%s is required when using -%s.",
-                  scanOptCf.getOpt(), scanOptCq.getOpt());
+      String formattedString = String.format("Option -%s is required when using -%s.",
+          scanOptCf.getOpt(), scanOptCq.getOpt());
       throw new IllegalArgumentException(formattedString);
     } else if (!cf.isEmpty() && cq.isEmpty()) {
       scanner.fetchColumnFamily(
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
@@ -270,9 +270,10 @@ public class ScanCommand extends Command {
 
   protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
       final ScanInterpreter formatter) throws UnsupportedEncodingException {
+
     if (cl.hasOption(scanOptColumns.getOpt())) {
       for (String a : cl.getOptionValue(scanOptColumns.getOpt()).split(",")) {
-        final String[] sa = a.split(":", 2);
+        final String[] sa = a.split("(?<!\\\\\\\\):", 2);
         if (sa.length == 1) {
           scanner.fetchColumnFamily(
               formatter.interpretColumnFamily(new Text(a.getBytes(Shell.CHARSET))));
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
@@ -386,7 +386,7 @@ public class ScanCommand extends Command {
     optEndRowExclusive.setArgName("end-exclusive");
     scanOptRow = new Option("r", "row", true, "row to scan");
     scanOptColumns = new Option("c", "columns", true,
-        "comma-separated columns.This" + "option is mutually exclusive with cf");
+        "comma-separated columns.This" + " option is mutually exclusive with cf and cq");
     scanOptCf = new Option("cf", "column-family", true, "column family to scan.");
     scanOptCq = new Option("cq", "column-qualifier", true, "column qualifier to scan");
 
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
@@ -388,7 +388,7 @@ public class ScanCommand extends Command {
     scanOptColumns = new Option("c", "columns", true,
         "comma-separated columns.This" + "option is mutually exclusive with cf");
     scanOptCf = new Option("cf", "column-family", true, "column family to scan.");
-    scanOptCq = new Option("cq", "column-qualifier", true, "column qualifier");
+    scanOptCq = new Option("cq", "column-qualifier", true, "column qualifier to scan");
 
     timestampOpt = new Option("st", "show-timestamps", false, "display timestamps");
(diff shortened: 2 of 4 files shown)


Manno15:
I was still unable to escape a ':' for the example mentioned in the original jira ticket (linked in the issue for this ticket). Did you do testing locally to confirm that functionality works?

milleruntime:
Complex regular expressions are difficult to read and maintain. Please try to avoid long regular expressions.

Also please include a test with these changes to verify it is working correctly.

harjitdotsingh:
> Complex regular expressions are difficult to read and maintain. Please try to avoid long regular expressions.
> 
> Also please include a test with these changes to verify it is working correctly.


Are there any existing test for the ShellCommands ?  I will see if I can simplify the RegularExpression. I don't want to add to many if and conditions to parse something, that's not a good practice in general. 

I actually tested the RE in a JavaPlayGround to make sure it was splitting the way we wanted. I used the expression test:\test:test1  it gave me two strings  1) test:\test and  2) test1.  I hope that is what is expected. Its possible that we may have to remove the \ from the first string so that the string is passed with : instead of it being escaped. DownStream code may not be handling the escaped string well.

milleruntime:
We have the `ShellServerIT` which tests a bunch of shell commands and then some of the commands have their own unit tests. You could add to the `ScanCommandTest` which tests the `ScanCommand` by itself.

Manno15:
>  I used the expression test:\test:test1 it gave me two strings 1) test:\test and 2) test1

The example in the jira ticket was scan -r foo -c person\:information:name, which has the '\' before the colon. Not sure if that changes anything but wanted to make sure that was clarified.

harjitdotsingh:
> > I used the expression test:\test:test1 it gave me two strings 1) test:\test and 2) test1
> 
> The example in the jira ticket was scan -r foo -c person:information:name, which has the '' before the colon. Not sure if that changes anything but wanted to make sure that was clarified.

No it doesn't that just a way to say its a character. Writing Tests now

harjitdotsingh:
@Manno15  @milleruntime  - I'm not sure what needs to be written for ScanCommandTest. There are not too many tests for those.
 Any pointers would be helpful.

Manno15:
You could look at `HistoryCommandTest` or `SetIterCommandTest` to get a better idea of how to implement a possible unit test for the command. If you think it needs to be tested on an actual cluster then adding a test to `ShellServerIT` would be the next step. For the actual test, I think you only need to create a table, insert the row into the table, and exec the scan command and if escape is done correctly then you should get the entire row back as output. 

Below is my shell history of when I tested this manually. I hope this helps. Let me know if you need anything else. 

```
createtable test
 insert foo person:information name bob
 scan
 scan -r foo -c person:information
 scan -r foo -c person\:information:name
 scan -r foo -c person\:information
```

harjitdotsingh:
@Manno15  I have the ShellServerIT tests done for the command, my questions was very specific for ScanCommandTest. I see lots of tests using the same validation like ScanCommandTest and MergeCommandTest, so I wasn't sure what was suppose to be done for that. Thanks for all your help.

Manno15:
> @Manno15 I have the ShellServerIT tests done for the command, my questions was very specific for ScanCommandTest. I see lots of tests using the same validation like ScanCommandTest and MergeCommandTest, so I wasn't sure what was suppose to be done for that. Thanks for all your help.

If you already created the test in ShellServerIT then you don't need to add anything to `ScanCommandTest`.

harjitdotsingh:
Pushed. Please let me know if you see any issues

Manno15 (changes requested):

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -1939,6 +1939,17 @@ public void scans() throws Exception {
...
   }
 
+  @Test
+  public void scanswithcolon() throws Exception {
+    ts.exec("createtable twithcolontest");
+    make10withcolon();
+    String result = ts.exec("scan -b row0 -c c\\:f:col0 -e row0");
+    assertEquals(1, result.split("\n").length);
+    result = ts.exec("scan -r row0 -c c\\:f:col0");
+    assertEquals(1, result.split("\n").length);
```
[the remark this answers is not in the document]
Manno15: You could also do something like the code below to confirm the output includes the expected value for each row. This is just a suggestion though. 

```java
ts.exec("scan -r row0 -c c\\:f:col0", true, "value");
```

harjitdotsingh:
Fixed the issues. Replaced RE with a function because we never get the escaped string in the method. Now I'm getting expected results.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -1939,6 +1939,17 @@ public void scans() throws Exception {
     ts.exec("deletetable -f t");
   }
 
+  @Test
+  public void scanswithcolon() throws Exception {
```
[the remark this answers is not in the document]
harjitdotsingh: Done

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -1939,6 +1939,19 @@ public void scans() throws Exception {
...
 
+  @Test
+  public void scansWithColon() throws Exception {
+    ts.exec("createtable twithcolontest");
+    make10WithColon();
+    ts.exec("scan -r row0 -c c\\:f:col0", true, "value");
+    String result = ts.exec("scan -b row1 -c c\\:f:col1 -e row1");
+    assertEquals(2, result.split("\n").length);
+    result = ts.exec("scan -r row0 -c c:f");
+    assertEquals(1, result.split("\n").length);
```
Manno15, line 1950: With the latest changes, it seems the use of an '\' character is unnecessary now. The reason your second check doesn't pass is because it does not contain the :col1 at the end.

Manno15: So `scan -b row1 -c c\:f:col1 -e row1` will return the row, `scan -b row1 -c c:f:col1 -e row1` will as well, but `scan -b row1 -c c\:f -e row1` and `scan -b row1 -c c:f -e row1` will not

harjitdotsingh: For Some reasons the '\' doesn't make it, the string is not escaped anymore and hence the RE was un-necessary. I made it more simpler. The last test is a negative test I deliberately left out to make sure that things are working. Hope that makes sense.

Manno15: > The last test is a negative test I deliberately left out to make sure that things are working. Hope that makes sense.

I understood that, I was using it to show the inconsistent behavior. My first comment was just poorly worded on my part.

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -285,6 +288,19 @@ protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
     }
   }
 
+  protected String[] extractColumnFamily(final String columnString) {
```
DomGarguilo, line 291: This might be able to be simplified by just using the substring method. Also it doesn't seem like this accounts for strings where there are more than 2 ':' characters (not sure if that will ever be the case), for example `a:c:d:e` would be sent to the `else` branch and possibly handled incorrectly (not sure if that will ever be the case).

harjitdotsingh: Are referring to that we just take a substring from the last index of : and that would do it?  If we have more than 2 that will be a problem, I can change the condition because anything  &gt; 2 should be split into two strings cf and qualifier.

EdColeman: The colFam and the colQual values are user defined so Accumulo really should be able to handle whatever they wish to do.  Unless there are specific characters that are prohibited, then they should be allowed.

One example could be an ipv6 address say colFam was IP and colQual was port - you may want to scan for a specific IP and specific port, just the IP, or even maybe IP that starts with and either with or without a port.

Using the IPv6 example from [wikipedia](https://en.wikipedia.org/wiki/IPv6_address) ```2001:0db8:85a3:0000:0000:8a2e:0370:7334``` If that was a colFam, then it would be desirable to scan for all or parts of that address plus (optionally) any colQual value that a user might want.

harjitdotsingh: makes sense, we should be able to fix it easily

Manno15: If that is the case, then having a separate opt for cf and cq could be easier. Take the entire string after the -cf opt instead of trying to parse everything accordingly.

harjitdotsingh: If we are taking input in this form 2001:0db8:85a3:0000:0000:8a2e:0370:7334 its hard to know what is part of cf and what is sq. I think if we would be seeing this then we should move to using options  to specify the column family and column qualifier using cf and cq.

DomGarguilo: > Are referring to that we just take a substring from the last index of : and that would do it?

Yea what I'm saying is, the substring method you have will correctly handle all cases so you could remove the else branch. @harjitdotsingh

harjitdotsingh: so if we go the cf and cq route do we get rid of -c or keep it ?

Manno15: I'd vote to get rid of it since cf and cq covers its functionality.

harjitdotsingh: Here are the comments from @ctubbsii  

`Alternatively, we can add -cf and -cq options to the scan command which is mutually exclusive with -c.`

Going by that  it would mean that we still want to keep -c but  IMHO its redundant to have  -c , -cf and -cq.  Would like to see what @ctubbsii  has to say about it? Do we want to keep -c with -cf and -cq. I think it will be confusing for people to have 2 options for a operation. Thoughts ?

EdColeman: It is likely that people have scripts that use -c. Removing that as an option will break those scripts.  People using -c currently are limited to not having : in the colFam, but they are coping.

Adding -cq and -cf provide additional capability when (the current -c) is insufficient for their needs.

harjitdotsingh:
We will keep it backwards compatible

shell/src/main/java/org/apache/accumulo/shell/commands/QuotedStringTokenizer.java
```diff
@@ -72,7 +72,7 @@ private void createTokens() throws BadArgumentException, UnsupportedEncodingExce
         inEscapeSequence = false;
         if (ch == 'x') {
           hexChars = "";
-        } else if (ch == ' ' || ch == '\'' || ch == '"' || ch == '\\') {
+        } else if (ch == ' ' || ch == '\'' || ch == '"' || ch == '\\' || ch == ':') {
```
Manno15, line 75: Not sure if this change is necessary for regards to this PR. If you think it still is then some of the comments should be updated to include the ':' being added.

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -59,7 +59,8 @@
 public class ScanCommand extends Command {
 
   private Option scanOptAuths, scanOptRow, scanOptColumns, disablePaginationOpt, showFewOpt,
-      formatterOpt, interpreterOpt, formatterInterpeterOpt, outputFileOpt;
+      formatterOpt, interpreterOpt, formatterInterpeterOpt, outputFileOpt, scanOptCfOptions,
+      scanOptColQualifier;
```
Manno15, line 63:
```suggestion
      formatterOpt, interpreterOpt, formatterInterpeterOpt, outputFileOpt, scanOptColFam,
      scanOptColQualifier;
```

Doesn't have to be this specific wording but the two added opts should have a similar naming convention.

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -270,6 +272,16 @@ protected ScanInterpreter getInterpreter(final CommandLine cl, final String tabl
 
   protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
       final ScanInterpreter formatter) throws UnsupportedEncodingException {
+
+    if ((cl.hasOption(scanOptCfOptions.getOpt()) || cl.hasOption(scanOptColQualifier.getOpt()))
+        && cl.hasOption(scanOptColumns.getOpt())) {
+
+      String formattedString =
+          String.format("Options - %s AND (- %s" + "OR - %s are mutually exclusive ",
```
Manno15, line 280: The spacing is a little off here. This is how it looks in the shell: `Options - c AND (- cfOR - cq are mutually exclusive`

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -285,6 +297,33 @@ protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
...
+    if (cl.hasOption(scanOptCfOptions.getOpt())) {
+      cf = cl.getOptionValue(scanOptCfOptions.getOpt());
+    }
+    if (cl.hasOption(scanOptColQualifier.getOpt())) {
+      cq = cl.getOptionValue(scanOptColQualifier.getOpt());
+    }
+
+    if (cf.isEmpty() && !cq.isEmpty()) {
+      String formattedString = String.format(
+          "Option - %s when used with (- %s" + "cannot be empty ", scanOptCfOptions.getOpt(),
```
Manno15, line 312: Like above, the spacing is also a little off here. This is how it looks in the shell: ` Option - cf when used with (- cqcannot be empty`

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -348,6 +387,9 @@ public Options getOptions() {
     optEndRowExclusive.setArgName("end-exclusive");
     scanOptRow = new Option("r", "row", true, "row to scan");
     scanOptColumns = new Option("c", "columns", true, "comma-separated columns");
+    scanOptCfOptions = new Option("cf", "column-family", true, "Column Family");
+    scanOptColQualifier = new Option("cq", "column-qualifier", true, "Column Qualifier");
```
Manno15, line 391:
```suggestion
    scanOptCfOptions = new Option("cf", "column-family", true, "column family");
    scanOptColQualifier = new Option("cq", "column-qualifier", true, "column qualifier");
```
To match the rest of the options. Could also add more detail such as `column family to scan`.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -1939,6 +1939,16 @@ public void scans() throws Exception {
     ts.exec("deletetable -f t");
   }
 
+  @Test
+  public void scansWithColon() throws Exception {
+    ts.exec("createtable twithcolontest");
+    make10WithColon();
+    ts.exec("scan -r row0 -cf c:f", true, "value");
+    String result = ts.exec("scan -b row1 -cf c:f  -cq col1 -e row1");
+    assertEquals(2, result.split("\n").length);
```
Manno15, line 1948: Could add a test case for if a user adds the `-cq` option without adding a `-cf` and one where they also add the `-c` option.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -1939,6 +1939,16 @@ public void scans() throws Exception {
     ts.exec("deletetable -f t");
   }
 
+  @Test
+  public void scansWithColon() throws Exception {
+    ts.exec("createtable twithcolontest");
+    make10WithColon();
```
Manno15, line 1945:
```suggestion
    ts.exec("insert row c:f cq value")
```
This test really only needs one row in the table.

Manno15 (commented):
Things are working well. I like the changes, just a few minor things to fix up.

shell/src/main/java/org/apache/accumulo/shell/commands/QuotedStringTokenizer.java
harjitdotsingh: We don't need this. We can revert this back. Let me do that

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
harjitdotsingh: sure

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -348,6 +387,9 @@ public Options getOptions() {
     optEndRowExclusive.setArgName("end-exclusive");
     scanOptRow = new Option("r", "row", true, "row to scan");
     scanOptColumns = new Option("c", "columns", true, "comma-separated columns");
```
Manno15, line 389: Could mention that -c is mutually exclusive here so users will avoid the thrown error above. One idea I had is if both options are passed in, we could just log a warning and then choose to ignore one of the arguments (either -c or -cf/cq). Thoughts on that?

harjitdotsingh:
We can Add it to the description. I think make it mutually exclusive make
it more easier and the code manageable . We will have to add checks now to
both the places. I like the suggestion though

harjitdotsingh:
All changes made. Tests look good. Hopefully this should be it. Please let me know if you find anything.

test/src/main/java/org/apache/accumulo/test/ShellServerIT.java
```diff
@@ -1939,6 +1939,22 @@ public void scans() throws Exception {
...
+    ts.exec("insert row c:f cq value");
+    ts.exec("scan -r row -cf c:f", true, "value");
+    String result = ts.exec("scan -b row -cf c:f  -cq cq -e row");
+    assertEquals(2, result.split("\n").length);
+    result = ts.exec("scan -b row -c cf -cf c:f  -cq cq -e row", false);
+    assertTrue(result.contains("mutually exclusive"));
+
+    result = ts.exec("scan -b row -cq col1 -e row", false);
+    assertTrue(result.contains("cannot be empty"));
+
```
Manno15, line 1954:
```suggestion
    ts.exec("scan -b row -cf c:f  -cq cq -e row", true, "value");
    ts.exec("scan -b row -c cf -cf c:f  -cq cq -e row", false, "mutually exclusive");
    ts.exec("scan -b row -cq col1 -e row", false, "cannot be empty");
```
A slight simplification so we don't have to create the results object to compare against.

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -347,7 +385,11 @@ public Options getOptions() {
         "make end row exclusive (by default it's inclusive)");
     optEndRowExclusive.setArgName("end-exclusive");
     scanOptRow = new Option("r", "row", true, "row to scan");
-    scanOptColumns = new Option("c", "columns", true, "comma-separated columns");
+    scanOptColumns = new Option("c", "columns", true,
+        "comma-separated columns.This" + "option is mutually exclusive with cf");
```
Manno15, line 389:
```suggestion
        "comma-separated columns.This" + " option is mutually exclusive with cf and cq");
```

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -347,7 +385,11 @@ public Options getOptions() {
...
+    scanOptCf = new Option("cf", "column-family", true, "column family to scan.");
+    scanOptCq = new Option("cq", "column-qualifier", true, "column qualifier");
```
Manno15, line 391:
```suggestion
    scanOptCq = new Option("cq", "column-qualifier", true, "column qualifier to scan");
```

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -285,6 +296,33 @@ protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
...
+      cf = cl.getOptionValue(scanOptCf.getOpt());
+    }
+    if (cl.hasOption(scanOptCq.getOpt())) {
+      cq = cl.getOptionValue(scanOptCq.getOpt());
+    }
+
+    if (cf.isEmpty() && !cq.isEmpty()) {
+      String formattedString =
+          String.format("Option - %s when used with (%s" + " cannot be empty )", scanOptCf.getOpt(),
+              scanOptCq.getOpt(), scanOptCq.getOpt());
```
Manno15, line 312:
```suggestion
          String.format("Option -%s is required when using -%s.", scanOptCf.getOpt(),
              scanOptCq.getOpt());
```
Some suggestions are wording to make it easier to understand. Also, extra parameter was removed and fixed spacing.

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -270,6 +271,16 @@ protected ScanInterpreter getInterpreter(final CommandLine cl, final String tabl
 
   protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
       final ScanInterpreter formatter) throws UnsupportedEncodingException {
+
+    if ((cl.hasOption(scanOptCf.getOpt()) || cl.hasOption(scanOptCq.getOpt()))
+        && cl.hasOption(scanOptColumns.getOpt())) {
+
+      String formattedString =
+          String.format("Options - %s AND (- %s" + " OR - %s are mutually exclusive )",
+              scanOptColumns.getOpt(), scanOptCf.getOpt(), scanOptCq.getOpt());
```
Manno15, line 280:
```suggestion
          String.format("Option -%s is mutually exclusive with options -%s and -%s.",
              scanOptColumns.getOpt(), scanOptCf.getOpt(), scanOptCq.getOpt());
```
Some suggestions on wording to make it easier to understand and fixed spacing.

Manno15 (commented):
This is getting close. Just a few more minor changes/suggestions.

harjitdotsingh:
All set. Just committed the changes

Manno15:
Appears some of my suggestions weren't properly formatted, Hard to know when doing it through Github UI. My apologies.

shell/src/main/java/org/apache/accumulo/shell/commands/ScanCommand.java
```diff
@@ -285,6 +296,33 @@ protected void fetchColumns(final CommandLine cl, final ScannerBase scanner,
...
+      cf = cl.getOptionValue(scanOptCf.getOpt());
+    }
+    if (cl.hasOption(scanOptCq.getOpt())) {
+      cq = cl.getOptionValue(scanOptCq.getOpt());
+    }
+
+    if (cf.isEmpty() && !cq.isEmpty()) {
+      String formattedString =
+          String.format("Option -%s is required when using -%s.", scanOptCf.getOpt(),
+              scanOptCq.getOpt());
```
Manno15, line 312:
```suggestion
          String formattedString = String.format("Option -%s is required when using -%s.",
          scanOptCf.getOpt(), scanOptCq.getOpt());
```

This should fix the formatting but feel free to format on your end

harjitdotsingh:
No worries let me run it again and see

harjitdotsingh:
Pushed. Lets see if that takes care of the formatting issues.

Manno15 (approved):
Thanks for the changes @harjitdotsingh!

harjitdotsingh:
Thanks… will try to work on other one

PR #2258, opened 2021-09-01
Prevents compaction when tablet does not know of files

keith-turner:
I noticed when looking into #2239 that when a compaction job starts and reserves files that there is no check to ensure the tablet still has those files.  I think this race condition is unlikely to occur, but not impossible.  This PR add a check for this.  This check prevents unnecessary work from occurring.  The code is correct without the check in that it would currently fail after the compaction runs if this situation were to occur.

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
@@ -1110,6 +1110,18 @@ public class CompactableImpl implements Compactable {
       cInfo.localCompactionCfg = this.compactionConfig;
     }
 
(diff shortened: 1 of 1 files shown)


Manno15 (approved):

EdColeman:
Asking for clarification - what happens if the check fails?  Is it rescheduled or is that left up to the caller that scheduled the compaction?

keith-turner:
> Asking for clarification - what happens if the check fails? Is it rescheduled or is that left up to the caller that scheduled the compaction?

@EdColeman tablets are periodically examined and a compaction planner is asked to create zero or more compaction jobs (an immutable object the describes compaction work to do).  These jobs are queued to run later.  When the job actually goes to run, maybe things have changed (like this PR is checking for) in which case the job is ignored.  Later when the tablet is examined again, new jobs will be created that incorporate the changes.

PR #2260, opened 2021-09-02
Reword some comments to clarify

milleruntime:
* Also rename some methods in CompactableImpl

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactableImpl.java
@@ -183,8 +183,7 @@ public class CompactableImpl implements Compactable {
     // Tracks if when a set of files was selected, if at that time the set was all of the tablets
     // files. Because a set of selected files can be compacted over one or more compactions, its
     // important to track this in order to know if the last compaction is a full compaction and
-    // should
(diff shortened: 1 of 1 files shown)


Manno15 (approved):

keith-turner (approved):

milleruntime:
@keith-turner One comment I was hoping we could expand on was "So the selected set must be initialized." I am not sure if that means in the current method or before calling the current method.

DomGarguilo (approved):
LGTM. See a PR I made to this branch [here](https://github.com/milleruntime/accumulo/pull/16) for a few other very small comment related fixes.

keith-turner:
> @keith-turner One comment I was hoping we could expand on was "So the selected set must be initialized." I am not sure if that means in the current method or before calling the current method.

Seems like it initializes some instance var in the method and also uses the returned value to initialize some more instance var later in the constructor.  I don't think anything is initialized before the method call.

milleruntime:
> > @keith-turner One comment I was hoping we could expand on was "So the selected set must be initialized." I am not sure if that means in the current method or before calling the current method.
> 
> Seems like it initializes some instance var in the method and also uses the returned value to initialize some more instance var later in the constructor. I don't think anything is initialized before the method call.

OK I think I am going to just drop that sentence since it doesn't really add anything and is just confusing me.

PR #2262, opened 2021-09-03
Add check for deleted Table to Tablet compactAll

milleruntime:
* Also drop unnecessary boolean and add comment


Manno15 (approved):
Changes seem reasonable.

PR #2263, opened 2021-09-08
Split compact method into two more specific methods

milleruntime:
* Split compact method in CompactionService into to 2 methods:
getCompactionPlan() and submitCompactionJob()
* Rename other compact() in CompactionManager and CompactionService to submitCompaction()
* Add javadoc comments to various methods
* Add final to various private members
* Drop redundant enum check and boolean assignment

Edit: ~~Add getCandidates() method to CompactionPlan~~

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/spi/compaction/CompactionPlan.java
+++ b/core/src/main/java/org/apache/accumulo/core/spi/compaction/CompactionPlan.java
@@ -19,6 +19,7 @@
 package org.apache.accumulo.core.spi.compaction;
 
 import java.util.Collection;
+import java.util.Set;
 
 import org.apache.accumulo.core.client.admin.compaction.CompactableFile;
 import org.apache.accumulo.core.spi.compaction.CompactionPlanner.PlanningParameters;
@@ -55,5 +56,13 @@ public interface CompactionPlan {
     CompactionPlan build();
   }
 
+  /**
+   * Return the set of compaction candidates.
+   */
+  Set<CompactableFile> getCandidates();
+
+  /**
+   * Return the set of jobs this plan will submit for compaction.
+   */
   Collection<CompactionJob> getJobs();
 }
--- a/core/src/main/java/org/apache/accumulo/core/spi/compaction/CompactionPlan.java
+++ b/core/src/main/java/org/apache/accumulo/core/spi/compaction/CompactionPlan.java
@@ -19,7 +19,6 @@
 package org.apache.accumulo.core.spi.compaction;
 
 import java.util.Collection;
-import java.util.Set;
 
 import org.apache.accumulo.core.client.admin.compaction.CompactableFile;
 import org.apache.accumulo.core.spi.compaction.CompactionPlanner.PlanningParameters;
@@ -55,11 +54,6 @@ public interface CompactionPlan {
 
     CompactionPlan build();
   }
-
-  /**
-   * Return the set of compaction candidates.
-   */
-  Set<CompactableFile> getCandidates();
 
   /**
    * Return the set of jobs this plan will submit for compaction.
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java
@@ -27,6 +27,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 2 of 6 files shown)


server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java
```diff
@@ -223,15 +224,22 @@ private boolean reconcile(Set<CompactionJob> jobs, Collection<SubmittedJob> subm
...
     Objects.requireNonNull(compactable);
 
+    // add tablet to planning queue and use planningExecutor to get the plan
     if (queuedForPlanning.get(kind).putIfAbsent(compactable.getExtent(), compactable) == null) {
       try {
         planningExecutor.execute(() -> {
           try {
-            planCompaction(kind, compactable, completionCallback);
+            Optional<CompactionPlan> plan = getCompactionPlan(kind, compactable);
+            plan.ifPresent(cp -> submitCompactionJob(cp, compactable, completionCallback));
```
keith-turner, line 242: As mentioned in other comments, should probably not add candidates to plan and the reason it was added does not actually log all of the info.  Think the following changes may get the needed information where it needs to go w/o modifying the plan interface.

```suggestion
            var files = compactable.getFiles(...);
            Optional<CompactionPlan> plan = getCompactionPlan(kind, files, compactable.getExtent());
            plan.ifPresent(cp -> submitCompactionJob(cp, files, compactable.getExtent(), completionCallback));
```

server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java
```diff
@@ -363,11 +374,11 @@ private void planCompaction(CompactionKind kind, Compactable compactable,
 
       if (!jobs.isEmpty()) {
         log.trace("Submitted compaction plan {} id:{} files:{} plan:{}", compactable.getExtent(),
-            myId, files, plan);
+            myId, plan.getCandidates(), plan);
```
keith-turner, line 377: With changes in other comment could revert to the following.  Also this changes is logging less information than it used to. Used to log the Compactable.Files obj now it logging a subset of that info.

```suggestion
            myId, files, plan);
```

core/src/main/java/org/apache/accumulo/core/spi/compaction/CompactionPlan.java
```diff
@@ -55,5 +56,13 @@ Builder addJob(short priority, CompactionExecutorId executor,
     CompactionPlan build();
   }
 
+  /**
+   * Return the set of compaction candidates.
+   */
+  Set<CompactableFile> getCandidates();
```
keith-turner, line 62: Thinking this should not be added to SPI.  It has nothing to do with the plan for a compaction, but is a subset of the information that went into creating a plan.   Also I suggested other changes that would not require this change.

milleruntime: I was thinking since the `CompactionPlan` is the output of the `CompactionPlanner`, it would have the candidates in it. But now I see that `CompactionJob` has files as well. I will drop this from the plan. I am thinking now, it would be nice to be able to differentiate between the files as to when `CompactableFile` actually gets selected for compaction or not. Maybe we could do something similar to what I did with `TabletFile` and `StoredTabletFile`. 

Also, trying to write javadoc for `CompactableFile` made me wonder if we should make it a self contained collection since you need at least 2 files to compact them into 1. There shouldn't ever be a single CompactableFile by itself right?

server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java
milleruntime: Thanks. I like this better. I thought getting the files was part of getting the plan but it looks like the plan is more about the jobs.

core/src/main/java/org/apache/accumulo/core/spi/compaction/CompactionPlan.java
milleruntime: > There shouldn't ever be a single CompactableFile by itself right?

@EdColeman reminded me of the use case where a user configures the `AgeOffFilter` and runs a user compaction to drop old data. Do you know where this would occur in CompactableImpl? I can't find it in the new code.

server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java
milleruntime: I reverted the changes to CompactionPlan but here I changed the method to use the `Compactable.Files` object as opposed to the Optional. I think this cleans it up a bit and should print better info than printing the Optional object.

milleruntime: I also noticed printing the `Optional.empty` after we check if its empty just clutters the log message, so I dropped it:
<pre>
TRACE: Compactable returned no files 4;row_0003250000;row_0003000000 USER Optional.empty
TRACE: Compactable returned no files 4;row_0003250000;row_0003000000 CHOP Optional.empty
TRACE: Compactable returned no files 4;row_0003500000;row_0003250000 SYSTEM Optional.empty
TRACE: Compactable returned no files 4;row_0003500000;row_0003250000 SELECTOR Optional.empty
</pre>

keith-turner (approved):

PR #2265, opened 2021-09-09
Fixes #2199 allow system compaction to preempt queued user compactions

### LINKED ISSUE #2199: Switch compaction file selection from early binding to late binding.

keith-turner:


dlmarion:
With this change, is it the case that a user compaction could be delayed indefinitely if ingest is occurring into that tablet at such a rate that system compactions are always being run? My concern here would be the case where an admin kicks off compactions to perform some necessary work that needs to be run in a timely fashion, but it's not run in a timely fashion. I'm wondering if the user compaction should preempt the system compaction.

keith-turner:
> With this change, is it the case that a user compaction could be delayed indefinitely if ingest is occurring into that tablet at such a rate that system compactions are always being run?

I think this is possible. I suspect that as the configurable timeout is increased, that the probability of this happening approaches zero (but never gets to zero).  The configurable expiration time reserves files for a user compactions while its queued.  I set it to 1 minute (not sure what a good default is, still thinking about that).  So for 1 minute after a user compaction is queued, a system compaction can not take any of the files reserved for it.  After a user compaction starts running a system compaction can never take any of its files. 


> I'm wondering if the user compaction should preempt the system compaction.

One of the motivating cases for this change is a tablet with lots of files that is queued for really long periods for a user compaction.  For example if a tablet had 1000 files and a user compaction was initiated, then the tablet would reserve those files and queue a user compaction.  If the user compaction sat on the queue for 9 hours then during that time system compactions would not be able to work down the number of files (if system compactions had the capacity to do so).  I feel this potential problem is worth mitigating.  The question is, is the cure (and its potential to cause starvation) worse than the problem.    I feel that its not, but I don't have any numbers to back that up.  I will poke at the likelihood of starvation a bit more and see if I can roughly estimate it.

keith-turner:
Expanding on the previous comment.  If system compactions for the tablet with 1000 files also sat on a queue and never actually ran, then it would never unreserve the user compaction files. Its only when a system compaction starts to run (after the expiration period) that it would break the reservation, the break does not happen when its queued.  This can be seen in the change to the PR that reserves files before a compaction actually starts running.

keith-turner:
I was running a test where I did a full table compaction while continuous ingest was running.  Below are some log messages for a single tablet that selected some files for the user compaction and then had a system compaction cancel the selection.  The system compaction ran and then files were reselected for the user compaction which then ran.

```
2021-09-09T21:32:28,780 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 NEW 5 []
2021-09-09T21:32:28,780 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 SELECTING
2021-09-09T21:32:28,780 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 SELECTED true [C0031umi.rf, C0048yry.rf, C0042902.rf, F0049k8v.rf, A002t9q4.rf, F0049hzr.rf, C00480x2.rf, C003qq36.rf, C003j8jf.rf]
2021-09-09T21:34:07,020 [tablet.files] DEBUG: Flushed 2;647ce03;6479ad9 created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/F0049tcw.rf from [memory]
2021-09-09T21:35:49,321 [tablet.files] DEBUG: Flushed 2;647ce03;6479ad9 created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/F004a15i.rf from [memory]
2021-09-09T21:35:49,321 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 NOT_ACTIVE because selection expired.
2021-09-09T21:35:49,321 [tablet.files] DEBUG: Compacting 2;647ce03;6479ad9 on i.cs1.small for SYSTEM from [F004a15i.rf, F0049hzr.rf, F0049tcw.rf, F0049k8v.rf] size 1 MB config []
2021-09-09T21:35:49,796 [tablet.files] DEBUG: Compacted 2;647ce03;6479ad9 for SYSTEM created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/C004a15m.rf from [F004a15i.rf, F0049hzr.rf, F0049tcw.rf, F0049k8v.rf]
2021-09-09T21:36:00,227 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 NEW 5 []
2021-09-09T21:36:00,227 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 SELECTING
2021-09-09T21:36:00,227 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 SELECTED true [C0031umi.rf, C0048yry.rf, C0042902.rf, C004a15m.rf, A002t9q4.rf, C00480x2.rf, C003qq36.rf, C003j8jf.rf]
2021-09-09T21:37:58,599 [tablet.files] DEBUG: Flushed 2;647ce03;6479ad9 created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/F004aghw.rf from [memory]
2021-09-09T21:39:56,398 [tablet.files] DEBUG: Flushed 2;647ce03;6479ad9 created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/F004aped.rf from [memory]
2021-09-09T21:42:01,817 [tablet.files] DEBUG: Flushed 2;647ce03;6479ad9 created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/F004azef.rf from [memory]
2021-09-09T21:43:15,659 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 RESERVED
2021-09-09T21:43:15,660 [tablet.files] DEBUG: Compacting 2;647ce03;6479ad9 on i.cs2.large for USER from [C0048yry.rf, C0042902.rf, A002t9q4.rf, C003j8jf.rf, C003qq36.rf, C00480x2.rf, C0031umi.rf, C004a15m.rf] size 540 MB config []
2021-09-09T21:44:52,322 [tablet.files] DEBUG: Compacted 2;647ce03;6479ad9 for USER created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/A004b3r0.rf from [C0048yry.rf, C0042902.rf, A002t9q4.rf, C003j8jf.rf, C003qq36.rf, C00480x2.rf, C0031umi.rf, C004a15m.rf]
2021-09-09T21:44:52,322 [tablet.CompactableImpl] TRACE: Selected compaction status changed 2;647ce03;6479ad9 NOT_ACTIVE
2021-09-09T21:44:52,423 [tablet.files] DEBUG: Flushed 2;647ce03;6479ad9 created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/F004b8xz.rf from [memory]
2021-09-09T21:44:52,424 [tablet.files] DEBUG: Compacting 2;647ce03;6479ad9 on i.cs1.small for SYSTEM from [F004aghw.rf, F004aped.rf, F004b8xz.rf, F004azef.rf] size 1 MB config []
2021-09-09T21:44:53,037 [tablet.files] DEBUG: Compacted 2;647ce03;6479ad9 for SYSTEM created hdfs://muchoshacluster/accumulo/tables/2/t-003nr5v/C004b8y5.rf from [F004aghw.rf, F004aped.rf, F004b8xz.rf, F004azef.rf]
```

keith-turner:
@dlmarion I thought about this some more and I do not think indefinite starvation can occur.  Before this change indefinite starvation could occur if a compactions was queued and higher priority compactions kept arriving while it was queued.  So I am thinking lets ignore that case and ask the question can this change cause starvation when only lower priority user compactions are arriving?  For this case I do not think indefinite starvation can occur.

There are two behaviors that avoid starvation
 * User compactions have an exclusive amount of time on the queue in which a system compaction can not remove them from the queue.
 * When a user compaction is removed from the queue by a system compaction, its guaranteed that the user compaction will be  requeued after that system compaction runs.  When its requeued it will have the exclusive time on the queue again.

So if a user compaction is queued and its next in line waiting for running compactions to finish, then there is a non-zero probability that one of those compactions will finish during its exclusive time.  I think this probability is a function of the average run time and number fo threads.  So if there are 2 threads, the average user compaction time is 5 mins, and the exclusive time is 1 mins then the probability that its gets to run in its exclusive time might be `1-(1-1/5)^2=.36`.  So in this case there is a 36% change its get to run during its first exclusive time.  If it does not get to run during its first exclusive time, then by the second condition above its guaranteed another exclusive time in which it will have another 36% chance to run.  So it will eventually run.

Alternating between queued times and non-queued times does prolong the time it takes. I think it may do this in a linear manner.  If the exclusive time is 1 min and the average system compaction takes 2 min (during which its not queued for user compaction), then user compactions may take on average 3 times as long to actually start running because they spend 2/3 of the time with a 0% chance of running the compaction and 1/3 of the time with a X% chance of running the compaction where X>0.

One possible wrinkle in the above argument is that system compactions may lower the number of files which lowers the priority of subsequently queued user compactions for the tablet.  However system compactions based on the compaction ratio do not continually lower the number of files, but converge to an average number of files based on the compaction ratio, tablet size, and size of incoming files. So there will not be a situation where the priority continually decreases as system compactions occur, it will oscillate within a range as new files arrive and system compactions occur.

core/src/main/java/org/apache/accumulo/core/conf/Property.java
```diff
@@ -887,6 +887,14 @@
       "2.1.0"),
   TABLE_COMPACTION_DISPATCHER_OPTS("table.compaction.dispatcher.opts.", null, PropertyType.PREFIX,
       "Options for the table compaction dispatcher", "2.1.0"),
+  TABLE_COMPACTION_SELECTION_EXPIRATION("table.compaction.selection.expiration.ms", "1m",
+      PropertyType.TIMEDURATION,
+      "User compactions select files and are then queued for compaction, preventing these files "
+          + "from being used in system compactions.  This timeout allows system compactions to cancel "
+          + "the hold queued user compactions have on files, when its queued for more than the "
+          + "specified time.  If a system compaction cancels a hold and runs, then the user compaction"
+          + " can reselect and hold files after the system compaction runs.   ",
```
dlmarion, line 896: small nit - there is some space at the end of the description.

dlmarion (approved):

PR #2269, opened 2021-09-10
Improve BatchWriter throughput

### LINKED ISSUE #1120: Improve batch writer throughput

dlmarion:
Modified the TabletServerBatchWriter to use concurrent data structures such that
mutations could be added and binned simultaneously which allowed me to remove the
synchronized modifier from several methods. Specifically, we:

  Removed startProcessing() which added queued mutations to the MutationWriter
  Removed BatchWriterLatencyTimer thread, which called startProcessing had not been called in the latency interval
  Removed calls to startProcessing() from flush and close
  Removed call to startProcessing when used memory is half of max memory when adding a mutation to the BatchWriter
  Modified TSBW.MutationSet to use a ConcurrentHashMap instead of a HashMap
  Modified some variables to use Atomic datatypes to remove the synchronized modified from addMutation
  Removed TSBW.FailedMutations object, which requeued failed mutations on a 500ms interval, instead requeues are done immediately
  Moved binning task to BatchWriter constructor and modified such that is is always running
  Added SendTasks to the sendThreadPool in the order of servers with the most work

Closes #1120

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
@@ -37,10 +37,9 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Queue;
 import java.util.Set;
-import java.util.TreeMap;
 import java.util.TreeSet;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
@@ -76,6 +75,7 @@ import org.apache.accumulo.core.tabletserver.thrift.TabletClientService;
 import org.apache.accumulo.core.trace.TraceUtil;
 import org.apache.accumulo.core.trace.thrift.TInfo;
 import org.apache.accumulo.core.util.HostAndPort;
+import org.apache.accumulo.core.util.Pair;
 import org.apache.accumulo.core.util.threads.ThreadPools;
 import org.apache.htrace.Trace;
 import org.apache.htrace.TraceScope;
@@ -110,6 +110,22 @@ import com.google.common.base.Joiner;
  */
 public class TabletServerBatchWriter implements AutoCloseable {
 
+  private static class TableIdAndMutation extends Pair<TableId,Mutation> {
+
+    public TableIdAndMutation(TableId f, Mutation s) {
+      super(f, s);
+    }
+
+    public TableId getTableId() {
+      return super.getFirst();
+    }
+
+    public Mutation getMutation() {
+      return super.getSecond();
+    }
+
+  }
+
   private static final Logger log = LoggerFactory.getLogger(TabletServerBatchWriter.class);
 
   // basic configuration
@@ -122,7 +138,7 @@ public class TabletServerBatchWriter implements AutoCloseable {
   // state
   private boolean flushing;
   private boolean closed;
-  private MutationSet incomingQueue;
+  private LinkedBlockingQueue<TableIdAndMutation> incomingQueue;
 
   // background writer
   private final MutationWriter writer;
@@ -139,7 +155,7 @@ public class TabletServerBatchWriter implements AutoCloseable {
   private final AtomicLong totalBinned = new AtomicLong(0);
   private final AtomicLong totalBinTime = new AtomicLong(0);
   private final AtomicLong totalSendTime = new AtomicLong(0);
-  private final AtomicLong startTime = new AtomicLong(0);
+  private long startTime = 0;
   private long initialGCTimes;
   private long initialCompileTimes;
   private double initialSystemLoad;
@@ -206,7 +222,7 @@ public class TabletServerBatchWriter implements AutoCloseable {
     this.maxLatency = config.getMaxLatency(TimeUnit.MILLISECONDS) <= 0 ? Long.MAX_VALUE
         : config.getMaxLatency(TimeUnit.MILLISECONDS);
     this.timeout = config.getTimeout(TimeUnit.MILLISECONDS);
-    this.incomingQueue = new MutationSet();
+    this.incomingQueue = new LinkedBlockingQueue<>();
     this.durability = config.getDurability();
     this.writer = new MutationWriter(config.getMaxWriteThreads());
   }
@@ -218,7 +234,8 @@ public class TabletServerBatchWriter implements AutoCloseable {
     }
   }
 
-  public void addMutation(final TableId table, final Mutation m) throws MutationsRejectedException {
+  public synchronized void addMutation(final TableId table, final Mutation m)
+      throws MutationsRejectedException {
 
     long start = System.currentTimeMillis();
 
@@ -246,7 +263,8 @@ public class TabletServerBatchWriter implements AutoCloseable {
 
     checkForFailures();
 
-    if (startTime.compareAndSet(0, System.currentTimeMillis())) {
+    if (startTime == 0) {
+      startTime = System.currentTimeMillis();
 
       List<GarbageCollectorMXBean> gcmBeans = ManagementFactory.getGarbageCollectorMXBeans();
       for (GarbageCollectorMXBean garbageCollectorMXBean : gcmBeans) {
@@ -269,7 +287,7 @@ public class TabletServerBatchWriter implements AutoCloseable {
     Mutation copy = new Mutation(m);
 
     totalMemUsed.getAndAdd(copy.estimatedMemoryUsed());
-    incomingQueue.add(table, copy);
+    incomingQueue.add(new TableIdAndMutation(table, copy));
     totalAdded.getAndIncrement();
 
   }
@@ -349,7 +367,7 @@ public class TabletServerBatchWriter implements AutoCloseable {
       }
 
       double averageRate = totalSent.get() / (totalSendTime.get() / 1000.0);
-      double overallRate = totalAdded.get() / ((finishTime - startTime.get()) / 1000.0);
+      double overallRate = totalAdded.get() / ((finishTime - startTime) / 1000.0);
 
       double finalSystemLoad = ManagementFactory.getOperatingSystemMXBean().getSystemLoadAverage();
 
@@ -359,8 +377,8 @@ public class TabletServerBatchWriter implements AutoCloseable {
       log.trace(String.format("Sent                 : %,10d mutations", totalSent.get()));
       log.trace(String.format("Resent percentage   : %10.2f%s",
           (totalSent.get() - totalAdded.get()) / (double) totalAdded.get() * 100.0, "%"));
-      log.trace(String.format("Overall time         : %,10.2f secs",
-          (finishTime - startTime.get()) / 1000.0));
+      log.trace(
+          String.format("Overall time         : %,10.2f secs", (finishTime - startTime) / 1000.0));
       log.trace(String.format("Overall send rate    : %,10.2f mutations/sec", overallRate));
       log.trace(
           String.format("Send efficiency      : %10.2f%s", overallRate / averageRate * 100.0, "%"));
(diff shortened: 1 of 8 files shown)


dlmarion:
@keith-turner  - is this what you had in mind for #1120? I still need to do some testing and figure out a way to compare the old and new implementation.

dlmarion:
FWIW - the failing check works for me locally....

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -201,47 +202,25 @@ public TabletServerBatchWriter(ClientContext context, BatchWriterConfig config)
...
   }
 
-  private synchronized void startProcessing() {
-    if (mutations.getMemoryUsed() == 0)
-      return;
-    lastProcessingStartTime = System.currentTimeMillis();
-    writer.queueMutations(mutations);
-    mutations = new MutationSet();
-  }
+  public void addMutation(final TableId table, final Mutation m) throws MutationsRejectedException {
```
keith-turner, line 221: I am thinking just sticking with synchronized for the method, and avoiding instance individual vars that support concurrency , make its much easier reason about operations across multiple instance vars.  For example in this method there is a check `if(closed){...}`, however since the method is not syncronized it could be closed at any point in the method.  Is the code still correct for the closed check?  Not sure, but its harder to reason about.

If the synchronized blocks of code only wrap code that manipulates in mem structs, then they should all be extremely fast relative to actually writing mutations to a tserver.  So w/ that in mind I am thinking the synchronized blocks are easy to reason about and will not be a bottleneck relative to the overall operation of the batch writer.  I think we could still avoid the wait till mem is half full behavior while still using sync block.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -753,15 +721,16 @@ private void addMutations(MutationSet mutationsToSend) {
         log.trace(String.format("Started sending %,d mutations to %,d tablet servers", count,
             binnedMutations.keySet().size()));
 
-      // randomize order of servers
-      ArrayList<String> servers = new ArrayList<>(binnedMutations.keySet());
-      Collections.shuffle(servers);
+      // order servers by decreasing number of mutations
+      TreeMap<Integer,String> servers = new TreeMap<>();
+      binnedMutations.forEach((server, tsm) -> servers.put(tsm.getMutations().size(), server));
```
keith-turner, line 726: The more I think about this I am not sure if its a good ideal.  Its probably a good thing for a single batch writer writing to many tservers.  However for many batch writers it seems like it could lead to a degenerate case where all the batch writer write to a few tservers, where each batch writer randomly ordering tservers would avoid this.

keith-turner (commented):
@dlmarion I made a few comments, but I am still looking over this so the comments were made w/o me having a complete picture of the changes.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -118,7 +122,7 @@
   // state
   private boolean flushing;
   private boolean closed;
-  private MutationSet mutations;
+  private MutationSet incomingQueue;
```
keith-turner, line 125: Seems like a BlockingQueue<Mutation> may be better for this, but not sure still looking over the code.  Why was the type MutationsSet chosen?

keith-turner: Looking at this a bit more, one thing MutationsSet does is keep track of the table id that goes with a mutation.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -250,15 +229,24 @@ public synchronized void addMutation(TableId table, Mutation m)
 
     checkForFailures();
```
keith-turner, line 230: Now that checkForFailures is not called in a synch block may throw ConcurrentModificationExceptions as its impl iterates over collections that other threads may modify.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -985,54 +955,43 @@ private MutationSet sendMutationsToTabletServer(String location,
...
+
+    private static final long serialVersionUID = 1L;
 
-    private final HashMap<TableId,List<Mutation>> mutations;
     private long memoryUsed = 0;
 
-    MutationSet() {
-      mutations = new HashMap<>();
+    public void add(TableId table, Mutation mutation) {
+      memoryUsed += mutation.estimatedMemoryUsed();
```
keith-turner, line 975: This add has a race condition and is not correct w/o sync.  This code may have relied on being called in a sync block when the obj was used by multiple threads (some mutation set objs are only used by a single thread within a method).

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -639,6 +590,39 @@ public MutationWriter(int numSendThreads) {
...
+            if (num > 0) {
+              try {
+                log.trace("{} - binning {} mutations", Thread.currentThread().getName(), num);
+                // Remove the mutations from the incoming queue
+                final MutationSet copy = new MutationSet();
+                incomingQueue.forEach((table, mutations) -> {
+                  mutations.forEach(mutation -> {
+                    copy.add(table, mutation);
+                    mutations.remove(mutation);
+                  });
```
keith-turner, line 607: Calling `remove()` on a collection while iterating over it may lead to unpredictable results.  Could use poll() since this is a Queue or user an iterator and call remove on it.

```suggestion
                  Mutation m = null;
                  while((m = mutations.poll()) != null) {
                	      copy.add(table, m);
                  }
```

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -985,54 +955,43 @@ private MutationSet sendMutationsToTabletServer(String location,
...
+      this.closed = true;
+      binningThreadPool.shutdownNow();
+      sendThreadPool.shutdownNow();
+    }
   }
 
   // END code for sending mutations to tablet servers using background threads
 
-  private static class MutationSet {
+  private static class MutationSet extends ConcurrentHashMap<TableId,Queue<Mutation>> {
```
keith-turner, line 968: These concurrent data structs are great at increasing the throughput of highly concurrent applications, however they are slower for single threaded uses cases because of heavy use of volatile pointers.  There are concurrent and single threaded uses of this class, so not sure if this changes helps, hurts, or is neutral.

If the type of incomingQueue were changed to BlockingQueue as suggested in another comment that may remove the concurrent use of this type, but not sure.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -639,6 +590,39 @@ public MutationWriter(int numSendThreads) {
...
+                    mutations.remove(mutation);
+                  });
+                });
+                addMutations(copy);
+              } catch (Exception e) {
+                updateUnknownErrors("Error processing mutation set", e);
+              }
+            } else {
+              try {
+                Thread.sleep(100);
```
keith-turner, line 615: If incomingQueue were of type `BlockingQueue<TableIDAndMutation>` the call to sleep could be avoided using `take()` followed by `drainTo()` or `poll()` methods on  BlockingQueue instead.

keith-turner (commented):
I have analyzed most of this change, but not all.  Thinking I will hold off on further analysis in case there are any changes based on these comments.

dlmarion:
@keith-turner  - thanks, looking at your comments now

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
dlmarion: I implemented your suggestion.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
dlmarion: I added the synchronized modifier back to `addMutation`

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
dlmarion: I added the synchronized modifier back to `addMutation`

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
dlmarion: I reverted this change.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
dlmarion: I changed `incomingQueue` to a BlockingQueue and `MutationSet` to extend HashMap.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
dlmarion: I sync'd this method

dlmarion:
This also got me thinking to something I found about 8 years ago, but we never fixed. https://issues.apache.org/jira/browse/ACCUMULO-2069. It looks like we might have fixed this in the Manager with:
```
    ThriftTransportPool.getInstance()
        .setIdleTime(aconf.getTimeInMillis(Property.GENERAL_RPC_TIMEOUT));
```

But, setIdleTime() is not called from anywhere else. So, after 3 seconds of being idle, the Thrift connection is removed from the ConnectionPool. When the client comes back around to trying to talk with the TabletServer, the Thrift connection has to be created again. Maybe it's different now, but back in 2013, increasing that timeout made a significant difference.

keith-turner:
> Maybe it's different now, but back in 2013, increasing that timeout made a significant difference.

It would be good to improve that if its still a problem.

dlmarion:
So, I'm working on a [benchmark](https://github.com/dlmarion/batchwriter-benchmark), but the [results](https://github.com/dlmarion/batchwriter-benchmark/blob/master/README) of this don't look too promising.

core/src/main/java/org/apache/accumulo/core/clientImpl/TabletServerBatchWriter.java
```diff
@@ -639,6 +607,39 @@ public MutationWriter(int numSendThreads) {
...
+              log.trace("{} - waiting for mutations to arrive", Thread.currentThread().getName());
+              try {
+                TableIdAndMutation tm = incomingQueue.take();
+                readyForBinning.add(tm.getTableId(), tm.getMutation());
+              } catch (InterruptedException e) {
+                if (!closed) {
+                  log.error("sleep interrupted", e);
+                }
+              }
+            }
```
keith-turner, line 638: Could do the following to handle close while waiting for data and handle all exceptions in one place.  

```suggestion
            try {
              TableIdAndMutation tm;
              while ((tm = incomingQueue.poll(100, TimeUnit.MILLISECONDS)) == null) {
                if (closed)
                  return;
              }
              readyForBinning.add(tm.getTableId(), tm.getMutation());
              while ((tm = incomingQueue.poll()) != null) {
                  readyForBinning.add(tm.getTableId(), tm.getMutation());
              }
              log.trace("{} - binning {} mutations", Thread.currentThread().getName(),
								readyForBinning.size());
              addMutations(readyForBinning);
              readyForBinning.clear();
            } catch (InterruptedException e) {
               if(!closed)
                  updateUnknownErrors("Error processing mutation set", e);
            } catch (Exception e) {
                updateUnknownErrors("Error processing mutation set", e);
            }
```

dlmarion:
So, I updated my benchmark to move the client and batchWriter creation to before the tested method, and the batchwriter and client close methods to after the tested method. The timed test method only adds the pre-created Mutations to the batch writer. Comparing the `main` branch and this branch, the times are almost identical. In the previous run, this branch took longer overall probably because the binning is being done earlier, therefore the batches are smaller and there are more round trips to the tablet server.

dlmarion:
I updated my benchmark to use 10 tablet servers and create splits for the table across the 10 servers. The time for the `main` branch and this branch are almost identical. I'm not sure that these modifications are improving anything at all.

PR #2270, opened 2021-09-14
Rename CompactionTask to InternalJob

milleruntime:
* Rename CompactionTask to InternalJob to be consistent with ExternalJob
* Rename some variables to match InternalJob type
* Fix some wording in comments
* Add javadoc to SubmittedJob

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionExecutor.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionExecutor.java
@@ -24,6 +24,12 @@ import org.apache.accumulo.core.dataImpl.KeyExtent;
 import org.apache.accumulo.core.spi.compaction.CompactionJob;
 import org.apache.accumulo.core.spi.compaction.CompactionServiceId;
 
+/**
+ * A non-pluggable component that executes compactions using multiple threads and has a priority
+ * queue. There are two types: Internal & External. The {@link InternalCompactionExecutor} runs
+ * within the Accumulo tserver process. The {@link ExternalCompactionExecutor} runs compactions
+ * outside the tserver.
+ */
 public interface CompactionExecutor {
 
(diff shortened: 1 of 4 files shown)


keith-turner (approved):

EdColeman (commented):
Without understanding the code - it seems that internalJob and externalJob miss a chance to provide context with what they are doing, esp. if strictly related to compactions.  Also, although task and job are interchangeable, task seems to imply something that is narrow in scope and maybe just a step - job seems to imply larger and self contained.  Like you submit a MR job that can contain multiple tasks. So not sure what is more appropriate in this context.

milleruntime:
> Without understanding the code - it seems that internalJob and externalJob miss a chance to provide context with what they are doing, esp. if strictly related to compactions. 

We could add "compaction" to the names but that seems unnecessary as they are only used within the `CompactionExecutor` and tserver.compaction package. If you follow the code from tserver, it is more clear then taken out of context in a PR. A `CompactionJob` is the immutable object that will get run. A `SubmittedJob` is a wrapper around a `CompactionJob`, that provides the status. Then `ExternalJob` and `InternalJob` are two types of `SubmittedJob` that can be run.

> Also, although task and job are interchangeable, task seems to imply something that is narrow in scope and maybe just a step - job seems to imply larger and self contained. Like you submit a MR job that can contain multiple tasks. So not sure what is more appropriate in this context.

It seems we could settle on Job here as a name for a unit of work. Perhaps @keith-turner may have had some other ideas?

keith-turner:
> Perhaps @keith-turner may have had some other ideas?

I think this change is good, it makes the type name more consistent with its sibling and parent types.  If there were to be a change it should probably encompass the entire type tree making them all consistent.  I am happy w/ `Job`, but there are possibilities like `plan`, `task`, `work`, etc..  The purpose of the parent type is to encapsulate information about compaction work that needs to be done.

milleruntime:
OK I will name the variables to be consistent with the `Job` types.

PR #2274, opened 2021-09-16
Make threadpools marked as mutable resizeable

### LINKED ISSUE #2272: Fate thread pool property listed as ZK mutable but does not dynamically resize.
**Describe the bug**

The property `manager.fate.threadpool.size`is listed as ZK mutable in the documentation implying that changes to it should cause changes in a running system.  However looking at the following code that creates the thread pool, I can not find a place in the code that changes the thread pool size after its created.

https://github.com/apache/accumulo/blob/ba83d8fa26a8adc92ca0249780b1ddf9d7135d7a/core/src/main/java/org/apache/accumulo/fate/Fate.java#L228-L234

**Expected behavior**

The thread pool should dynamically resize or the documentation should be updated to indicate its not ZK mutable.

dlmarion:
Closes #2272

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
+++ b/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
@@ -35,6 +35,29 @@ public class ThreadPools {
   // the number of seconds before we allow a thread to terminate with non-use.
   public static final long DEFAULT_TIMEOUT_MILLISECS = 180000L;
 
+  private static void makeResizeable(final ThreadPoolExecutor pool,
+      final AccumuloConfiguration conf, final Property p) {
+    final String threadName = p.name().concat("_watcher");
+    Threads.createThread(threadName, () -> {
+      int count = conf.getCount(p);
+      while (Thread.currentThread().isAlive() && !Thread.currentThread().isInterrupted()) {
+        try {
+          Thread.sleep(1000);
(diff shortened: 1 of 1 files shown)


dlmarion:
I found one other property that had the same issue

core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
```diff
@@ -58,7 +82,9 @@ public static ExecutorService createExecutorService(AccumuloConfiguration conf,
       case MANAGER_RENAME_THREADS:
         return createFixedThreadPool(conf.getCount(p), "bulk move", false);
       case MANAGER_FATE_THREADPOOL_SIZE:
-        return createFixedThreadPool(conf.getCount(p), "Repo Runner", false);
+        ThreadPoolExecutor pool = createFixedThreadPool(conf.getCount(p), "Repo Runner", false);
+        makeResizeable(pool, conf, p);
```
milleruntime, line 86: Seems like you could just have one method to create the resizeablePool.

dlmarion: I figured it would just be easier to make one method that could modify any of the created thread pools rather than making more methods to create a thread pool or adding more parameters to all of the existing methods.

milleruntime (approved):

PR #2275, opened 2021-09-17
Added additional tests in CompactionIT

### LINKED ISSUE #2114: Consolidate compaction functionality testing between internal and external compactions
External compactions were added in #2096 and in this change there is a new IT called ExternalCompactionIT.  This IT tests some things that specific to External compactions (like fault conditions) and it test some functionality that is common between external and internal compactions.  It would be good to consolidate testing of common functionality and write test that can run against both.  For example functionality like the following is common to both and it would be nice to write test once and run the test against both.

  * Setting compaction iterators
  * Selecting a subset of tablet files for a user compaction
  * Overriding file output settings for a compaction like compression
  * Compacting encrypted files
  * Etc

dlmarion:
Added tests in CompactionIT for:
  compaction cancelled on table during user compaction
  table delete during user compaction
  compaction of subset of files
  setting compression compaction iterator

Closes #2114


dlmarion:
I was unable to find a common root for the tests, so I copied them from ExternalCompactionIT to CompactionIT and modified them such that they would work for the internal compaction code

core/src/main/java/org/apache/accumulo/core/iterators/SleepyIterator.java
```diff
@@ -0,0 +1,80 @@
+/*
```
EdColeman, line 1: There is a similar functioning iterator with org.apache.accumulo.test.functional.SlowIterator.  Assuming that this iterator is used only for testing, would a test package be a better fit?

Also, many of the tests that use SlowIterator are flaky - beloved to be because of resource contention and those tests usually needing multiple, concurrent operations.  

- Would sleep iterator help with those tests?
- Are tests that use sleepy iterator likely to have similar issues?

dlmarion: I will look at moving this to a test package. For clarification, this iterator is only used in tests that are trying to make a compaction take long enough in MAC that the test thread can cancel or delete it. The iterator is not used in the main test thread, so I don't think it should make the test flaky. I'll have to look at SlowIterator to see how/where it's used and if SleepyIterator can be used in it's place or vice versa.

dlmarion: @EdColeman  - thanks for pointing out SlowIterator. I went ahead and just removed my iterator instead.

EdColeman: I was hoping yours was an improvement :-(

dlmarion: lol - no, looking at SlowIterator, mine duplicated it.

Manno15 (approved):

Issue #2276, opened 2021-09-17
Look into using new method for adjusting thread pool sizes more widely.


keith-turner:
Currently in the code there are three bits of code that adjust thread pool sizes as config changes.  Would it be possible to make these use the new method introduced in #2274?  Below are the places in the code that adjust thread pool sizes.

https://github.com/apache/accumulo/blob/ba83d8fa26a8adc92ca0249780b1ddf9d7135d7a/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServerResourceManager.java#L137-L160
https://github.com/apache/accumulo/blob/ba83d8fa26a8adc92ca0249780b1ddf9d7135d7a/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/InternalCompactionExecutor.java#L190-L206
https://github.com/apache/accumulo/blob/ba83d8fa26a8adc92ca0249780b1ddf9d7135d7a/server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionService.java#L429
https://github.com/apache/accumulo/blob/ba83d8fa26a8adc92ca0249780b1ddf9d7135d7a/server/base/src/main/java/org/apache/accumulo/server/rpc/TServerUtils.java#L314-L340

keith-turner:
The code in TServerUtils.java would probably need to continue to be custom as it starts a  configurable minimum and grows/shrinks based on usage.

PR #2277, opened 2021-09-17
fixes order of thread pool adjustment

keith-turner:
When looking into #2276 I noticed that some of the existing methods called thread pool methods in a very specific order depending on if they were increasing or decreasing.  I think this order used to matter and avoided bugs, not sure if it still does.  Opened this PR just in case it does matter, have not test it though.

Also copied a logging statement over from one of the existing methods.


dlmarion:
@keith-turner - I didn't realize that the order of those calls mattered. Good catch. Looks good to me.

dlmarion (approved):

EdColeman (approved):

dlmarion:
I included this change in https://github.com/apache/accumulo/pull/2278

keith-turner:
> I included this change in #2278

After #2278 is merged we can close this.

ctubbsii (approved):

dlmarion:
@keith-turner  - #2278 has been merged.

PR #2278, opened 2021-09-20
Resize FaTE thread pool correctly, attempt at making sure it's constantly running

### LINKED ISSUE #2272: Fate thread pool property listed as ZK mutable but does not dynamically resize.

dlmarion:
Closes #2272

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
+++ b/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
@@ -26,6 +26,7 @@ import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.function.IntSupplier;
 
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
 import org.apache.accumulo.core.conf.Property;
@@ -39,11 +40,24 @@ public class ThreadPools {
   // the number of seconds before we allow a thread to terminate with non-use.
   public static final long DEFAULT_TIMEOUT_MILLISECS = 180000L;
 
-  public static void makeResizeable(final ThreadPoolExecutor pool, final AccumuloConfiguration conf,
-      final Property p) {
+  /**
+   * Resize ThreadPoolExecutor based on current value of maxThreads
+   *
+   * @param pool
+   *          the ThreadPoolExecutor to modify
+   * @param maxThreads
+   *          supplier of maxThreads value
+   * @param poolName
+   *          name of the thread pool
+   */
+  public static void resizePool(final ThreadPoolExecutor pool, final IntSupplier maxThreads,
+      String poolName) {
     int count = pool.getMaximumPoolSize();
-    int newCount = conf.getCount(p);
-    LOG.info("Changing max threads for {} from {} to {}", p.getKey(), count, newCount);
+    int newCount = maxThreads.getAsInt();
+    if (count == newCount) {
+      return;
+    }
+    LOG.info("Changing max threads for {} from {} to {}", poolName, count, newCount);
     if (newCount > count) {
       // increasing, increase the max first, or the core will fail to be increased
       pool.setMaximumPoolSize(newCount);
@@ -53,6 +67,22 @@ public class ThreadPools {
       pool.setCorePoolSize(newCount);
       pool.setMaximumPoolSize(newCount);
     }
+
+  }
+
+  /**
+   * Resize ThreadPoolExecutor based on current value of Property p
+   *
+   * @param pool
+   *          the ThreadPoolExecutor to modify
+   * @param conf
+   *          the AccumuloConfiguration
+   * @param p
+   *          the property to base the size from
+   */
+  public static void resizePool(final ThreadPoolExecutor pool, final AccumuloConfiguration conf,
+      final Property p) {
+    resizePool(pool, () -> conf.getCount(p), p.getKey());
   }
 
   /**
@@ -79,9 +109,7 @@ public class ThreadPools {
       case MANAGER_RENAME_THREADS:
         return createFixedThreadPool(conf.getCount(p), "bulk move", false);
       case MANAGER_FATE_THREADPOOL_SIZE:
-        ThreadPoolExecutor pool = createFixedThreadPool(conf.getCount(p), "Repo Runner", false);
-        makeResizeable(pool, conf, p);
-        return pool;
+        return createFixedThreadPool(conf.getCount(p), "Repo Runner", false);
       case MANAGER_STATUS_THREAD_POOL_SIZE:
         int threads = conf.getCount(p);
         if (threads == 0) {
@@ -114,10 +142,7 @@ public class ThreadPools {
       case GC_DELETE_THREADS:
         return createFixedThreadPool(conf.getCount(p), "deleting", false);
       case REPLICATION_WORKER_THREADS:
-        ThreadPoolExecutor pool2 =
-            createFixedThreadPool(conf.getCount(p), "replication task", false);
-        makeResizeable(pool2, conf, p);
-        return pool2;
+        return createFixedThreadPool(conf.getCount(p), "replication task", false);
       default:
         throw new RuntimeException("Unhandled thread pool property: " + p);
     }
--- a/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
+++ b/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
@@ -29,33 +29,30 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.accumulo.core.conf.AccumuloConfiguration;
 import org.apache.accumulo.core.conf.Property;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ThreadPools {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ThreadPools.class);
 
   // the number of seconds before we allow a thread to terminate with non-use.
   public static final long DEFAULT_TIMEOUT_MILLISECS = 180000L;
 
-  private static void makeResizeable(final ThreadPoolExecutor pool,
-      final AccumuloConfiguration conf, final Property p) {
-    final String threadName = p.name().concat("_watcher");
-    Threads.createThread(threadName, () -> {
-      int count = conf.getCount(p);
-      while (Thread.currentThread().isAlive() && !Thread.currentThread().isInterrupted()) {
-        try {
-          Thread.sleep(1000);
-          int newCount = conf.getCount(p);
-          if (newCount != count) {
-            pool.setCorePoolSize(newCount);
-            pool.setMaximumPoolSize(newCount);
-            count = newCount;
-          }
-        } catch (InterruptedException e) {
-          // throw a RuntimeException and let the AccumuloUncaughtExceptionHandler deal with it.
-          throw new RuntimeException("Thread " + threadName + " was interrupted.");
-        }
-      }
-    }).start();
-
+  public static void makeResizeable(final ThreadPoolExecutor pool, final AccumuloConfiguration conf,
+      final Property p) {
+    int count = pool.getMaximumPoolSize();
+    int newCount = conf.getCount(p);
+    LOG.info("Changing max threads for {} from {} to {}", p.getKey(), count, newCount);
+    if (newCount > count) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
```diff
@@ -29,33 +29,30 @@
...
-            count = newCount;
-          }
-        } catch (InterruptedException e) {
-          // throw a RuntimeException and let the AccumuloUncaughtExceptionHandler deal with it.
-          throw new RuntimeException("Thread " + threadName + " was interrupted.");
-        }
-      }
-    }).start();
-
+  public static void makeResizeable(final ThreadPoolExecutor pool, final AccumuloConfiguration conf,
```
keith-turner, line 42: This method name seems off w/ the changes.
```suggestion
  public static void resizePool(final ThreadPoolExecutor pool, final AccumuloConfiguration conf,
```

core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
```diff
@@ -29,33 +29,30 @@
...
+      final Property p) {
+    int count = pool.getMaximumPoolSize();
+    int newCount = conf.getCount(p);
```
keith-turner, line 45: Thinking the following may avoid logging when there is nothing to do.

```suggestion
    int newCount = conf.getCount(p);
    if(count == newCount) 
       return;
```

core/src/main/java/org/apache/accumulo/fate/Fate.java
```diff
@@ -226,11 +223,17 @@ public Fate(T environment, TStore<T> store, Function<Repo<T>,String> toLogStrFun
...
-      executor.execute(new TransactionRunner());
-    }
+    final ThreadPoolExecutor pool = (ThreadPoolExecutor) ThreadPools.createExecutorService(conf,
+        Property.MANAGER_FATE_THREADPOOL_SIZE);
+    fatePoolWatcher = ThreadPools.createGeneralScheduledExecutorService(conf);
+    fatePoolWatcher.schedule(() -> {
+      ThreadPools.makeResizeable(pool, conf, Property.MANAGER_FATE_THREADPOOL_SIZE);
+      int remaining = 1000 - pool.getQueue().size();
+      for (int i = 0; i < remaining; i++) {
+        pool.execute(new TransactionRunner());
```
keith-turner, line 233: This may throw a bunch of exceptions if the pool is shutdown.  Would be nice to avoid that.  Could wait for `fatePoolWatcher` to shutdown before calling `executor.shutdown()` in the shutdown() method below in this class.  That might avoid the exceptions here.

Alternatively could catch rejectedexecutionexception here and ignore it if the pool is shutdown.

core/src/main/java/org/apache/accumulo/fate/Fate.java
```diff
@@ -226,11 +223,17 @@ public Fate(T environment, TStore<T> store, Function<Repo<T>,String> toLogStrFun
...
    */
   public void startTransactionRunners(AccumuloConfiguration conf) {
-    int numThreads = conf.getCount(Property.MANAGER_FATE_THREADPOOL_SIZE);
-    executor = ThreadPools.createExecutorService(conf, Property.MANAGER_FATE_THREADPOOL_SIZE);
-    for (int i = 0; i < numThreads; i++) {
-      executor.execute(new TransactionRunner());
-    }
+    final ThreadPoolExecutor pool = (ThreadPoolExecutor) ThreadPools.createExecutorService(conf,
+        Property.MANAGER_FATE_THREADPOOL_SIZE);
+    fatePoolWatcher = ThreadPools.createGeneralScheduledExecutorService(conf);
```
keith-turner, line 228: The Accumulo servers used to have a shared timer pool.  Not sure if that is still a thing, if it is could be used to avoid creating these watcher threads.

core/src/main/java/org/apache/accumulo/fate/Fate.java
```diff
@@ -226,11 +223,17 @@ public Fate(T environment, TStore<T> store, Function<Repo<T>,String> toLogStrFun
...
+    fatePoolWatcher.schedule(() -> {
+      ThreadPools.makeResizeable(pool, conf, Property.MANAGER_FATE_THREADPOOL_SIZE);
+      int remaining = 1000 - pool.getQueue().size();
```
keith-turner, line 231: Could replace 1000 w/ a function of the numThreads and time like below.

```suggestion
      // Assume a thread could execute up to 100 operations a second.  We are sleeping for three seconds. So to calculate max to queue do 3* 100 * numThreads. 
      int maxToQueue = 300 * conf.getCount(  Property.MANAGER_FATE_THREADPOOL_SIZE);
      int remaining = 1000 - pool.getQueue().size();
```

core/src/main/java/org/apache/accumulo/fate/Fate.java
dlmarion: I looked back in 1.10 for shared SimpleTimer's. Please let me know if you happen to have a location on this as I would be interested to see if I missed something when I did the Threads/ThreadPools work. In this specific case, only 1 scheduled runnable is going to be created as `startTransactionRunners` is only called once in Manager.run().

core/src/main/java/org/apache/accumulo/fate/Fate.java
dlmarion: Thanks for this, I wasn't sure what a reasonable value would be.

core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
dlmarion: I included this in 8bed9e5.

core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
dlmarion: I included this in 8bed9e5.

core/src/main/java/org/apache/accumulo/fate/Fate.java
dlmarion: I addressed this in 8bed9e5.

core/src/main/java/org/apache/accumulo/fate/Fate.java
dlmarion: I included this in 8bed9e5.

core/src/main/java/org/apache/accumulo/fate/Fate.java
```diff
@@ -226,11 +224,26 @@ public Fate(T environment, TStore<T> store, Function<Repo<T>,String> toLogStrFun
...
+      // Assume a thread could execute up to 100 operations a second.
+      // We are sleeping for three seconds. So to calculate max to queue do 3* 100 * numThreads.
+      int maxToQueue = 300 * conf.getCount(Property.MANAGER_FATE_THREADPOOL_SIZE);
+      int remaining = maxToQueue - pool.getQueue().size();
+      for (int i = 0; i < remaining; i++) {
+        try {
+          pool.execute(new TransactionRunner());
+        } catch (RejectedExecutionException e) {
+          // RejectedExecutionException could be shutting down
+          log.warn("Error adding transaction runner to FaTE executor pool.");
```
keith-turner, line 241: I think if its shutdown we can avoid spamming the logs.  If its not shutdown, then its really bad and we should include the exception.  For the case where its not shutdown, should we log an error or rethrow the exception?

```suggestion
      if(pool.isShutdown()){
          // The exception is expected in this case, no need to spam the logs.
          log.trace("Error adding transaction runner to FaTE executor pool.",e);
       } else {
        // This is bad, FaTE may no longer work!
         log.error("Error adding transaction runner to FaTE executor pool.",e);
       }
```

server/base/src/main/java/org/apache/accumulo/server/rpc/TServerUtils.java
```diff
@@ -323,16 +323,11 @@ public static ThreadPoolExecutor createSelfResizingThreadPool(final String serve
...
-        pool.setCorePoolSize(larger);
+        ThreadPools.resizePool(pool, () -> larger, serverName + "-ClientPool");
       } else {
         if (pool.getCorePoolSize() > pool.getActiveCount() + 3) {
           int smaller = Math.max(executorThreads, pool.getCorePoolSize() - 1);
-          if (smaller != pool.getCorePoolSize()) {
-            log.info("Decreasing server thread pool size on {} to {}", serverName, smaller);
-            pool.setCorePoolSize(smaller);
-          }
+          ThreadPools.resizePool(pool, () -> smaller, serverName + "-ClientPool");
```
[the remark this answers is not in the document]
dlmarion: So, it would [appear](https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.base/share/classes/java/util/concurrent/ThreadPoolExecutor.java#L1533) that lowering the corePoolSize below the MaxPoolSize throws an error. Maybe this is what you were referring to?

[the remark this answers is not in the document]
keith-turner: I just noticed that this code was no longer making the same method calls on the thread pool that it used to make, however I was not sure if that change mattered.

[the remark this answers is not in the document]
keith-turner: Do you think the existing code before your change had a bug?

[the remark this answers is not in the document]
dlmarion: Right, it's now using the common way of doing the resizing.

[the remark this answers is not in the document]
keith-turner: hmm so maybe this code never achieved its intent.. I wonder if it only increased but never decreased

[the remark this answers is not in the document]
dlmarion: I can't say about before my change, but in ThreadPools.java as it is now, if the thread pool was created with a pool size of 5 and no max pool specified, then the max pool size would also be set to 5. If the core pool size was then decreased, it would error.

core/src/main/java/org/apache/accumulo/fate/Fate.java
keith-turner: Looking back at 1.10 I See the following which created a singleton timer that was shared in the process with a configurable number of threads.

https://github.com/apache/accumulo/blob/bbe2de8a8c00eb0b7cc01766ac6e6db672bf8444/server/base/src/main/java/org/apache/accumulo/server/util/time/SimpleTimer.java#L60-L71

AFAICT now in 2.1.0 there is no longer a process wide singleton timer related to the config `general.server.simpletimer.threadpool.size`?

dlmarion: That was my concern, that maybe I missed that. However, now that the creation of that ThreadPool is singularly located in `ThreadPools.createGeneralScheduledExecutorService`, it should be easy to make it a singleton that is only created if it does not already exist. It's called from 33 locations across the code base.

keith-turner (approved):

core/src/main/java/org/apache/accumulo/fate/Fate.java
dlmarion: It is used by the TabletServerBatchWriter. I think putting it on the ServerContext would render that unreachable.

dlmarion: I might be able to put it on AccumuloConfiguration.

keith-turner: > It is used by the TabletServerBatchWriter. I think putting it on the ServerContext would render that unreachable.

May make sense to have it in the client context for client side code.  Server context extends client context, so maybe that would work out nicely.

milleruntime:
I think this change still might be a problem. It is causing some of our previously stable ITs to timeout. I am not sure why but they were fine before the initial commit of https://github.com/apache/accumulo/pull/2274. The two ITs timing out are MergeIT and ShellServerIT

PR #2283, opened 2021-09-22
Add Active Compactions page to Monitor

### LINKED ISSUE #2230: Display more compaction information in the monitor
Original JIRA ticket: https://issues.apache.org/jira/browse/ACCUMULO-3154

Important excerpt: 
<pre>
Showing the age of the longest running scan...is helping users to understand their system better.
 It was suggested that showing the oldest compaction would be useful, too...
</pre>

An 'Active Compactions' page in the monitor similar to that of 'Active Scans' where the server, count, and oldest age is displayed.

milleruntime:
* Closes #2230
* Add code required to populate active compactions on new page
* Rename scans page title to Active Scans
* Rename ScansResource method to getActiveScans()


keith-turner:
@milleruntime  can you share a screenshot of what this looks like?

milleruntime:
> @milleruntime can you share a screenshot of what this looks like?

It looks almost identical to the Active Scans page but I used datatables for the table
monitor
.

server/monitor/src/main/java/org/apache/accumulo/monitor/rest/compactions/Compactions.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package org.apache.accumulo.monitor.rest.compactions;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Generates a new list of compactions as a JSON object
+ *
+ * @since 2.1
```
Manno15, line 27:
```suggestion
 * @since 2.1.0
```

server/monitor/src/main/java/org/apache/accumulo/monitor/rest/compactions/CompactionsResource.java
```diff
@@ -0,0 +1,69 @@
...
+
+import org.apache.accumulo.core.manager.thrift.ManagerMonitorInfo;
+import org.apache.accumulo.core.master.thrift.TabletServerStatus;
+import org.apache.accumulo.core.util.HostAndPort;
+import org.apache.accumulo.monitor.Monitor;
+
+/**
+ * Generate a new Compaction list JSON object
+ *
+ * @since 2.0.0
```
Manno15, line 37: Same as the above comment.

Manno15 (approved):
Changes look good. The monitor does seem slightly inconsistent with text alignment (inside each column) but I don't really have a preference either way.

milleruntime:
> The monitor does seem slightly inconsistent with text alignment (inside each column)

Good catch. The default text alignment for datatables is different from our custom jquery generated tables. So it might be easier to fix them across all pages. I can do that in a separate commit.

milleruntime:
Created #2286 for fixing text alignment across all tables.

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
```diff
@@ -484,6 +485,17 @@ public void run() {
...
+      while (true) {
+        try {
+          fetchCompactions();
+        } catch (Exception e) {
+          log.warn("{}", e.getMessage(), e);
+        }
+        sleepUninterruptibly(5, TimeUnit.SECONDS);
+      }
+    }).start();
+
```
ctubbsii, line 498: The problem with this pattern is that there's a constant background thread doing RPC calls to the tservers to collect these stats, even if nobody is accessing the page on the monitor. And, it has to talk to every tserver, of which there could be thousands. This is a lot of constant activity that might not even be needed. The general pattern of the monitor is to avoid refreshing data if nobody is accessing the page, and to limit the refresh of the data shown on the page to collect it no more frequently than once every 5 seconds. This is how we collect the "mmi" object, anyway.

Fast display should not bog down the cluster like this.

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
```diff
@@ -553,7 +565,24 @@ private ServletHolder getRestServlet() {
...
+    public final long fetched;
+
+    CompactionStats(List<ActiveCompaction> active) {
+      this.count = active.size();
+      long oldest = -1;
+      for (ActiveCompaction a : active) {
+        oldest = Math.max(oldest, a.age);
+      }
+      this.oldest = oldest < 0 ? null : oldest;
+      this.fetched = System.currentTimeMillis();
```
ctubbsii, line 580: This should use nano time to compute time durations.

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
```diff
@@ -589,6 +624,33 @@ private void fetchScans() throws Exception {
...
+      } finally {
+        ThriftUtil.returnClient(tserver);
+      }
+    }
+    // Age off old compaction information
+    var entryIter = allCompactions.entrySet().iterator();
+    long now = System.currentTimeMillis();
+    while (entryIter.hasNext()) {
+      var entry = entryIter.next();
+      if (now - entry.getValue().fetched > 5 * 60 * 1000) {
```
ctubbsii, line 648: This should use nanoTime to compute time durations, to avoid system clock adjustments affecting the computation.

server/monitor/src/main/java/org/apache/accumulo/monitor/rest/compactions/CompactionInfo.java
```diff
@@ -0,0 +1,53 @@
...
+   * Stores new compaction information
+   *
+   * @param tserverInfo
+   *          status of the tserver
+   * @param count
+   *          number of compactions
+   * @param oldest
+   *          time of oldest compaction
+   */
+  public CompactionInfo(TabletServerStatus tserverInfo, long count, Long oldest) {
```
ctubbsii, line 48: Not sure, but OptionalLong might be better here, rather than passing a nullable boxed version of Long.

ctubbsii (commented):
@milleruntime I have performance concerns about constantly collecting information from RPC calls to every tablet server every 5 seconds. I think that could be handled better.

milleruntime:
> @milleruntime I have performance concerns about constantly collecting information from RPC calls to every tablet server every 5 seconds. I think that could be handled better.

I could make the stats for the page disabled by default.

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
milleruntime: A lot of this code was copy and pasted but I can do a change replacing `System.currentTimeMillis()` in the Monitor.

EdColeman:
The utility of seeing running compactions seems that it would be highest for long running compactions - polling on an order of minutes might provide enough fidelity to still be usefull.

dlmarion:
What is "Oldest Age" column in the UI? Is that the time since the last response time from the tserver? Are you planning on showing anything other than a count? `ThriftClientHandler.getActiveCompactions` only returns information about the compactions running internally. However, the TabletServer also has information on the compactions that are running externally. Conveying just the count of internal and external metrics might require a new Thrift endpoint though as I don't think the information will fit into `getActiveCompactions`.

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
milleruntime: > The problem with this pattern is that there's a constant background thread doing RPC calls to the tservers to collect these stats, even if nobody is accessing the page on the monitor. And, it has to talk to every tserver, of which there could be thousands. This is a lot of constant activity that might not even be needed. The general pattern of the monitor is to avoid refreshing data if nobody is accessing the page, and to limit the refresh of the data shown on the page to collect it no more frequently than once every 5 seconds. This is how we collect the "mmi" object, anyway.
> 
> Fast display should not bog down the cluster like this.

This is also currently being done for active scans. So if it's an issue with the fetch compactions thread, we should do something with the scans fetch as well.

milleruntime:
> What is "Oldest Age" column in the UI? Is that the time since the last response time from the tserver? Are you planning on showing anything other than a count? 

That is the age returned by the thrift object `ActiveCompaction`.

> `ThriftClientHandler.getActiveCompactions` only returns information about the compactions running internally. However, the TabletServer also has information on the compactions that are running externally. Conveying just the count of internal and external metrics might require a new Thrift endpoint though as I don't think the information will fit into `getActiveCompactions`.

This would be a good opportunity to include External compaction information as well. Now that you mention it, I don't think the Monitor has anything about External compactions.

dlmarion:
IIRC, the compactions queued and running in the `tables` view does include external compactions. I'm tracing through the code now, it looks like the `tables` view gets that information from the Manager.

milleruntime:
Another option would be to add a "Fetch Compactions" button on the Monitor to make it a manual process performed by the admin.

EdColeman:
One thing that is nice when a manual (or slow refresh) is to also display the time that the page was fetched - that way you can determine when the results were actually gathered.

dlmarion:
The LiveTServerSet in the Manager is already collecting the stats, that won't work?

milleruntime:
> The LiveTServerSet in the Manager is already collecting the stats, that won't work?

Possibly. I know that I have had issues getting that information from the Manager in the past.

milleruntime:
> The LiveTServerSet in the Manager is already collecting the stats, that won't work?

The Monitor uses the `ManagerMonitorInfo` object to get all its information by calling getManagerStats() on the manager client.

dlmarion:
`ManagerMonitorInfo.tServerInfo.tableMap` values are `TableInfo` objects which contain the number of queued and running major compactions. Just need to sum them up for each tserver.

milleruntime:
Yeah that is what we do for the tables page in https://github.com/apache/accumulo/blob/f2b0936029fcaadb925bd9c229403c82f0283211/server/monitor/src/main/java/org/apache/accumulo/monitor/rest/tables/TablesResource.java#L91

But that doesn't have the age of the compactions. That is in a separate RPC call for some reason.

dlmarion:
Can you use the last contact field in the TabletServerStatus? I assume that is what the `/tables` display is using.

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
ctubbsii: > This is also currently being done for active scans. So if it's an issue with the fetch compactions thread, we should do something with the scans fetch as well.

Yeah, I saw that. It seems like it's a problem there, too. I'm not sure when that got added, but we definitely don't need the monitor constantly talking to every server in the cluster every 5 seconds, if the monitor doesn't even have any active clients requesting pages.

milleruntime:
FYI I created a follow on to include something about external compactions in the monitor: https://github.com/apache/accumulo/issues/2290

server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
milleruntime: I am working on the idea to have a button to fetch the info. I like the manual fetch in this case. We can still restrict it to only fetch every minute so an Admin can't spam the tservers. This way if no one cares, there won't ever be any wasted work. If a user does care, they will have the info provided on the page through an AJAX call.

milleruntime:
I created #2294 as a follow on for the concerns discussed in this PR.

PR #2305, opened 2021-10-06
Replace hadoop metrics with micrometer

dlmarion:
Supersedes #2188


dlmarion:
I'm thinking that we should remove the [Rate](https://github.com/apache/accumulo/blob/main/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/Rate.java) class and the metrics used by it. Rates can be calculated by downstream systems.

PR #2307, opened 2021-10-06
Replace thrown exception with log warning for custom FileSystem implementations

cbevard1:
The underlying FileSystem implementation is abstracted in Accumulo, but custom FileSystem implementations are restricted by this section of code in the HadoopLogCloser. I'm working on an S3 classpath extension that would allow S3 to be used as the underlying file system, but this thrown exception makes that impossible.

Instead of throwing an exception here couldn't the log closer simply log a warning? It's very unlikely someone would mistakenly make all the configuration changes necessary in order for this like of code to be executed.

I considered making the custom FileSystem implementation extend DistributedFileSystem instead, but that class is too tightly coupled to HDFS for that route to make sense.


cbevard1:
Oh, I see. I can set the property manager.wal.closer.implementation so use another implementation of LogCloser. That should work perfectly. Thanks.

PR #2309, opened 2021-10-07
fixes #2111 adds unit test for external compaction metadata validation

### LINKED ISSUE #2111: Need to test rebuild of selected set on tablet load.
When a tablet it initiates a user compaction it selects a set of file and then works to get those files compacted across possibly multiple compactions.  For external compactions the set of selected files file is rebuilt when a tablet unload and reloads on another tserver.  If possible this scenario should be tested, if its not already.

keith-turner:


keith-turner:
> FYI the phrase "sanity check" currently gets flagged by the new Apache CLC - Conscious Language Checker tool https://clc.diversity.apache.org/

@milleruntime I updated the wording for those comments in 72e49b97a31476860256fd2a4913140a77f549fc

Issue #2184, opened 2021-06-30
Thrift exception during upgrade


milleruntime:
I was testing upgrade code for upgrading 2.0 to 2.1 and saw this exception quite a bit. 
<pre>
2021-06-30T10:45:31,675 [thrift.ProcessFunction] ERROR: Internal error processing reportTabletStatus
org.apache.accumulo.core.clientImpl.thrift.ThriftNotActiveServiceException: null
        at org.apache.accumulo.server.rpc.HighlyAvailableServiceInvocationHandler.invoke(HighlyAvailableServiceInvocationHandler.java:53) ~[accumulo-server-base-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at com.sun.proxy.$Proxy38.reportTabletStatus(Unknown Source) ~[?:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[?:?]
        at jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
        at org.apache.accumulo.core.trace.TraceUtil.lambda$wrapService$6(TraceUtil.java:235) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at com.sun.proxy.$Proxy38.reportTabletStatus(Unknown Source) ~[?:?]
        at org.apache.accumulo.core.manager.thrift.ManagerClientService$Processor$reportTabletStatus.getResult(ManagerClientService.java:1874) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.core.manager.thrift.ManagerClientService$Processor$reportTabletStatus.getResult(ManagerClientService.java:1855) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.thrift.ProcessFunction.process(ProcessFunction.java:38) ~[libthrift-0.12.0.jar:0.12.0]
        at org.apache.thrift.TBaseProcessor.process(TBaseProcessor.java:39) ~[libthrift-0.12.0.jar:0.12.0]
[...]

PR #2321, opened 2021-10-19
Handle case where manager isn't active due to upgrade

Manno15:
Solution for #2184. 

An exception would be thrown by `HighlyAvailableServiceInvocationHandler` if the manager wasn't initialized (set by an atomic boolean inside of Manager) by the time it got a ManagerMessage from a tserver. The initialization process gets delayed due to things upgrading in the background (we wait for metadata to finish upgrading before proceeding to allow fate usage for example).

This proof of concept solution adds another boolean to keep track of the manager upgrade status. So the exception will only be thrown if the manager isn't initialized due to an error instead of during an upgrade. 

WIP still as I am unsure on what `HighlyAvailableServiceInvocationHandler` should do if it notices manager isn't active due to upgrade. The placements of when to set and unset the new boolean could also be tweaked.


Manno15:
Still testing around for what to return. Things are seemingly in working order even when returning null.

PR #2325, opened 2021-10-22
Revert "Move AbstractLexicoder to API (#1400)"

milleruntime:
This reverts commit 2e8144ec89e74d37c7ff97d301051fea747e05c7.

The move of this class breaks API compatibility in the class hierarchy. I am not sure how the parent class being in the `clientImpl` is not picked up by the apilyzer plugin. It could be since the super class of AbstractLexicoder in `clientImpl` happens to be in the public API that it doesn't actually expose any internal class. While it is awkward that AbstractLexicoder is in a different non-API package, it seems that it ultimately doesn't matter for the public API.


milleruntime:
Closing since I don't think we want to do this revert.

PR #2345, opened 2021-11-03
Create tool to validate Compaction configuration

### LINKED ISSUE #2313: Tool to validate Compaction configuration
It would be nice to have a CLI tool to run against a compaction configuration to allow you to check it before deploying. There are a lot of requirements and differences between `internal` and `external` Compactable types that can be tricky and probably won't get caught until the compaction manager tries to create them as a compaction service. These would fall under the `tserver.compaction.major.service` property, specifically the JSON used to configure `planner.opts.executors`.

DomGarguilo:
Fixes #2313

* CLI tool to verify a compaction configuration


DomGarguilo:
As is stands, this currently only accepts a JSON used to configure `planner.opts.executors`. I wanted to put up a PR to elicit suggestions while I'm working. I'm planning to add test cases to cover all the errors that are tested for by this tool.

I wanted to get input on:
* What other verification should be done?
* Should this tool only accept a JSON string as input? (or should should it accept a whole config file or something else?)

milleruntime:
Others might have ideas but my use case was a command to run to prevent deployment errors that could be checked ahead of time. One possibility is to pass a properties file to the command. Another option could be to pass a property to check in the current running configuration. When a user sets a property with JSON in the shell it is very prone to errors so any way to avoid that as well would be great. I am just throwing some ideas out there since I didn't have specific requirements but I think you are on the right track.

DomGarguilo:
> One possibility is to pass a properties file to the command

That's what I was thinking originally. I'll go ahead and try that.

DomGarguilo:
> One possibility is to pass a properties file to the command

I made this change and fixed some test failures with the latest changes

DomGarguilo:
The current changes accept a file as input then check verify the contents of that file. It seems like there should be more verification of the properties (possibly verifying the planner is set correctly) but I'm not sure what else should be added. Suggestions are welcome.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
```diff
@@ -0,0 +1,154 @@
...
+    // Convert json string to array of ExecutorConfig objects
+    var executorConfigs =
+        new Gson().fromJson(executorJson, DefaultCompactionPlanner.ExecutorConfig[].class);
+    if (Objects.isNull(executorConfigs)) {
+      throw new RuntimeException("Compaction executors could not be parsed from given file");
+    }
+
+    boolean hasSeenNullMaxSize = false;
+
+    for (var executorConfig : executorConfigs) {
```
dlmarion, line 91: Suggest re-using the logic at https://github.com/apache/accumulo/blob/main/core/src/main/java/org/apache/accumulo/core/spi/compaction/DefaultCompactionPlanner.java#L158 so that it's not done in two places.

DomGarguilo: Addressed in 13b39cb

DomGarguilo:
In 13b39cb I ended up moving things around so that we are able to use the logic in DefaultCompactionPlanner to verify the compaction config. I think with this changed, many of the tests I've written here can be removed because they are mostly testing  DefaultCompactionPlanner logic (which has its own test) more than CheckCompactionConfig.

dlmarion (approved):

test/src/main/java/org/apache/accumulo/test/start/KeywordStartIT.java
```diff
@@ -161,10 +163,12 @@ public void checkHasMain() {
 
     HashSet<Class<?>> expectSet = new HashSet<>();
     expectSet.add(Admin.class);
+    expectSet.add(CheckCompactionConfig.class);
     expectSet.add(CreateToken.class);
     expectSet.add(Info.class);
     expectSet.add(Initialize.class);
     expectSet.add(LoginProperties.class);
+    // should this be replaced?
```
milleruntime, line 171: Leftover comment from previous change?

DomGarguilo: whoops, yep I'll get rid of that soon.

DomGarguilo:
I'm working on getting rid of some of the tests that are redundant (overlapping with DefaultCompactionPlannerTest)

DomGarguilo:
The most recent changes replace the error throwing with `System.err.println()` and `System.exit()` so that the errors can be seen from the shell. I removed most of the tests since im not too sure how we can verify which errors are being thrown while using `System.exit()` but from my manual testing, everything seems to be working.

milleruntime:
> The most recent changes replace the error throwing with `System.err.println()` and `System.exit()` so that the errors can be seen from the shell. I removed most of the tests since im not too sure how we can verify which errors are being thrown while using `System.exit()` but from my manual testing, everything seems to be working.

You should be able to see Exceptions thrown by the utility. Maybe something is wrong with your Log4j.properties file? How are you running the command? The only time I think we want to use `System.out.println()` is for a standalone command line tool that can pipe the output to other CLI tools and you don't want logging information mucking up the output. We only want to use `System.exit()` in very rare circumstances. There was work done in Accumulo over the years to avoid calls to `System.exit()`. And I can't think of a reason to call `System.err.println()`.

DomGarguilo:
> You should be able to see Exceptions thrown by the utility. Maybe something is wrong with your Log4j.properties file?

Maybe so, I was having trouble seeing the exceptions. It would just quietly fail.

> How are you running the command?

I would start up an instance with fluo-uno then call `accumulo check-compaction-config <file path>`.

For some reason if I catch the exceptions and println their message it works consistently but if I just let the exception be thrown, that does not appear.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    // Create a props object for compaction props
+    StringReader sr = new StringReader(compactionPropertiesString.replace(' ', '\n'));
+    Properties serverProps = new Properties();
+    serverProps.load(sr);
+
+    // Extract executors options from compactions props
+    Map<String,String> executorsProperties =
+        getPropertiesWithSuffix(serverProps, ".planner.opts.executors");
```
keith-turner, line 117: Multiple compaction services can be configured, each with its own planner and options.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
```diff
@@ -0,0 +1,180 @@
...
+          @Override
+          public CompactionExecutorId getExternalExecutor(String name) {
+            return CompactionExecutorIdImpl.externalId(name);
+          }
+        };
+      }
+    };
+
+    try {
+      new DefaultCompactionPlanner().parseExecutors(params);
```
keith-turner, line 162: Did you consider using the planner property to create the planner class instead of assuming its the DefaultPlanner?

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: Whoops, good catch, ill make sure to incorporate that.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: I hadn't, but now that you mention it that does make more sense. I'll look into that.

DomGarguilo: I am having trouble implementing this because I would need to call `init()` rather than `parseExecutors()` (a method that I created to extract and separate the logic in `init`). This is an issue because `init` relies on objects that will not exist while using this tool. For example, if calling `init()` for DefaultCompactionPlanner, which in turn calls `determineMaxFilesToCompact()`, the params object that is passed in would need to have reference to a `ServiceEnvironment` object. This maybe be an issue for other implementations of `CompactionPlanner` as well since we do not know what will be required within the passed in `CompactionPlanner.InitParameters`.

I'm still looking into whether there is a workaround for this.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: Addressed in 4232ea1

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
keith-turner: @DomGarguilo I attempted to refactor some existing code so that it could be used by this utility.  I did that in this [commit](https://github.com/keith-turner/accumulo/commit/b17492345b777b5a8fc267c64b3e02b2333b472b) that is in this [branch](https://github.com/keith-turner/accumulo/commits/DomGarguilo-checkCompactionConfig).

Its not complete and I am not completely sure it will work, but I suspect it will.  Feel free to use it and build on it if you think its workable.

DomGarguilo: Thanks @keith-turner! I think what you have looks good. I'm working through a few details but I think its going to work.

DomGarguilo:
The newest changes seem to be working except they only allow for specifically formatted properties. For example, if using nested properties like those [found in accumulo testing](https://github.com/apache/accumulo-testing/blob/main/conf/accumulo-testing.properties#L43-L48):
```
test.ci.common.accumulo.server.props=\
tserver.compaction.major.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner \
tserver.compaction.major.service.cs1.planner.opts.executors=\
[{"name":"small","type":"internal","maxSize":"16M","numThreads":8},\
{"name":"medium","type":"internal","maxSize":"128M","numThreads":4},\
{"name":"large","type":"internal","numThreads":2}]
```
The tool will not find the executor config since it is within `test.ci.common.accumulo.server.props`. We might be able to account for this by accepting an optional argument that points to the parent property (in this case `test.ci.common.accumulo.server.props`). This example is one of the test cases that is currently failing.

For contrast, the following properties work fine with this tool:
```
tserver.compaction.major.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner 
tserver.compaction.major.service.cs1.planner.opts.executors=\
[{"name":"small","type":"internal","maxSize":"16M","numThreads":8},\
{"name":"medium","type":"internal","maxSize":"128M","numThreads":4},\
{"name":"large","type":"internal","numThreads":2}]
```

milleruntime:
FYI while looking at #2386, it made me think we need a catch all server utility for external compactions. I think this utility may be a good candidate to fall under that as well. I was thinking maybe we could have an "ec-admin" command, similar to the current "admin" command.

DomGarguilo:
> FYI while looking at #2386, it made me think we need a catch all server utility for external compactions. I think this utility may be a good candidate to fall under that as well. I was thinking maybe we could have an "ec-admin" command, similar to the current "admin" command.

Do you think a follow on ticket should be created for that? Also, would this tool fall under an "ec-admin" command even though it checks both internal and external compactions?

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: > @DomGarguilo I attempted to refactor some existing code so that it could be used by this utility. I did that in this [commit](https://github.com/keith-turner/accumulo/commit/b17492345b777b5a8fc267c64b3e02b2333b472b) that is in this [branch](https://github.com/keith-turner/accumulo/commits/DomGarguilo-checkCompactionConfig).

The current changes the changes you made.

milleruntime:
> Do you think a follow on ticket should be created for that? 

I may just do it to close #2386

> Also, would this tool fall under an "ec-admin" command even though it checks both internal and external compactions?

That is a good point. If a user can make use of "checkCompactionConfig" without using external compactions, then it should probably hang off of the main command, with other generic server utilities.

DomGarguilo:
I think these changes are finished for now and ready for review. Here is an example of the output:
<details open>
<summary>Input file</summary>

```
tserver.compaction.major.service.cs1.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner 
tserver.compaction.major.service.cs1.planner.opts.executors=\
[{"name":"small","type":"internal","maxSize":"16M","numThreads":8},\
{"name":"medium","type":"internal","maxSize":"128M","numThreads":4},\
{"name":"large","type":"internal","numThreads":2}]
tserver.compaction.major.service.cs2.planner=org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner 
tserver.compaction.major.service.cs2.planner.opts.executors=\
[{"name":"small","type":"internal","maxSize":"16M","numThreads":7},\
{"name":"medium","type":"internal","maxSize":"128M","numThreads":5},\
{"name":"large","type":"external","queue":"dcq1"}]
```

</details>  
<details open>
<summary>Output from tool</summary>

```
dgarguilo@thor:~/github/accumulo$ accumulo check-compaction-config ~/Desktop/example.properties 
2021-12-22T11:21:23,366 [conf.CheckCompactionConfig] INFO : Service id: cs1, planner class:org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner
2021-12-22T11:21:23,409 [conf.CheckCompactionConfig] INFO : Compaction service 'cs1' requested creation of thread pool 'i.cs1.small' with 8 threads.
2021-12-22T11:21:23,409 [conf.CheckCompactionConfig] INFO : Compaction service 'cs1' requested creation of thread pool 'i.cs1.large' with 2 threads.
2021-12-22T11:21:23,409 [conf.CheckCompactionConfig] INFO : Compaction service 'cs1' requested creation of thread pool 'i.cs1.medium' with 4 threads.
2021-12-22T11:21:23,409 [conf.CheckCompactionConfig] INFO : Service id: default, planner class:org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner
2021-12-22T11:21:23,411 [conf.CheckCompactionConfig] INFO : Compaction service 'default' requested creation of thread pool 'i.default.small' with 2 threads.
2021-12-22T11:21:23,411 [conf.CheckCompactionConfig] INFO : Compaction service 'default' requested creation of thread pool 'i.default.medium' with 2 threads.
2021-12-22T11:21:23,411 [conf.CheckCompactionConfig] INFO : Compaction service 'default' requested creation of thread pool 'i.default.large' with 2 threads.
2021-12-22T11:21:23,411 [conf.CheckCompactionConfig] INFO : Service id: cs2, planner class:org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner
2021-12-22T11:21:23,413 [conf.CheckCompactionConfig] INFO : Compaction service 'cs2' requested creation of thread pool 'i.cs2.medium' with 5 threads.
2021-12-22T11:21:23,413 [conf.CheckCompactionConfig] INFO : Compaction service 'cs2' requested creation of thread pool 'i.cs2.small' with 7 threads.
2021-12-22T11:21:23,413 [conf.CheckCompactionConfig] INFO : Compaction service 'cs2' requested with external execution queue 'e.dcq1'
2021-12-22T11:21:23,413 [conf.CheckCompactionConfig] INFO : Service id: meta, planner class:org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner
2021-12-22T11:21:23,414 [conf.CheckCompactionConfig] INFO : Compaction service 'meta' requested creation of thread pool 'i.meta.huge' with 2 threads.
2021-12-22T11:21:23,414 [conf.CheckCompactionConfig] INFO : Compaction service 'meta' requested creation of thread pool 'i.meta.small' with 2 threads.
2021-12-22T11:21:23,414 [conf.CheckCompactionConfig] INFO : Service id: root, planner class:org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner
2021-12-22T11:21:23,415 [conf.CheckCompactionConfig] INFO : Compaction service 'root' requested creation of thread pool 'i.root.small' with 1 threads.
2021-12-22T11:21:23,415 [conf.CheckCompactionConfig] INFO : Compaction service 'root' requested creation of thread pool 'i.root.huge' with 1 threads.
2021-12-22T11:21:23,415 [conf.CheckCompactionConfig] INFO : Properties file has passed all checks.

```

</details>

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
```diff
@@ -0,0 +1,186 @@
...
+        configs = Map.copyOf(configsCopy);
+      }
+    }
+
+    return configs;
+
+  }
+
+  public CompactionServicesConfig(AccumuloConfiguration aconf,
+      Consumer<String> deprecationWarningConsumer) {
```
milleruntime, line 116: Passing a consumer for the logger is nifty but I think this could be confusing for someone debugging looking at line numbers for a logger. Also, if you just passed the logger, you could use it for debug logging too.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
```diff
@@ -0,0 +1,186 @@
...
+      String[] tokens = suffix.split("\\.");
+      if (tokens.length == 4 && tokens[1].equals("planner") && tokens[2].equals("opts")) {
+        getOptions().computeIfAbsent(tokens[0], k -> new HashMap<>()).put(tokens[3], val);
+      } else if (tokens.length == 2 && tokens[1].equals("planner")) {
+        getPlanners().put(tokens[0], val);
+      } else if (tokens.length == 3 && tokens[1].equals("rate") && tokens[2].equals("limit")) {
+        var eprop = Property.getPropertyByKey(prop);
+        if (eprop == null || aconf.isPropertySet(eprop, true)
+            || !isDeprecatedThroughputSet(aconf)) {
+          getRateLimits().put(tokens[0], ConfigurationTypeHelper.getFixedMemoryAsBytes(val));
```
milleruntime, line 132: Should access the private maps (planners, rateLimits and options) directly here instead of the public methods (getOptions(), getplanners() and getRateLimits()). The methods are public so could be overridden and could change the behavior of the constructor. I don't think that is what you want since this code seems very specific to that one property.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
```diff
@@ -0,0 +1,186 @@
...
+
+import org.apache.accumulo.core.conf.AccumuloConfiguration;
+import org.apache.accumulo.core.conf.ConfigurationTypeHelper;
+import org.apache.accumulo.core.conf.Property;
+import org.apache.accumulo.core.spi.compaction.CompactionServiceId;
+import org.apache.accumulo.core.spi.compaction.DefaultCompactionPlanner;
+
+import com.google.common.collect.Sets;
+
+public class CompactionServicesConfig {
```
milleruntime, line 34: Since this class is specific to certain properties, it would be nice to note them in the javadoc. Examples of properties is always helpful too.

server/tserver/src/main/java/org/apache/accumulo/tserver/compactions/CompactionManager.java
```diff
@@ -91,144 +89,10 @@ public ExtCompInfo(KeyExtent extent, CompactionExecutorId executor) {
     }
   }
 
-  private class Config {
```
milleruntime, line 94: This is a nice change, breaking out this class and giving it a more proper class name.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
```diff
@@ -0,0 +1,186 @@
...
+          .filter(key -> key.startsWith(defaultServicePrefix)).map(Property::getPropertyByKey)
+          .anyMatch(prop -> prop == null || aconf.isPropertySet(prop, true));
+
+      if (defaultServicePropsSet) {
+
+        String warning = String.format(
+            "The deprecated property %s was set. Properties with the prefix %s "
+                + "were also set, which replace the deprecated properties. The deprecated "
+                + "property was therefore ignored.",
+            Property.TSERV_MAJC_MAXCONCURRENT.getKey(), defaultServicePrefix);
```
milleruntime, line 77: The performance of `String.format()` is not great. It seems like this could be called a lot in the case where a user stays with the default properties. I am not sure how often the CompactionManager code calls this though but either way, it doesn't seem like you really need the format since you can just concatenate the strings.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
```diff
@@ -0,0 +1,147 @@
...
+
+    Path path = Path.of(opts.filePath);
+    if (!path.toFile().exists())
+      throw new FileNotFoundException("File at given path could not be found");
+
+    AccumuloConfiguration config = SiteConfiguration.fromFile(path.toFile()).build();
+    var servicesConfig = new CompactionServicesConfig(config, log::warn);
+    ServiceEnvironment senv = createServiceEnvironment(config);
+
+    Set<String> defaultServices = Set.of("default", "meta", "root");
```
milleruntime, line 83: Could make these static final at the top of the class. Sometimes this makes hard coded strings easier to find.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
```diff
@@ -0,0 +1,147 @@
...
+    if (servicesConfig.getPlanners().keySet().equals(defaultServices)) {
+      throw new IllegalArgumentException("Only the default compaction services were created");
```
milleruntime, line 85: Could make this a warning and return. Throwing an error is fine, I am just not sure how useful showing the user a stack trace would be for this situation.

milleruntime (commented):
This is looking good. I just had a few comments

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionPlannerInitParams.java
```diff
@@ -0,0 +1,98 @@
...
+import org.apache.accumulo.core.conf.Property;
+import org.apache.accumulo.core.spi.common.ServiceEnvironment;
+import org.apache.accumulo.core.spi.compaction.CompactionExecutorId;
+import org.apache.accumulo.core.spi.compaction.CompactionPlanner;
+import org.apache.accumulo.core.spi.compaction.CompactionServiceId;
+import org.apache.accumulo.core.spi.compaction.ExecutorManager;
+
+import com.google.common.base.Preconditions;
+
+public class CompactionPlannerInitParams implements CompactionPlanner.InitParameters {
```
milleruntime, line 35: This is another nice well named class breakout.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
DomGarguilo: Yea good point. A consumer is used here so that we could ignore duplicate logs coming from CompationManager. See `CompactionManager.warnAboutDeprecation()`. I can't think of a way to preserve this functionality with just a logger being passed.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: Good idea. On second thought, I think a warning is a better idea too.

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
milleruntime: Ah OK I see now. That is a nice feature so its worth using the consumer to prevent spamming the logs.

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: Addressed in 76b4e70

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
DomGarguilo: Addressed in 76b4e70

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
DomGarguilo: Addressed in 76b4e70

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
DomGarguilo: Addressed in 76b4e70

server/base/src/main/java/org/apache/accumulo/server/conf/CheckCompactionConfig.java
```diff
@@ -0,0 +1,147 @@
...
+import com.beust.jcommander.Parameter;
+import com.google.auto.service.AutoService;
+
+@AutoService(KeywordExecutable.class)
+public class CheckCompactionConfig implements KeywordExecutable {
+
+  private final static Logger log = LoggerFactory.getLogger(CheckCompactionConfig.class);
+
+  static class Opts extends Help {
+    @Parameter(description = "<path/to/props/file>", required = true)
```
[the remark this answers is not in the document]
DomGarguilo: Addressed in 76b4e70

milleruntime (approved):
There are some nice general improvements to compaction code in this PR, in addition to creating the validation tool. Looks good!

Manno15:
:+1:

core/src/main/java/org/apache/accumulo/core/util/compaction/CompactionServicesConfig.java
DomGarguilo: Do you think something similar to the example input file I have in my comment above ([here](https://github.com/apache/accumulo/pull/2345#issuecomment-999719164)) would be a good example to include in the javadoc? And then just mention that the class is specific to compaction properties? @milleruntime

milleruntime: It might be a pain to try to put the json in the javadoc so not sure how feasable that would be. At the very least I think we could mention the property names that it is looking at.

DomGarguilo: Gave it a shot in bffe840 let me know what you think.

milleruntime: LGTM

Manno15:
:+1:

Manno15:
:+1:

PR #2354, opened 2021-11-10
#2353: Update property name output

alerman:


ctubbsii (changes requested):
This looks good to me, but there's a few other places in this class that does the same thing. Would you be willing to update the others as well?

alerman:
Updated to also change the other 2 places I saw that should be changed.

milleruntime (approved):

ctubbsii (approved):

PR #2358, opened 2021-11-16
Create page in Monitor for external compactions

### LINKED ISSUE #2290: Include some External Compaction info in Monitor
There is currently nothing reported about External Compactions in the Monitor

At the minimum, include the same info about External compactions that we currently provide for internal compactions (number running, queued). If possible it would be helpful to list the process ID or external compaction ID of the process performing the compactions. We may need to add a separate page since the information would likely be separate from what we currently provide.

milleruntime:
* Create multiple new classes for displaying 3 different tables of data
in the new external compaction page in the monitor
* Create 3 new ajax endpoints in ECResource
* Modify Compactor and ExternalCompactionUtil to return Optional for the
compaction coordinator instead of null
* Add check for compaction coordinator to Monitor.fetchData()
* New ExternalCompactionProgressIT for testing progress
* Closes #2290

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
@@ -137,7 +137,7 @@ public class ExternalCompactionProgressIT extends AccumuloClusterHarness {
             log.warn("{} did not progress. It went from {} to {}", ecid, previous.progress,
                 rci.progress);
           } else {
-          log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
+            log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
             if (rci.progress > 0 && rci.progress < 25)
               progressList.add(EC_PROGRESS.STARTED);
             else if (rci.progress > 25 && rci.progress < 50)
--- a/test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
@@ -128,23 +128,23 @@ public class ExternalCompactionProgressIT extends AccumuloClusterHarness {
     if (ecMap != null) {
       ecMap.forEach((ecid, ec) -> {
         // returns null if it's a new mapping
-        var rci = new RunningCompactorInfo(System.currentTimeMillis(), ecid, ec);
-        var previous = runningMap.put(ecid, rci);
-        if (previous == null) {
+        RunningCompactorInfo rci = new RunningCompactorInfo(System.currentTimeMillis(), ecid, ec);
+        RunningCompactorInfo previousRci = runningMap.put(ecid, rci);
+        if (previousRci == null) {
           log.debug("New ECID {} with inputFiles: {}", ecid, rci.inputFiles);
         } else {
-          if (rci.progress <= previous.progress) {
-            log.warn("{} did not progress. It went from {} to {}", ecid, previous.progress,
+          if (rci.progress <= previousRci.progress) {
+            log.warn("{} did not progress. It went from {} to {}", ecid, previousRci.progress,
                 rci.progress);
           } else {
-            log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
-            if (rci.progress > 0 && rci.progress < 25)
+            log.debug("{} progressed from {} to {}", ecid, previousRci.progress, rci.progress);
+            if (rci.progress > 0 && rci.progress <= 25)
               progressList.add(EC_PROGRESS.STARTED);
-            else if (rci.progress > 25 && rci.progress < 50)
+            else if (rci.progress > 25 && rci.progress <= 50)
               progressList.add(EC_PROGRESS.QUARTER);
-            else if (rci.progress > 50 && rci.progress < 75)
+            else if (rci.progress > 50 && rci.progress <= 75)
               progressList.add(EC_PROGRESS.HALF);
-            else if (rci.progress > 75 && rci.progress < 100)
+            else if (rci.progress > 75 && rci.progress <= 100)
               progressList.add(EC_PROGRESS.THREE_QUARTERS);
           }
           if (!rci.status.equals(TCompactionState.IN_PROGRESS.name())) {
--- a/test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
@@ -133,11 +133,11 @@ public class ExternalCompactionProgressIT extends AccumuloClusterHarness {
         if (previous == null) {
           log.debug("New ECID {} with inputFiles: {}", ecid, rci.inputFiles);
         } else {
-          log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
           if (rci.progress <= previous.progress) {
-            log.warn("Compaction did not progress. It went from {} to {}", previous.progress,
+            log.warn("{} did not progress. It went from {} to {}", ecid, previous.progress,
                 rci.progress);
           } else {
+          log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
             if (rci.progress > 0 && rci.progress < 25)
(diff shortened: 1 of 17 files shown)


test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
```diff
@@ -0,0 +1,166 @@
...
+   * Check running compaction progress.
+   */
+  private void checkRunning() throws TException {
+    var ecList = ExternalCompactionTestUtils.getRunningCompactions(getCluster().getServerContext());
+    var ecMap = ecList.getCompactions();
+    if (ecMap != null) {
+      ecMap.forEach((ecid, ec) -> {
+        // returns null if it's a new mapping
+        var rci = new RunningCompactorInfo(System.currentTimeMillis(), ecid, ec);
+        var previous = runningMap.put(ecid, rci);
```
DomGarguilo, line 132: This could potentially return null if the previously assigned value was null, meaning its not a new mapping. Not sure if that would ever happen though.

test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
```diff
@@ -0,0 +1,166 @@
...
+        if (previous == null) {
+          log.debug("New ECID {} with inputFiles: {}", ecid, rci.inputFiles);
+        } else {
+          log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
+          if (rci.progress <= previous.progress) {
+            log.warn("Compaction did not progress. It went from {} to {}", previous.progress,
+                rci.progress);
+          } else {
```
DomGarguilo, line 140:
```suggestion
          if (rci.progress <= previous.progress) {
            log.warn("{} did not progress. It went from {} to {}", ecid, previous.progress,
                rci.progress);
          } else {
          log.debug("{} progressed from {} to {}", ecid, previous.progress, rci.progress);
```
The log could be moved here to avoid redundant logging. It also could be helpful to include the ecid in the log.warn message.

test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
```diff
@@ -0,0 +1,166 @@
...
+            if (rci.progress > 0 && rci.progress < 25)
+              progressList.add(EC_PROGRESS.STARTED);
+            else if (rci.progress > 25 && rci.progress < 50)
+              progressList.add(EC_PROGRESS.QUARTER);
+            else if (rci.progress > 50 && rci.progress < 75)
+              progressList.add(EC_PROGRESS.HALF);
+            else if (rci.progress > 75 && rci.progress < 100)
+              progressList.add(EC_PROGRESS.THREE_QUARTERS);
```
DomGarguilo, line 148: what if progress is 0, 25, 50, 75 or 100?

test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionTestUtils.java
```diff
@@ -158,8 +160,19 @@ public static void writeData(AccumuloClient client, String table1)
     client.tableOperations().flush(table1);
   }
 
+  public static void writeData(AccumuloClient client, String table1)
+      throws MutationsRejectedException, TableNotFoundException, AccumuloException,
```
DomGarguilo, line 164:
```suggestion
      throws TableNotFoundException, AccumuloException,
```
Looks like this is covered under AccumuloException

test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionTestUtils.java
```diff
@@ -144,11 +146,11 @@ public static void createTable(AccumuloClient client, String tableName, String s
 
   }
 
-  public static void writeData(AccumuloClient client, String table1)
+  public static void writeData(AccumuloClient client, String table1, int rows)
       throws MutationsRejectedException, TableNotFoundException, AccumuloException,
```
DomGarguilo, line 150:
```suggestion
      throws TableNotFoundException, AccumuloException,
```
Looks like this is covered under AccumuloException

test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
milleruntime: There won't ever be an update for 100% or 0% since these states are represented by other actions instead of passing updates between the processes. I will add comparisons to catch the other values.

test/src/main/java/org/apache/accumulo/test/compaction/ExternalCompactionProgressIT.java
milleruntime: It shouldn't ever store a null value for the `RunningCompactorInfo` object. I replaced var with the actual type and renamed the variable to make it more clear.

milleruntime:
The data being populated in running compactions table could grow quite large and be inefficient when loading the table. I am going to do a follow on task to optimize the performance of fetching the data.

Manno15:
Does this PR align with issue #2286? I did notice in my testing that The Compaction Coordinator section has most of its column values right-aligned while the Compactors section has the column values left-aligned. Not a big deal but I wanted to point it out.

milleruntime:
> Does this PR align with issue #2286? I did notice in my testing that The Compaction Coordinator section has most of its column values right-aligned while the Compactors section has the column values left-aligned. Not a big deal but I wanted to point it out.

I will find a new Bootstrap component for the Coordinator instead of the one row table.

milleruntime:
@Manno15 I redid the coordinator info to be in a panel with badges

new-coordinator

Manno15:
@milleruntime I like the look of that. I am going to take another pass through the PR tomorrow.

Manno15:
@milleruntime For some reason the compaction coordinator panel did not populate any data on my end. Did you have to start the coordinator in a specific way or do any other configuring? 

I do really like the Running Compactions table and the info it provides.

Manno15:
I did notice something interesting, the value for 'Last Update' and 'Duration' inside the 'Running Compactions' table will be the same value while the progress bar shows 0%. This updates multiple times before the progress bar changes but those two values will still be equal to each other. Once, the progress bar changes (in my testing, it generally jumped to 78% or all the way to 98%) the values for those two columns will be their expected value. I.e. multiple seconds for duration and last update a relatively small number.

milleruntime:
> @milleruntime For some reason the compaction coordinator panel did not populate any data on my end. Did you have to start the coordinator in a specific way or do any other configuring? For reference (this was taken while compactions were going on and I do have coordinator logs): image
> 
> I do really like the Running Compactions table and the info it provides.

If you have brought up the page with previous updates then your browser could have the JS cached. Doing a CTRL+F5 should do a hard refresh.

Manno15:
> If you have brought up the page with previous updates then your browser could have the JS cached. Doing a CTRL+F5 should do a hard refresh.

Thanks, that worked!

Manno15 (approved):
These changes look really good. Could something similar to the 'running compactions' table be added to the active compactions page to cover internal compactions?

milleruntime:
> These changes look really good. Could something similar to the 'running compactions' table be added to the active compactions page to cover internal compactions?

There is some work that would need to be done to the internal plumbing but it might not be much work. @dlmarion had some ideas of how this could be done.

dlmarion:
@Manno15 The Compactor starts the FileCompactor in a separate thread, then in the main thread it periodically gets the stats and sends them to the Coordinator. The FileCompactor is the same object used in the TabletServers to perform major compactions. In theory, the TabletServer could periodically iterate over the major compaction threads that are in progress and do the same thing and send the stats to the Manager. See https://github.com/apache/accumulo/blob/main/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java#L682

dlmarion:
@milleruntime - Are you creating follow-on tickets for the recently completed compactions? If not, we might want to remove that map from the Coordinator. I think there was one other follow-on issue too, but I can't remember it right now.

milleruntime:
> @milleruntime - Are you creating follow-on tickets for the recently completed compactions? If not, we might want to remove that map from the Coordinator. I think there was one other follow-on issue too, but I can't remember it right now.

I wasn't planning on it since there were already concerns over the amount of data being displayed on the page. Maybe once I can improve the performance for a large cluster, that would be a good change.

PR #2367, opened 2021-11-22
Add support for LZ4 and BZip2 Codecs

trietopsoft:
Hadoop compression support for 'lz4' and 'bzip2' codecs.  Updates javadoc and normalizes comments for lz4, bzip2, and zstd.

Implements #2366

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/client/mapred/AccumuloFileOutputFormat.java
+++ b/core/src/main/java/org/apache/accumulo/core/client/mapred/AccumuloFileOutputFormat.java
@@ -67,7 +67,7 @@ public class AccumuloFileOutputFormat extends FileOutputFormat<Key,Value> {
    * @param job
    *          the Hadoop job instance to be configured
    * @param compressionType
-   *          one of "none", "gz", "lzo", or "snappy"
+   *          one of "none", "gz", "bzip2", "lzo", "lz4", "snappy", or "zstd"
    * @since 1.5.0
    */
   public static void setCompressionType(JobConf job, String compressionType) {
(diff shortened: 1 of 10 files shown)


EdColeman:
The git workflow build is failing because of file formatting issues.  You can run the maven build locally and it will run the formatter (as well as other checks)  To run the
same command as the workflow fast build you can run:
```
mvn clean verify package javadoc:jar -Psec-bugs -DskipITs=true -DskipTests=true```

If you get that to pass, you may want to run the unit tests at least once, commit the formatting changes and then push a new commit once the build is passing.

milleruntime (approved):
These changes look good but I haven't had a chance to test with hadoop. I noticed that you modified several classes which have been deprecated. Changes to deprecated classes are usually only made for critical bug fixes. But in this case, I think it is OK because otherwise it would create confusion to not have all the compression types listed.

trietopsoft:
@milleruntime thank you for reviewing.  The deprecated classes were updated to maintain parity with the codecs available in the system.  We understand these classes will no longer be supported but would like equivalent functionality during the transition.

dlmarion (approved):
LGTM

milleruntime:
I had to drop recent commits that are currently causing compactions to fail (see https://github.com/apache/accumulo/issues/2362) but I was able to test these changes in Uno. I was able to use the 2 new compression types with Hadoop 3.3.0 and inspected files using `accumulo rfile-info`. LGTM

milleruntime:
Thanks for the contribution @trietopsoft and welcome to the Apache Accumulo community. If you would like to be listed as a contributor on Accumulo's people page please make a website PR for people.md. https://github.com/apache/accumulo-website/edit/main/pages/people.md

If you intend to make more contributions, please consider subscribing to the dev list and introducing yourself. Also, you can use the dev list or slack to reach out if you have questions about anything. https://accumulo.apache.org/contact-us/

PR #2369, opened 2021-11-29
Fix incorrect scan range output in getsplits command

### LINKED ISSUE #2356: Incorrect scan range in Accumulo shell getsplits command.
I have experienced an issue using the getsplits command in the accumulo shell.  It occurs when the tableId of the table happens to be a single letter ("w" for example ) and there are other tables in the system that start with that letter ("w1", "w3" for example).  The result is the output of getsplits is displaying splits for other tables.  This has produced outputs with multiple default tablets for example and has been demonstrated in multiple installations.

Looking at the current getsplits code, 

https://github.com/apache/accumulo/blob/6cfb9180a0d3e5115922314ff2062e0706ef0795/shell/src/main/java/org/apache/accumulo/shell/commands/GetSplitsCommand.java#L72-L76

it seems to be scanning a range where the start and end of the range are: <tableId>, <tableId>+'<'.  For our example above this amounts to:
accumulo shell -u <user>
table <my table>
scan -b w -e w<

And this can be shown to pick up tablets from tables w1, w3 etc.

**Versions**
  Affected version(s) of this project: 1.9.2 but seems to be still in the code-base as of Nov 12 2021

**To Reproduce**
    1. Require to have a table with a single character tableId such as w or 3 and a table with a multiple character tableId that start with it, such as w1 or 33.
    2. accumulo shell -u \<user\>
    3. table \<table\>
    4. getsplits -v -o /tmp/splits.txt
    5. See error:  There may be observed splits for tables that are not the current table.

**Expected behavior**
Only splits for the current table should be displayed.

**Fix**
It has been suggested to add a line after line 73 above:
`start.append(new byte[] {';'}, 0, 1);
`

jmark99:
In the Accumulo shell, calling getsplits with the verbose option can result in incorret output. It occurs when the tableId of the table
happens to be a single character and there are other tables where the tableId starts with the same character. This results in the output of getsplits displaying splits for the other tables as well.

Closes #2356

### DIFF
--- a/test/src/main/java/org/apache/accumulo/test/ShellIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/ShellIT.java
@@ -520,7 +520,7 @@ public class ShellIT extends SharedMiniClusterBase {
   // This test addresses a bug (#2356) where if a table with a tableId of character length 1
   // exists and another table(s) exist starting with the same character but with a tableId of
   // length > 1, the verbose version of the getsplits command will return information from multiple
-  // tables when a table with lexigraphically earlier ID is queried.
+  // tables when a lexicographical ordered table with an earlier ID is queried.
   //
   // In order to test, enough tables need to be created until the required condition exists.
   // Since table ID counts increment using 1..9a..z, this test creates tables in groups of 36
--- a/test/src/main/java/org/apache/accumulo/test/ShellIT.java
+++ b/test/src/main/java/org/apache/accumulo/test/ShellIT.java
@@ -531,7 +531,7 @@ public class ShellIT extends SharedMiniClusterBase {
   //
   // Due to the number for createtable calls, this test will time out if a match is not found
   // within some number of operations. Therefore, if a match is not found within the creation
-  // of the first 360 or so tables, the test exits witn no results. In initial runs of the ITs
+  // of the first 360 or so tables, the test exits with no results. In initial runs of the ITs
   // this never occurred.
   @Test
   public void testGetSplitsScanRange() throws Exception {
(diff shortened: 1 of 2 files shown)


test/src/main/java/org/apache/accumulo/test/ShellIT.java
```diff
@@ -515,4 +516,77 @@ public void setIterTest() throws IOException {
 
     exec("deletetable t -f", true, "Table: [t] has been deleted");
   }
+
+  // This test addresses a bug (#2356) where if a table with a tableId of character length 1
+  // exists and another table(s) exist starting with the same character but with a tableId of
+  // length > 1, the verbose version of the getsplits command will return information from multiple
+  // tables when a table with lexigraphically earlier ID is queried.
```
milleruntime, line 523:
```suggestion
  // tables when a lexicographical ordered table with an earlier ID is queried.
```

test/src/main/java/org/apache/accumulo/test/ShellIT.java
```diff
@@ -515,4 +516,77 @@ public void setIterTest() throws IOException {
...
+  // In order to test, enough tables need to be created until the required condition exists.
+  // Since table ID counts increment using 1..9a..z, this test creates tables in groups of 36
+  // until a second table meeting the criteria is created.
+  // It then adds splits to the single digit ID table and one of the others. It performs a
+  // "getsplits -v" command on the single digit ID table and verifies that data from the other
+  // table is not present.
+  //
+  // Due to the number for createtable calls, this test will time out if a match is not found
+  // within some number of operations. Therefore, if a match is not found within the creation
+  // of the first 360 or so tables, the test exits witn no results. In initial runs of the ITs
```
milleruntime, line 534:
```suggestion
  // of the first 360 or so tables, the test exits with no results. In initial runs of the ITs
```

milleruntime (approved):
Looks good. Just found a couple minor typos. I like the explanation in the test, it definitely helped clarify the bug.

PR #2370, opened 2021-11-29
Fix incorrect scan range output in getsplits command

### LINKED ISSUE #2356: Incorrect scan range in Accumulo shell getsplits command.

jmark99:
In the Accumulo shell, calling getsplits with the verbose option can result in incorret output. It occurs when the tableId of the table
happens to be a single character and there are other tables where the tableId starts with the same character. This results in the output of getsplits displaying splits for the other tables as well.

Closes #2356

### DIFF
--- a/shell/src/main/java/org/apache/accumulo/shell/commands/GetSplitsCommand.java
+++ b/shell/src/main/java/org/apache/accumulo/shell/commands/GetSplitsCommand.java
@@ -81,6 +81,8 @@ public class GetSplitsCommand extends Command {
         final Text start =
             new Text(shellState.getConnector().tableOperations().tableIdMap().get(tableName));
         final Text end = new Text(start);
+        // do not append the ';' until the end value above is assigned.
+        start.append(new byte[] {';'}, 0, 1);
         end.append(new byte[] {'<'}, 0, 1);
         scanner.setRange(new Range(start, end));
         for (Iterator<Entry<Key,Value>> iterator = scanner.iterator(); iterator.hasNext();) {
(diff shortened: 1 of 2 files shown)


EdColeman:
Questions more than a comment.

Did you consider mocking and using a unit test rather than (or in conjunction with) the IT test?
At what point should we consider breaking up the shell server IT test?  That test was lengthy as it was - If it was split up, would there be more parallelism?

Manno15:
This PR seems to be a copy of #2369

jmark99:
@Manno15 one is for 2.x and the other for 1.x

Manno15:
My apologies. I didn't look that cloesly.

jmark99:
@EdColeman I must admit I did not consider mocking. The test does seems to be a bit of an overkill for a one line bug fix, but I wanted to verify it using the actual command calls. I haven't had a lot of experience using mocking in the past so that probably played a part in me not going that route. 

As for breaking up the IT, we could put in an issue if someone wanted to take on that task.

milleruntime:
The fix seems small enough that you didn't need both PRs but either way, I approved the other one.

PR #2372, opened 2021-11-30
Exception and cancellation changes in Compactor process

### LINKED ISSUE #2350: External compaction error when table is deleted
I saw a few errors that I did not expect to see after deleting a table that had active running external compactions. Here is what gets printed in the compactor log:
<pre>
2021-11-08T12:20:31,706 [compaction.FileCompactor] ERROR: File does not exist: /accumulo/tables/2/t-000002w/C0000115.rf_tmp (inode 17654) Holder DFSClient_NONMAPREDUCE_55427
0523_18 does not have any open files.
        at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.checkLease(FSNamesystem.java:3050)
        at org.apache.hadoop.hdfs.server.namenode.FSDirWriteFileOp.completeFileInternal(FSDirWriteFileOp.java:704)
        at org.apache.hadoop.hdfs.server.namenode.FSDirWriteFileOp.completeFile(FSDirWriteFileOp.java:690)
        at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.completeFile(FSNamesystem.java:3094)
        at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.complete(NameNodeRpcServer.java:963)
        at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.complete(ClientNamenodeProtocolServerSideTranslatorPB.java:639)
        at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
        at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:532)
        at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1070)
        at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1020)
        at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:948)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
        at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1845)
        at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2952)
</pre>
<pre>
org.apache.hadoop.ipc.RemoteException: File does not exist: /accumulo/tables/2/t-000002w/C0000115.rf_tmp (inode 17654) Holder DFSClient_NONMAPREDUCE_554270523_18 does not ha
ve any open files.
        at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.checkLease(FSNamesystem.java:3050)
        at org.apache.hadoop.hdfs.server.namenode.FSDirWriteFileOp.completeFileInternal(FSDirWriteFileOp.java:704)
        at org.apache.hadoop.hdfs.server.namenode.FSDirWriteFileOp.completeFile(FSDirWriteFileOp.java:690)
        at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.completeFile(FSNamesystem.java:3094)
        at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.complete(NameNodeRpcServer.java:963)
        at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.complete(ClientNamenodeProtocolServerSideTranslatorPB.java:639)
        at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
        at org.apache.hadoop.ipc.ProtobufRpcEngine2$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine2.java:532)
        at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1070)
        at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:1020)
        at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:948)
        at java.base/java.security.AccessController.doPrivileged(Native Method)
        at java.base/javax.security.auth.Subject.doAs(Subject.java:423)
        at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1845)
        at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2952)

at org.apache.hadoop.ipc.Client.getRpcResponse(Client.java:1562) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.ipc.Client.call(Client.java:1508) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.ipc.Client.call(Client.java:1405) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.ipc.ProtobufRpcEngine2$Invoker.invoke(ProtobufRpcEngine2.java:234) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.ipc.ProtobufRpcEngine2$Invoker.invoke(ProtobufRpcEngine2.java:119) ~[hadoop-client-api-3.3.0.jar:?]
        at com.sun.proxy.$Proxy34.complete(Unknown Source) ~[?:?]
        at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolTranslatorPB.complete(ClientNamenodeProtocolTranslatorPB.java:570) ~[hadoop-client-api-3.3.0.jar:?]
        at jdk.internal.reflect.GeneratedMethodAccessor3.invoke(Unknown Source) ~[?:?]
        at jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[?:?]
        at java.lang.reflect.Method.invoke(Method.java:566) ~[?:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler.invokeMethod(RetryInvocationHandler.java:422) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler$Call.invokeMethod(RetryInvocationHandler.java:165) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler$Call.invoke(RetryInvocationHandler.java:157) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler$Call.invokeOnce(RetryInvocationHandler.java:95) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.io.retry.RetryInvocationHandler.invoke(RetryInvocationHandler.java:359) ~[hadoop-client-api-3.3.0.jar:?]
        at com.sun.proxy.$Proxy35.complete(Unknown Source) ~[?:?]
        at org.apache.hadoop.hdfs.DFSOutputStream.completeFile(DFSOutputStream.java:957) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.hdfs.DFSOutputStream.completeFile(DFSOutputStream.java:914) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.hdfs.DFSOutputStream.closeImpl(DFSOutputStream.java:897) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.hdfs.DFSOutputStream.close(DFSOutputStream.java:852) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.fs.FSDataOutputStream$PositionCache.close(FSDataOutputStream.java:72) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.hadoop.fs.FSDataOutputStream.close(FSDataOutputStream.java:101) ~[hadoop-client-api-3.3.0.jar:?]
        at org.apache.accumulo.core.file.streams.RateLimitedOutputStream.close(RateLimitedOutputStream.java:54) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.core.file.rfile.bcfile.BCFile$Writer.close(BCFile.java:369) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.core.file.rfile.RFile$Writer.close(RFile.java:635) ~[accumulo-core-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.server.compaction.FileCompactor.call(FileCompactor.java:236) ~[accumulo-server-base-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
        at org.apache.accumulo.compactor.Compactor$6.run(Compactor.java:553) ~[accumulo-compactor-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
</pre>
<pre>
2021-11-08T12:20:31,706 [threads.AccumuloUncaughtExceptionHandler] ERROR: Caught an Exception in Thread[Compaction job for tablet TKeyExtent(table:32, endRow:31 33 33 33 33 33 33 33 33 33 33 33 33 33 33 35, prevEndRow:30 63 63 63 63 63 63 63 63 63 63 63 63 63 63 65),5,main]. Thread is dead.
java.lang.RuntimeException: Compaction failed
        at org.apache.accumulo.compactor.Compactor$6.run(Compactor.java:568) ~[accumulo-compactor-2.1.0-SNAPSHOT.jar:2.1.0-SNAPSHOT]
</pre>

Right before the errors this is what the same compactor was reporting in its log:
<pre>
2021-11-08T12:20:30,189 [compactor.Compactor] INFO : Starting up compaction runnable for job: TExternalCompactionJob(externalCompactionId:ECID:cd145bea-6d06-45cc-b2a2-5a9575
ab3e42, extent:TKeyExtent(table:32, endRow:31 33 33 33 33 33 33 33 33 33 33 33 33 33 33 35, prevEndRow:30 63 63 63 63 63 63 63 63 63 63 63 63 63 63 65), files:[InputFile(met
adataFileEntry:hdfs://localhost:8020/accumulo/tables/2/t-000002w/F00000ym.rf, size:2306065, entries:61204, timestamp:-1), InputFile(metadataFileEntry:hdfs://localhost:8020/a
ccumulo/tables/2/t-000002w/F00000z8.rf, size:2194819, entries:58318, timestamp:-1), InputFile(metadataFileEntry:hdfs://localhost:8020/accumulo/tables/2/t-000002w/F00000zu.rf
, size:2325338, entries:61640, timestamp:-1), InputFile(metadataFileEntry:hdfs://localhost:8020/accumulo/tables/2/t-000002w/F000010l.rf, size:2322428, entries:61581, timesta
mp:-1)], iteratorSettings:IteratorConfig(iterators:[]), outputFile:hdfs://localhost:8020/accumulo/tables/2/t-000002w/C0000115.rf_tmp, propagateDeletes:true, kind:SYSTEM, use
rCompactionId:0, overrides:{})
2021-11-08T12:20:30,191 [compactor.Compactor] DEBUG: Progress checks will occur every 1 seconds
2021-11-08T12:20:31,191 [compactor.Compactor] DEBUG: Updating coordinator with compaction progress: Compaction in progress, read 172032 of 242743 input entries ( 70.87001 % 
), written 172032 entries.
</pre>

dlmarion:
The Thread in the Compactor that runs the FileCompactor was throwing an
exception as part of the error handling routine. It's not necessary for
that to occur to convey the exception information to the main thread,
the exception information was already being conveyed. This will reduce
the error logging in the Compactor log regarding the uncaught exception
being thrown and handled by the AccumuloUncaughtExceptionHandler.

Another background thread checks every 5 seconds to determine whether
or not the current compaction should be canceled. Call this method
when the compaction completes in case something happens that would
cause the compaction to be canceled and it has not been noticed yet.

Closes #2350

### DIFF
--- a/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
+++ b/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
@@ -711,10 +711,11 @@ public class Compactor extends AbstractServer implements CompactorService.Iface 
           compactionThread.join();
           LOG.trace("Compaction thread finished.");
 
-   if(err.get() != null){
-           //maybe the error occured because the table was deleted or something like that, so force a cancel check to possibly reduce noise in the logs
-          checkIfCanceled();
-   }
+          if (err.get() != null) {
+            // maybe the error occured because the table was deleted or something like that, so
+            // force a cancel check to possibly reduce noise in the logs
+            checkIfCanceled();
+          }
 
           if (compactionThread.isInterrupted() || JOB_HOLDER.isCancelled()
               || ((err.get() != null && err.get().getClass().equals(InterruptedException.class)))) {
--- a/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
+++ b/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
@@ -711,7 +711,10 @@ public class Compactor extends AbstractServer implements CompactorService.Iface 
           compactionThread.join();
           LOG.trace("Compaction thread finished.");
 
+   if(err.get() != null){
+           //maybe the error occured because the table was deleted or something like that, so force a cancel check to possibly reduce noise in the logs
           checkIfCanceled();
+   }
 
           if (compactionThread.isInterrupted() || JOB_HOLDER.isCancelled()
               || ((err.get() != null && err.get().getClass().equals(InterruptedException.class)))) {
--- a/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
+++ b/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
@@ -184,7 +184,7 @@ public class Compactor extends AbstractServer implements CompactorService.Iface 
         TimeUnit.MILLISECONDS);
   }
 
-  protected synchronized void checkIfCanceled() {
+  protected void checkIfCanceled() {
     TExternalCompactionJob job = JOB_HOLDER.getJob();
     if (job != null) {
       try {
--- a/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
+++ b/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
@@ -184,7 +184,7 @@ public class Compactor extends AbstractServer implements CompactorService.Iface 
         TimeUnit.MILLISECONDS);
   }
 
-  protected void checkIfCanceled() {
+  protected synchronized void checkIfCanceled() {
     TExternalCompactionJob job = JOB_HOLDER.getJob();
     if (job != null) {
       try {
@@ -566,7 +566,6 @@ public class Compactor extends AbstractServer implements CompactorService.Iface 
         } catch (Exception e) {
           LOG.error("Compaction failed", e);
           err.set(e);
-          throw new RuntimeException("Compaction failed", e);
         } finally {
           stopped.countDown();
           Preconditions.checkState(compactionRunning.compareAndSet(true, false));
@@ -711,6 +710,8 @@ public class Compactor extends AbstractServer implements CompactorService.Iface 
           }
           compactionThread.join();
           LOG.trace("Compaction thread finished.");
+
+          checkIfCanceled();
 
           if (compactionThread.isInterrupted() || JOB_HOLDER.isCancelled()
               || ((err.get() != null && err.get().getClass().equals(InterruptedException.class)))) {
--- a/test/src/main/java/org/apache/accumulo/test/compaction/ExternalDoNothingCompactor.java
+++ b/test/src/main/java/org/apache/accumulo/test/compaction/ExternalDoNothingCompactor.java
@@ -53,6 +53,9 @@ public class ExternalDoNothingCompactor extends Compactor implements Iface {
       LongAdder totalInputBytes, CountDownLatch started, CountDownLatch stopped,
       AtomicReference<Throwable> err) {
 
+    // Set this to true so that only 1 external compaction is run
+    this.shutdown = true;
+
     return new Runnable() {
       @Override
       public void run() {
@@ -76,7 +79,6 @@ public class ExternalDoNothingCompactor extends Compactor implements Iface {
         } catch (Exception e) {
           LOG.error("Compaction failed", e);
           err.set(e);
-          throw new RuntimeException("Compaction failed", e);
         } finally {
           stopped.countDown();
         }
--- a/server/compactor/src/test/java/org/apache/accumulo/compactor/CompactorTest.java
+++ b/server/compactor/src/test/java/org/apache/accumulo/compactor/CompactorTest.java
@@ -220,6 +220,9 @@ public class CompactorTest {
     }
 
     @Override
+    protected synchronized void checkIfCanceled() {}
+
+    @Override
     protected Runnable createCompactionJob(TExternalCompactionJob job, LongAdder totalInputEntries,
         LongAdder totalInputBytes, CountDownLatch started, CountDownLatch stopped,
         AtomicReference<Throwable> err) {
@@ -337,6 +340,7 @@ public class CompactorTest {
     EasyMock.expect(conf.getTimeInMillis(Property.INSTANCE_ZK_TIMEOUT)).andReturn(86400000L);
 
     ServerContext context = PowerMock.createNiceMock(ServerContext.class);
+    EasyMock.expect(context.getConfiguration()).andReturn(conf);
     ZooReaderWriter zrw = PowerMock.createNiceMock(ZooReaderWriter.class);
     ZooKeeper zk = PowerMock.createNiceMock(ZooKeeper.class);
     EasyMock.expect(context.getZooReaderWriter()).andReturn(zrw).anyTimes();
@@ -390,6 +394,7 @@ public class CompactorTest {
     EasyMock.expect(conf.getTimeInMillis(Property.INSTANCE_ZK_TIMEOUT)).andReturn(86400000L);
 
     ServerContext context = PowerMock.createNiceMock(ServerContext.class);
+    EasyMock.expect(context.getConfiguration()).andReturn(conf);
     ZooReaderWriter zrw = PowerMock.createNiceMock(ZooReaderWriter.class);
     ZooKeeper zk = PowerMock.createNiceMock(ZooKeeper.class);
     EasyMock.expect(context.getZooReaderWriter()).andReturn(zrw).anyTimes();
@@ -443,6 +448,7 @@ public class CompactorTest {
     EasyMock.expect(conf.getTimeInMillis(Property.INSTANCE_ZK_TIMEOUT)).andReturn(86400000L);
 
     ServerContext context = PowerMock.createNiceMock(ServerContext.class);
+    EasyMock.expect(context.getConfiguration()).andReturn(conf);
     ZooReaderWriter zrw = PowerMock.createNiceMock(ZooReaderWriter.class);
     ZooKeeper zk = PowerMock.createNiceMock(ZooKeeper.class);
     EasyMock.expect(context.getZooReaderWriter()).andReturn(zrw).anyTimes();


test/src/main/java/org/apache/accumulo/test/compaction/ExternalDoNothingCompactor.java
```diff
@@ -53,6 +53,9 @@ protected Runnable createCompactionJob(TExternalCompactionJob job, LongAdder tot
       LongAdder totalInputBytes, CountDownLatch started, CountDownLatch stopped,
       AtomicReference<Throwable> err) {
 
+    // Set this to true so that only 1 external compaction is run
+    this.shutdown = true;
```
milleruntime, line 57: Is this the best place to set shutdown to true? Would it be better to set this right before `throw new CompactionCanceledException();`?

dlmarion: This method is called [here](https://github.com/apache/accumulo/blob/main/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java#L670) to create a Thread, then the Thread is started a few lines later. The shutdown variable is checked in the enclosing while loop, so I don't think it matters in this case.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -184,7 +184,7 @@ protected void startCancelChecker(ScheduledThreadPoolExecutor schedExecutor,
         TimeUnit.MILLISECONDS);
   }
 
-  protected void checkIfCanceled() {
+  protected synchronized void checkIfCanceled() {
```
keith-turner, line 187: What does the addition of synchonized here defend against?   There are reads from metadata table and ZK in this method, so the sync could possibly cause problems.

dlmarion: it's being called by a background thread every 5 seconds and from the main thread when the compaction is completed. I was just trying to serialize the calls.

milleruntime: I noticed the `cancel()` method call on `CompactionJobHolder` is synchronized, so maybe that is enough. Could add another check to this method to see if the compactionThread is already interrupted so we don't keep calling interrupt while waiting for the thread to die.

keith-turner: > I noticed the cancel() method call on CompactionJobHolder is synchronized, so maybe that is enough.

I think the existing sync in CompactionJobHolder will prevent any race conditions from happening.

keith-turner: > Could add another check to this method to see if the compactionThread is already interrupted so we don't keep calling interrupt while waiting for the thread to die.

It may be useful to keep interrupting a thread because some code may ignore interrupts w/o throwing an exception.

dlmarion: I have removed the synchronized modifier from this method.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -712,6 +711,8 @@ public void run() {
           compactionThread.join();
           LOG.trace("Compaction thread finished.");
 
+          checkIfCanceled();
```
keith-turner, line 714: The intent of doing periodic cancelation checks is to stop long running external compactions that are canceled.  If lots of quick external compaction (like 30 s) run, then a cancelation check will now be done for each of those where it was not before adding this   During the commit process atomic checks will be done, so doing the check here for every compaction does not offer a benefit over that. The compaction could still be canceled after this check before the atomic checks are done.  So maybe checking here places additional load on the metadata table and ZK w/o a clear benefit.

Thinking about only checking long running compactions, we could also make the checkIfCanceled method ignore young compactions. We currently check every 5 mins, if  a compaction is running and its age is <5min we could return and not do the check until a later call when the age is > 5min (or the check period).

dlmarion: I think this was an issue with short-lived compactions. IIRC, the original issue was that @milleruntime deleted the ci table during some testing and a bunch of errors were logged and the compaction was not canceled (at least not in coordinator log).

keith-turner: Would something like the following achieve the goal of less noisy logs while not always doing the check?

```suggestion
   if(err.get() != null){
           //maybe the error occured because the table was deleted or something like that, so force a cancel check
          checkIfCanceled();
   }
```

keith-turner (approved):

PR #2381, opened 2021-12-10
Refactor listSplits operation when using maxSplits

### LINKED ISSUE #2371: listSplits table operation with a max num splits may have a bug.
In the following code there is a while loop, however in the while loop the variable `i` does not change.   If the while loop will execute more than once then it will add the same split multiple times which seems like a bug.

https://github.com/apache/accumulo/blob/45a4a93d2926760d6a916ae2da1d7e668ec4dcb1/core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java#L706-L710

Not completely sure if the while loop could execute more than once, needs further investigation.  If the outer variable `r` could be greater than 1, then it seems like the while loop could execute more than once. 

If the while loop could never execute more than once then maybe it could be changed to an if stmt.

jmark99:
Refactored listSplits method in TableOperationsImpl. This change affects the listSplits command which takes maxSplits as an option.

* Renamed variable names to enhance readability.
* Added documentation for the method.
* Replaced while-loop with if-loop after determining while-loop was only run at most once each time.
* Created IT test for method in ShellIT class.

Closes #2371

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
@@ -733,7 +733,7 @@ public class TableOperationsImpl extends TableOperationsHelper {
     double selectionTrigger = 0.0;
 
     for (Text existingSplit : existingSplits) {
-      if (selectedSoFar >= maxSplits) {
+      if (splitsSubset.size() >= maxSplits) {
         break;
       }
       selectionTrigger += stepSize;
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
@@ -725,7 +725,6 @@ public class TableOperationsImpl extends TableOperationsHelper {
     // When the number of maxSplits requested is less than the number of existing splits, the
     // following code populates the splitsSubset list 'evenly' from the existing splits
     ArrayList<Text> splitsSubset = new ArrayList<>(maxSplits);
-    int selectedSoFar = 0;
     final int SELECTION_THRESHOLD = 1;
 
     // stepSize can never be greater than 1 due to the if-loop check above.
@@ -739,7 +738,6 @@ public class TableOperationsImpl extends TableOperationsHelper {
       selectionTrigger += stepSize;
       if (selectionTrigger > SELECTION_THRESHOLD) {
         splitsSubset.add(existingSplit);
-        selectedSoFar++;
         selectionTrigger -= 1;
       }
     }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
```diff
@@ -689,28 +689,61 @@ public void deleteRows(String tableName, Text start, Text end)
...
+    // following code populates the splitsSubset list 'evenly' from the existing splits
+    ArrayList<Text> splitsSubset = new ArrayList<>(maxSplits);
+    int selectedSoFar = 0;
+    final int SELECTION_THRESHOLD = 1;
+
+    // stepSize can never be greater than 1 due to the if-loop check above.
+    final double stepSize = (maxSplits + 1) / (double) (existingSplits.size());
+    double selectionTrigger = 0.0;
+
+    for (Text exitingSplit : existingSplits) {
```
keith-turner, line 735:
```suggestion
    for (Text existingSplit : existingSplits) {
```

core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
```diff
@@ -689,28 +689,61 @@ public void deleteRows(String tableName, Text start, Text end)
...
+      if (selectedSoFar >= maxSplits) {
+        break;
+      }
+      selectionTrigger += stepSize;
+      if (selectionTrigger > SELECTION_THRESHOLD) {
+        splitsSubset.add(exitingSplit);
```
keith-turner, line 741:
```suggestion
        splitsSubset.add(existingSplit);
```

core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
```diff
@@ -689,28 +689,61 @@ public void deleteRows(String tableName, Text start, Text end)
...
+    ArrayList<Text> splitsSubset = new ArrayList<>(maxSplits);
+    int selectedSoFar = 0;
+    final int SELECTION_THRESHOLD = 1;
+
+    // stepSize can never be greater than 1 due to the if-loop check above.
+    final double stepSize = (maxSplits + 1) / (double) (existingSplits.size());
+    double selectionTrigger = 0.0;
+
+    for (Text exitingSplit : existingSplits) {
+      if (selectedSoFar >= maxSplits) {
```
keith-turner, line 736: May be able to do the following and drop the selectedSoFar variable.

```suggestion
      if (splitsSubset.size() >= maxSplits) {
```

keith-turner (approved):
This looks good, the new variable names and test are nice.  Made a few suggestions.

core/src/main/java/org/apache/accumulo/core/clientImpl/TableOperationsImpl.java
```diff
@@ -689,28 +689,59 @@ public void deleteRows(String tableName, Text start, Text end)
...
+      return existingSplits;
     }
 
-    return subset;
+    // When the number of maxSplits requested is less than the number of existing splits, the
+    // following code populates the splitsSubset list 'evenly' from the existing splits
+    ArrayList<Text> splitsSubset = new ArrayList<>(maxSplits);
+    final int SELECTION_THRESHOLD = 1;
+
+    // stepSize can never be greater than 1 due to the if-loop check above.
```
milleruntime, line 730: I think you meant if statement or code block.

PR #2392, opened 2021-12-17
Fix coordinator check in Monitor

milleruntime:
* Monitor would stop checking for the coordinator once it found it
* This fix will detect if the coordinator moves or is no longer available

### DIFF
--- a/server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
+++ b/server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java
@@ -171,6 +171,7 @@ public class Monitor extends AbstractServer implements HighlyAvailableService {
   private Exception problemException;
   private GCStatus gcStatus;
   private Optional<HostAndPort> coordinatorHost = Optional.empty();
+  private long coordinatorCheckNanos = 0L;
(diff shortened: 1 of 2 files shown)


dlmarion:
FWIW, the Compactor uses [this](https://github.com/apache/accumulo/blob/e0853b7715f9712ad05168b727ebc916a7aeb200/server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java#L476) to get the Coordinator connection. This has the downside of calling ZK every time, which I think might be cached, but should pick up the case that the Coordinator is down (throws a Transport Exception) or if it moves.

milleruntime:
If the Monitor can no longer get the connection to the coordinator it will through an exception here:
https://github.com/apache/accumulo/blob/c19fc7c62bc4458e2530db18125352fd04680731/server/monitor/src/main/java/org/apache/accumulo/monitor/Monitor.java#L676-L682

This PR is mostly for the page to detect when the Coordinator can't be found or moved. The page itself will throw Datatables pop-up errors with a message that is not very helpful. With this change you can still see those errors before the next interval is reached but after a minute the monitor will detect that its gone and show the "Compaction Coordinator Not Running" message on the page.

PR #2393, opened 2021-12-17
Make test scripts clean up duplicate classes

milleruntime:
* The test scripts in start create multiple classes with the same name
that can cause problems for IDEs. The classes are only used to create
jars for the VFS tests so just drop the intermediate class files when done
creating the jars.


milleruntime:
This can be fixed in 1.10 as well.

EdColeman (approved):

ctubbsii (approved):

milleruntime:
Merged a separate branch into 1.10 first so I could merge it forward.

PR #2394, opened 2021-12-17
fixes #2373 support reading multiple metadata levels in Ample

### LINKED ISSUE #2373: Push tablet partitioning based on DataLevel to Ample
In  #2320 we partition tablets based on their DataLevel into two groups: one set who's DataLevel is USER and another containing the remaining tablets (those who's DataLevel is ROOT or METADATA). This needs to happen so we can use `.readTablets()` on the DataLevel.USER tablets in order to reduce RPCs.

This partitioning logic should be pushed to ample.

EDIT: this will need to wait until #2320 is merged.

keith-turner:


DomGarguilo:
These changes look good from what I can tell. Have you been able to verify that this works as intended? I can't think of a good way to test these changes.

Did you consider using a MuliMap here? It looks like it might make for a more complicated stream so not sure that its the best option but thought I would mention it just in case.
https://github.com/apache/accumulo/blob/e9fc27ce04138189057e495ceb3e238bae69e8b6/core/src/main/java/org/apache/accumulo/core/metadata/schema/TabletsMetadata.java#L134-L135

keith-turner:
> Did you consider using a MuliMap here? It looks like it might make for a more complicated stream so not sure that its the best option but thought I would mention it just in case.

@DomGarguilo  no I did not consider that.  I used the java stream library to group the data that code produced a Map of List.  So for that case I would stick w/ whatever the stream lib produces when grouping data.

PR #2396, opened 2021-12-20
Add metric for number of open files for scans

### LINKED ISSUE #2192: Add metric for files open during scans
Issue #2174 looked at adding the number of files to the listscans command.  Because the number of files opened is not currently available, the issue was closed as not viable at this time. This may be useful as a metric. 

Examine current scan and other metrics to see if there are additions / changes that would help in performance monitoring and troubleshooting. It is hard to determine what is achievable vs. how useful additional metrics are to end-users and in what circumstances. A preliminary task might be to catalog existing metrics and see if there are any significant gaps on what is being reported vs what would be useful.

If clear improvements are not needed, then leave current metrics as is.

**Additional context**

dlmarion:
Closes #2192

### DIFF
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
@@ -48,7 +48,6 @@ public class TabletServerScanMetrics implements MetricsProducer {
   }
 
   public void incrementOpenFiles(int delta) {
-    Math.max(0, delta);
     openFiles.addAndGet(Math.max(0, delta));
   }
 
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
@@ -52,7 +52,7 @@ public class TabletServerScanMetrics implements MetricsProducer {
   }
 
   public void decrementOpenFiles(int numOpenFiles) {
-    openFiles.addAndGet(numOpenFiles > 0 ? numOpenFiles : numOpenFiles * -1);
+    openFiles.addAndGet(numOpenFiles < 0 ? numOpenFiles : numOpenFiles * -1);
   }
 
   @Override
--- a/server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
+++ b/server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
@@ -19,15 +19,18 @@
 package org.apache.accumulo.tserver.metrics;
 
 import java.time.Duration;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.accumulo.core.metrics.MetricsProducer;
 
 import io.micrometer.core.instrument.DistributionSummary;
+import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Timer;
 
 public class TabletServerScanMetrics implements MetricsProducer {
 
+  private final AtomicInteger openFiles = new AtomicInteger(0);
   private Timer scans;
   private DistributionSummary resultsPerScan;
   private DistributionSummary yields;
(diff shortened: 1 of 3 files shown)


milleruntime (approved):

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
```diff
@@ -44,8 +47,18 @@ public void addYield(long value) {
     yields.record(value);
   }
 
+  public void incrementOpenFiles(int numOpenFiles) {
+    openFiles.addAndGet(numOpenFiles);
+  }
+
+  public void decrementOpenFiles(int numOpenFiles) {
```
DomGarguilo, line 54: In the cases where this is used, will the value of `openFiles` ever actually be decremented? It looks like `incrementOpenFiles()` and `decrementOpenFiles()` will both do the same thing at the moment. Is the idea that we always pass a negative number to `decrementOpenFiles()`?

dlmarion: LOL - yes, in my mind, that's what was happening. I should probably multiple the argument by -1. Good catch.

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
```diff
@@ -52,7 +52,7 @@ public void incrementOpenFiles(int numOpenFiles) {
   }
 
   public void decrementOpenFiles(int numOpenFiles) {
-    openFiles.addAndGet(numOpenFiles);
+    openFiles.addAndGet(numOpenFiles > 0 ? numOpenFiles : numOpenFiles * -1);
```
DomGarguilo, line 55: So with this change, if the number passed to `decrementOpenFiles()` is positive it will be added to `openFiles` else if its negative, it will be made positive (multiplied by -1) and added to `openFiles`?

dlmarion: What you are saying is that I'm doing a really bad job of multi-tasking today. :-)

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
milleruntime: Good catch. I don't think its possible that the fileManager will return a negative number though since it is just giving the size of the ArrayList. You could put assert checks at the beginning of each method to make sure the parameter is positive, then always multiplying by -1 for the decrementing.

milleruntime: Otherwise, the parameter name is odd with the way you have it now since you can't have negative "numOpenFiles". If you want to allow passing a negative parameter you could name it "deltaOpenFiles" or something similar since the parameter to `addAndGet` is called "delta". https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/atomic/AtomicInteger.html#addAndGet(int)

DomGarguilo: I agree, I think it makes more sense to either:
keep both methods and only allowing non-negative numbers for each, multiplying by -1 for decrement()
**or,**
just use one method that will add to `openFiles` regardless of whether the input is positive or negative. (preserving the functionality of `addAndGet()`)

dlmarion: See 4b0db48

milleruntime (approved):

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
```diff
@@ -44,8 +49,19 @@ public void addYield(long value) {
     yields.record(value);
   }
 
+  public void incrementOpenFiles(int delta) {
+    Preconditions.checkArgument(delta >= 0, "Parameter must be zero or positive");
```
EdColeman, line 53: Do we need to throw an unchecked exception here?  Would Math.max(0, val) work instead?

dlmarion: I like this better, changed in b1fb539

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
```diff
@@ -44,8 +47,19 @@ public void addYield(long value) {
     yields.record(value);
   }
 
+  public void incrementOpenFiles(int delta) {
+    Math.max(0, delta);
+    openFiles.addAndGet(delta);
```
EdColeman, line 52:
```suggestion
    openFiles.addAndGet(Math.max(0,delta));
```

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerScanMetrics.java
```diff
@@ -49,7 +49,7 @@ public void addYield(long value) {
 
   public void incrementOpenFiles(int delta) {
     Math.max(0, delta);
```
EdColeman, line 51: Without assignment, this statement is unnecessary.  Sorry - I didn't make it a multi-edit suggestion.

dlmarion: No, you're right, good catch. I've made like 4 mistakes on this simple change. I think I just need to start my vacation now.

EdColeman: Or - more coffee

PR #2400, opened 2021-12-27
Correct user-facing misspellings

foster33:
The following PR serves to correct misspellings throughout Accumulo that users may see.

Corrections were found in log messages, exceptions, print statements, etc.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/file/BloomFilterLayer.java
+++ b/core/src/main/java/org/apache/accumulo/core/file/BloomFilterLayer.java
@@ -544,7 +544,7 @@ public class BloomFilterLayer {
 
     t2 = System.currentTimeMillis();
 
-    out.printf("existant lookup rate %6.2f%n", 500 / ((t2 - t1) / 1000.0));
+    out.printf("existent lookup rate %6.2f%n", 500 / ((t2 - t1) / 1000.0));
     out.println("expected hits 500.  Receive hits: " + count);
     bmfr.close();
   }
--- a/core/src/main/java/org/apache/accumulo/core/file/BloomFilterLayer.java
+++ b/core/src/main/java/org/apache/accumulo/core/file/BloomFilterLayer.java
@@ -544,7 +544,7 @@ public class BloomFilterLayer {
 
     t2 = System.currentTimeMillis();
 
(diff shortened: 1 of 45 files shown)


core/src/main/java/org/apache/accumulo/core/file/BloomFilterLayer.java
```diff
@@ -544,7 +544,7 @@ public static void main(String[] args) throws IOException {
 
     t2 = System.currentTimeMillis();
 
-    out.printf("existant lookup rate %6.2f%n", 500 / ((t2 - t1) / 1000.0));
+    out.printf("existent lookup rate %6.2f%n", 500 / ((t2 - t1) / 1000.0));
```
EdColeman, line 547: Maybe this should be existing?
``` suggest
    out.printf("existing lookup rate %6.2f%n", 500 / ((t2 - t1) / 1000.0));
```

core/src/main/java/org/apache/accumulo/core/util/AddressUtil.java
```diff
@@ -61,7 +61,7 @@ public static int getAddressCacheNegativeTtl(UnknownHostException originalExcept
     try {
       negativeTtl = Integer.parseInt(Security.getProperty("networkaddress.cache.negative.ttl"));
     } catch (NumberFormatException exception) {
-      log.warn("Failed to get JVM negative DNS respones cache TTL due to format problem "
+      log.warn("Failed to get JVM negative DNS responses cache TTL due to format problem "
```
EdColeman, line 64: Maybe response?
``` suggest
      log.warn("Failed to get JVM negative DNS response cache TTL due to format problem "
```

DomGarguilo, line 64:
```suggestion
      log.warn("Failed to get JVM negative DNS response cache TTL due to format problem "
```

DomGarguilo (changes requested):

EdColeman (approved):
Mainly looked for changes that would lead to errors - I'm not the best speller so saying those changes look correct is not much of an endorsement, but other than a few questions on word choice - I do not see other issues.

DomGarguilo (approved):

EdColeman (approved):
LGTM - same caveats to my spelling apply though.

Issue #946, opened 2019-02-08
Consider adding critical thread metrics for monitoring


EdColeman:
Exposing metrics for critical process threads could improve monitoring and provide additional insight for performance trending.

For example, certain threads in master and in the tserver processes need to run periodically, if they do not, this is an indication that the process is likely unhealthy / having issues.  Exposing the fact that the threads are running, progressing or had a successful completion of a task would improve monitoring capabilities Additionally, if the "run-time" was provided, this could be used to gauge relative health by trending the performance over time / across upgrades,....

This issue is to capture possible candidate threads / processes that would be beneficial to incorporate into metrics reporting.

keith-turner:
The tablet server has a thread that continual checks if tablets need compaction or split.

https://github.com/apache/accumulo/blob/c7a54c80b937fd606ebdd6b3672f837f65b6258f/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServer.java#L2175

The master has three really important threads that assign tablets for the root, metadata, and user tablets.  If these threads are not running then tablets will not be assigned.  These threads are all run by TabletGroupWatcher

https://github.com/apache/accumulo/blob/050dec2003a786ea014c994a38e180a82b997c0d/server/master/src/main/java/org/apache/accumulo/master/TabletGroupWatcher.java#L136

There is currently code that watches compactions and logs a warning if it has not read or written any data in a certain amount of time.  It addition to logging a warning it might be nice to  increment a stuck compaction counter.  Could decrement when unstuck.

https://github.com/apache/accumulo/blob/2b9c9275ea5f992cfa2bd1a7e3f8994a41e69df3/server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactionWatcher.java#L31

There is also code that looks stuck tablet loads and logs a warning.   Would also be nice to have a counter for stuck loads.

https://github.com/apache/accumulo/blob/b915947c0c22d9db717067b601c62829205d1505/server/tserver/src/main/java/org/apache/accumulo/tserver/TabletServerResourceManager.java#L437

EdColeman:
PR: https://github.com/apache/accumulo/pull/1379 - added improved metadata consistency checking - this may be another candidate for improved reporting.  Initially considered adding it to the gc metrics improvements, but decided that a more comprehensive look and additional testing makes it better suited as a separate change.

dlmarion:
[...]

PR #2405, opened 2022-01-03
Created additional metrics for long assignments and stuck compactions

dlmarion:
Related to #946


server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -107,8 +108,58 @@
...
+    private long written;
+    private boolean stuck = false;
+
+    public CompactionProgress(AccumuloConfiguration config) {
+      warnTime = config.getTimeInMillis(Property.TSERV_COMPACTION_WARN_TIME);
+    }
+
+    @Override
+    public void registerMetrics(MeterRegistry registry) {
+      Gauge.builder(METRICS_TSERVER_MAJC_STUCK, this, CompactionProgress::getNumStuck)
```
EdColeman, line 129: With modification - this could show the "stuck" time rather than a boolean.  This would allow users to set external alarms based on time rather than on a boolean condition. It could also provide insight if things can somehow become "unstuck" without intervention, you could plot the stuck time over time to see how things are progressing.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -107,8 +108,58 @@
...
+          .description("Number of stuck major compactions").register(registry);
+    }
+
+    public long getNumStuck() {
+      return stuck ? 1 : 0;
+    }
```
EdColeman, line 135: Something like if not stuck return 0, else return System.currentTimeMillis() - time

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
dlmarion: So, each Compactor runs one majc at a time, but the TabletServer can run many of them at once. I was approaching this from the perspective that a user could just plot stuck majc's over time and be able to drill down to one of the processes. For your suggestion to work, I think I would need to emit multiple Timers , one per stuck extent?

server/tserver/src/main/java/org/apache/accumulo/tserver/metrics/TabletServerMetrics.java
```diff
@@ -46,6 +51,10 @@ public void registerMetrics(MeterRegistry registry) {
         .description("Number of queued major compactions").register(registry);
     Gauge.builder(METRICS_TSERVER_MINC_RUNNING, util, TabletServerMetricsUtil::getMinorCompactions)
         .description("Number of active minor compactions").register(registry);
+    Gauge
+        .builder(METRICS_TSERVER_MAJC_STUCK, util,
```
EdColeman, line 55: If this is separate from Compactor METRICS_TSERVER_MAJC_STUCK and is the sum of stuck counts, then maybe named METRICS_TSERVER_MAJC_STUCK_COUNT or METRICS_TSERVER_MAJC_STUCK_TOTAL?  (I don't recall suggested naming conventions)

dlmarion: They are the same metric, just being reported from different processes (the process name is in a tag). `METRICS_TSERVER_MAJC_STUCK` is just the variable name, the metric being reported is `metrics.tserver.majc.stuck`.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
EdColeman: From looking at line 134 - it seemed that the value would only be 0 or 1 - and that there is a summation occurring later?

It seemed that more info could be conveyed other than a boolean. If it is being summed elsewhere, does it need to be a metric here?

I could just be reading this wrong...

dlmarion: Right, the Compactor only runs one majc at a time. So, the current one is either stuck, or not stuck. The summation across all processes (or by some other aggregation) would be done in a downstream time series database.

EdColeman: +1 to the summation across all.

I think I am just triggered by the compactor count of 0 or 1 being a metric.  I guess I'm asking: Can it be kept as a count - but not as a metric? Or, if kept as a metric provide additional information?

dlmarion: I'm confused by your use of `metric` in the latest comment. I have it set as a `gauge` because the number of stuck compactions is going to go up and down, in both the TabletServer and Compactor, over time.

EdColeman: I'm using metric as - a value that is reported by the metrics system

dlmarion: > Can it be kept as a count - but not as a metric? Or, if kept as a metric provide additional information?

So, if it's not a value reported by the metrics system, then what do you mean by keeping it as a `count` ? What additional information do you think need needs to be conveyed?

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactionWatcher.java
```diff
@@ -75,35 +80,42 @@ public void run() {
...
+            e.setStackTrace(trace);
+            LoggerFactory.getLogger(CompactionWatcher.class).warn("Compaction of "
+                + oci.compactionInfo.getExtent() + " to " + oci.compactionInfo.getOutputFile()
+                + " has not made progress for at least " + (time - oci.firstSeen) + "ms", e);
+            oci.loggedWarning = true;
+          }
         }
       }
     }
+    totalStuck += stuck;
```
keith-turner, line 118: Could the following be done?  If so could remove the unstuck calculation and deduction earlier in the code.

```suggestion
    totalStuck = stuck;
```

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -107,8 +108,58 @@
 import com.beust.jcommander.Parameter;
 import com.google.common.base.Preconditions;
 
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+
 public class Compactor extends AbstractServer implements CompactorService.Iface {
 
+  private static class CompactionProgress implements MetricsProducer {
```
keith-turner, line 116: Would it be possible to move CompactionWatcher from the tserver to server-base and reuse it in the compactor instead of introducing similar code in the compactor?

server/tserver/src/main/java/org/apache/accumulo/tserver/tablet/CompactionWatcher.java
dlmarion: addressed in f5b3223f4cfb0d7ac2d99664f81898962a54c5c9

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
dlmarion: addressed in f5b3223f4cfb0d7ac2d99664f81898962a54c5c9

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -693,6 +712,7 @@ public void run() {
                     "Compaction in progress, read %d of %d input entries ( %s %s ), written %d entries",
                     info.getEntriesRead(), inputEntries, percentComplete, "%",
                     info.getEntriesWritten());
+                watcher.run();
```
keith-turner, line 715: Not sure if this can happen, but I think there may be a case where stuck compaction stats are not cleaned up because of how this is called.  It seems like its only called periodocally when a compaction is running.  If a compaction that was stuck starts again, but watcher.run() is not called because of timing issues (like the previously stuck compaction quickly completes) then it would continue to look stuck.

dlmarion: In bec830ff3d0dc606864e5172fc463f80224f5ca7 I call `watcher.run()` after the compaction has finished to clear out the state.

server/base/src/main/java/org/apache/accumulo/server/compaction/CompactionWatcher.java
```diff
@@ -29,14 +29,13 @@
...
 import org.apache.accumulo.server.ServerContext;
-import org.apache.accumulo.server.compaction.CompactionInfo;
-import org.apache.accumulo.server.compaction.FileCompactor;
 import org.slf4j.LoggerFactory;
 
 public class CompactionWatcher implements Runnable {
   private final Map<List<Long>,ObservedCompactionInfo> observedCompactions = new HashMap<>();
   private final AccumuloConfiguration config;
   private static boolean watching = false;
+  private static long totalStuck = 0;
```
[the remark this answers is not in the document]
dlmarion: Changed in bec830ff3d0dc606864e5172fc463f80224f5ca7

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
dlmarion: Check out the changes in bec830ff3d0dc606864e5172fc463f80224f5ca7, I modified the code to use a LongTaskTimer.

server/base/src/main/java/org/apache/accumulo/server/compaction/CompactionWatcher.java
```diff
@@ -79,6 +84,10 @@ public void run() {
     copy.keySet().removeAll(newKeys);
 
     for (ObservedCompactionInfo oci : copy.values()) {
+      if (oci.stuckSample != null) {
```
EdColeman, line 87: The null check / set sample to null may be unnecessary - from the micrometer docs, LongTaskTimers do not report on completed tasks - if may be enough to record the start / stop which would provide the active task count, and the active count duration info.

dlmarion: A single compaction could get stuck several times over the life of the compaction. I'm trying to account for that.

EdColeman (approved):

EdColeman (approved):

PR #2415, opened 2022-01-12
Create new external compactions admin utility

### LINKED ISSUE #2386: Cancel specific external compaction
A user can call "compact --cancel" on a table which will cancel all compactions for that table. It would be nice to be able to cancel a specific compaction, which I think can now be done easily using the ECID. I think this will only work with external compactions though as we don't have a way to specify an internal compaction.

milleruntime:
* Create a few options for external compactions
* Create new RPC calls for cancelling an external compaction
* Closes #2386

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionId.java
+++ b/core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionId.java
@@ -56,4 +56,14 @@ public class ExternalCompactionId extends AbstractId<ExternalCompactionId> {
     return new ExternalCompactionId(id);
   }
 
+  /**
+   * Sanitize user input for the ECID string with proper "ECID:" prefix.
+   */
+  public static ExternalCompactionId from(String ecid) {
+    ecid = ecid.replace("ecid:", "ECID:");
+    if (!ecid.startsWith("ECID:"))
+      ecid = "ECID:" + ecid;
+    return of(ecid);
+  }
+
 }
--- a/core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionId.java
+++ b/core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionId.java
@@ -60,9 +60,9 @@ public class ExternalCompactionId extends AbstractId<ExternalCompactionId> {
    * Sanitize user input for the ECID string with proper "ECID:" prefix.
    */
   public static ExternalCompactionId from(String ecid) {
-    ecid = ecid.replace("ecid:", "ECID:");
-    if (!ecid.startsWith("ECID:"))
-      ecid = "ECID:" + ecid;
+    ecid = ecid.replace(PREFIX.toLowerCase(), PREFIX);
+    if (!ecid.startsWith(PREFIX))
+      ecid = PREFIX + ecid;
     return of(ecid);
   }
 
--- a/server/base/src/main/java/org/apache/accumulo/server/util/ECAdmin.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/util/ECAdmin.java
@@ -130,6 +130,7 @@ public class ECAdmin implements KeywordExecutable {
 
   private void cancelCompaction(ServerContext context, String ecid) {
     CompactionCoordinatorService.Client coordinatorClient = null;
+    ecid = ExternalCompactionUtil.prefixECID(ecid);
     try {
       coordinatorClient = getCoordinatorClient(context);
       coordinatorClient.cancel(TraceUtil.traceInfo(), context.rpcCreds(), ecid);
--- a/server/base/src/main/java/org/apache/accumulo/server/util/ECAdmin.java
+++ b/server/base/src/main/java/org/apache/accumulo/server/util/ECAdmin.java
@@ -1,0 +1,210 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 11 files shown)


DomGarguilo:
> I was thinking of adding a thrift method to the `CompactionCoordinator` called `cancel()` that will call cancel on the appropriate compactor.

Would this address #2386?

milleruntime:
> > I was thinking of adding a thrift method to the `CompactionCoordinator` called `cancel()` that will call cancel on the appropriate compactor.
> 
> Would this address #2386?

Yup, that is the plan. I couldn't find a good way for the user to cancel the compaction so I made this external compactions admin utility.

milleruntime:
I think this PR is ready for review. This creates the admin tool for external compactions but only with useful commands that currently work. I added RPCs for cancelling but have some ideas for a follow on task to provide more command and control.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
```diff
@@ -360,6 +360,12 @@ private void cancel(String externalCompactionId) throws TException {
     }
   }
 
+  @Override
+  public void cancel(TInfo tinfo, TCredentials credentials, String externalCompactionId)
+      throws TException {
```
dlmarion, line 365: The other thrift api methods perform a security check first, do we want to do that here?

dlmarion:
This looks correct. Just one comment about API security check.

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
milleruntime: I didn't think it needed SYSTEM level permission (root) like the other method required. I thought it should require at least some level of USER permission but wasn't sure since the permissions are so confusing. Any ideas?

dlmarion: oh ok. I'm not up to speed on all of the permissions, I just know I used it on all of the other thrift api methods. If it's not appropriate, then that is good with me.

dlmarion (approved):

server/compactor/src/main/java/org/apache/accumulo/compactor/Compactor.java
milleruntime: I saw that the FATE Table cancel compact operation has the same permissions as a user needs to compact. So we could try that:
From FateServiceHandler:
`          canCancelCompact = manager.security.canCompact(c, tableId, namespaceId);
`

milleruntime:
FYI @dlmarion I made some improvements in 4c78355 based on your comments. I just wanted to give you a heads up since you already approved this PR.

server/base/src/main/java/org/apache/accumulo/server/util/ECAdmin.java
```diff
@@ -130,6 +130,7 @@ public void execute(final String[] args) {
 
   private void cancelCompaction(ServerContext context, String ecid) {
     CompactionCoordinatorService.Client coordinatorClient = null;
+    ecid = ExternalCompactionUtil.prefixECID(ecid);
```
dlmarion, line 133: I assume you are trying to clean up bad user input here? Can we just use `ExternalCompactionId.of(String)` here and let the user pass the right thing? Maybe we can clean up the error messages in that method to be more specific about the issues.

milleruntime: > I assume you are trying to clean up bad user input here?

Yes. 

> Can we just use ExternalCompactionId.of(String) here and let the user pass the right thing? Maybe we can clean up the error messages in that method to be more specific about the issues.

I wasn't sure. I did this in the monitor just because it wasn't clear to the user that the server code required the prefix and I didn't like relying on trial and error to figure that out. Even with a nice error message, I don't like users having to guess if you need the prefix or not.

I pinged @keith-turner on chat asking if he had any thoughts behind the origin of the prefix but haven't heard back yet.

dlmarion: `ExternalCompactionId.of` also validates that the id is a UUID. If we are going to have another method to take a String and massage it into an ExternalCompactionId, then maybe we can move your method body into `ExternalCompactionId.from(String)` or something like that. Then at least we can re-use PREFIX.

milleruntime: I like the idea of consolidating the logic into `ExternalCompactionId`. The only issue is that both places I need a String and not the object. What about making the prefix method static member of `ExternalCompactionId`? Then we could reuse the validation in the `of()` method.

dlmarion: Right, but on the server side, it parses the incoming ecid string into an ExternalCompactionId. It would be good to know that on the client side it's a valid ExternalCompactionId before sending it. You can always get the String value from the ExternalCompactionId, right? To be clear, I'm suggesting something like `ExternalCompactionId.from(String).toString()`. Whether you call the method `from`,`parse`, etc. doesn't matter to me. You called the method `prefix`, but it looks like it's just trying to create a valid ECID from the user input.

milleruntime: Created `from` method in b233afe

core/src/main/java/org/apache/accumulo/core/metadata/schema/ExternalCompactionId.java
```diff
@@ -56,4 +56,14 @@ public static ExternalCompactionId of(String id) {
     return new ExternalCompactionId(id);
   }
 
+  /**
+   * Sanitize user input for the ECID string with proper "ECID:" prefix.
+   */
+  public static ExternalCompactionId from(String ecid) {
+    ecid = ecid.replace("ecid:", "ECID:");
```
dlmarion, line 63: You can use PREFIX now

milleruntime: Whoops, good catch.

dlmarion:
Changes LGTM.

PR #2432, opened 2022-01-26
Modified ThreadPools to conditionally register metrics

### LINKED ISSUE #2423: Seeing out of memory error in manager
**Describe the bug**

When running Accumulo 2.1.0-SNAPSHOT for extended periods (like 24+hours) I would see the manager process die with an out of memory error. 

 - Affected version(s) of this project: 2.1.0-SNAPSHOT
 
**To Reproduce**

Run the Accumulo 2.1.0 Manager for an extended period of time with 512M of RAM.  512M is the default config set by Muchos for the manager.  It has been this size for a very long time and I don't recall seeing this error before.

**Expected behavior**

No OOME

**Screenshots**

After the first time this happened, I modified the manager process to heap dump on OOME.  Below is a screenshot of analyzing the heap dump.  I saw a lot of micrometer objects in the heap dump.  Randomly poking around, a lot of these seemed to have `exectuor` in the name.  Based on this I looked in the manager code and found it was [periodically creating thread pools to get tserver info](https://github.com/apache/accumulo/blob/1f80b7445d1320efaa8db1fb58c7b67d73c9fdbc/server/manager/src/main/java/org/apache/accumulo/manager/Manager.java#L920) and this [may register the thread pool statically for monitoring](https://github.com/apache/accumulo/blob/1f80b7445d1320efaa8db1fb58c7b67d73c9fdbc/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java#L287).  I have yet to connect the objects in the heap dump to the linked Manager code.  The objects seem to be statically referenced so I am guessing the refs from Accumulo code are gone and only the static metric registry refs are left. I am not sure yet if the Manager code I linked is the culprit though.

image

**Additional context**

dlmarion:
When ThreadPools calls ExecutorServiceMetrics, it creates Meters
that are registered with the MeterRegistry. However, there is no
easy way to remove the Meters from the MeterRegistry when the
ThreadPool is shut down. Modified the ThreadPools methods to accept
a boolean value for registering metrics for the newly created
ThreadPool. Modified classes that use ThreadPools class to only
pass true for this boolean value when the ThreadPool is long lived.

Closes #2423

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
+++ b/core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
@@ -104,81 +104,86 @@ public class ThreadPools {
    */
   @SuppressWarnings("deprecation")
   public static ExecutorService createExecutorService(final AccumuloConfiguration conf,
-      final Property p) {
+      final Property p, boolean emitThreadPoolMetrics) {
 
     switch (p) {
       case GENERAL_SIMPLETIMER_THREADPOOL_SIZE:
-        return createScheduledExecutorService(conf.getCount(p), "SimpleTimer");
+        return createScheduledExecutorService(conf.getCount(p), "SimpleTimer",
+            emitThreadPoolMetrics);
       case MANAGER_BULK_THREADPOOL_SIZE:
         return createFixedThreadPool(conf.getCount(p),
             conf.getTimeInMillis(Property.MANAGER_BULK_THREADPOOL_TIMEOUT), TimeUnit.MILLISECONDS,
-            "bulk import");
+            "bulk import", emitThreadPoolMetrics);
       case MANAGER_RENAME_THREADS:
-        return createFixedThreadPool(conf.getCount(p), "bulk move");
+        return createFixedThreadPool(conf.getCount(p), "bulk move", emitThreadPoolMetrics);
       case MANAGER_FATE_THREADPOOL_SIZE:
-        return createFixedThreadPool(conf.getCount(p), "Repo Runner");
+        return createFixedThreadPool(conf.getCount(p), "Repo Runner", emitThreadPoolMetrics);
       case MANAGER_STATUS_THREAD_POOL_SIZE:
         int threads = conf.getCount(p);
         if (threads == 0) {
           return createThreadPool(0, Integer.MAX_VALUE, 60L, TimeUnit.SECONDS,
-              "GatherTableInformation", new SynchronousQueue<Runnable>(), OptionalInt.empty());
+              "GatherTableInformation", new SynchronousQueue<Runnable>(), OptionalInt.empty(),
+              emitThreadPoolMetrics);
         } else {
-          return createFixedThreadPool(threads, "GatherTableInformation");
+          return createFixedThreadPool(threads, "GatherTableInformation", emitThreadPoolMetrics);
         }
       case TSERV_WORKQ_THREADS:
-        return createFixedThreadPool(conf.getCount(p), "distributed work queue");
+        return createFixedThreadPool(conf.getCount(p), "distributed work queue",
+            emitThreadPoolMetrics);
       case TSERV_MINC_MAXCONCURRENT:
-        return createFixedThreadPool(conf.getCount(p), 0L, TimeUnit.MILLISECONDS,
-            "minor compactor");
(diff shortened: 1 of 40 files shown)


core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
```diff
@@ -104,81 +104,86 @@ public static void resizePool(final ThreadPoolExecutor pool, final AccumuloConfi
    */
   @SuppressWarnings("deprecation")
   public static ExecutorService createExecutorService(final AccumuloConfiguration conf,
-      final Property p) {
+      final Property p, boolean emitThreadPoolMetrics) {
```
keith-turner, line 107: Could add some javadoc for this param like the following

```java
@param emitThreadPoolMetrics When set to true will emit metrics and register the metrics in a static registry.  After the thread pool is deleted, there will still be metrics objects related to it in the static registry.  There is no way to clean these left over objects up therefore its recommened that this option only be set true for long lived thread pools.  Creating lots of short lived thread pools and registering them can lead to out of memory errors over long time periods.  
```

keith-turner (approved):

core/src/main/java/org/apache/accumulo/fate/Fate.java
```diff
@@ -230,7 +230,7 @@ public Fate(T environment, TStore<T> store, Function<Repo<T>,String> toLogStrFun
    */
   public void startTransactionRunners(AccumuloConfiguration conf) {
     final ThreadPoolExecutor pool = (ThreadPoolExecutor) ThreadPools.createExecutorService(conf,
-        Property.MANAGER_FATE_THREADPOOL_SIZE);
+        Property.MANAGER_FATE_THREADPOOL_SIZE, false);
```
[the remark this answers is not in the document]
dlmarion: Addressed in 1a7f098

core/src/main/java/org/apache/accumulo/core/util/threads/ThreadPools.java
dlmarion: Addressed in 1a7f098

ctubbsii (commented):
Seems fine, but one suggestion on API design:

Boolean parameters can make code harder to read. Instead of boolean parameters, you could have a separate method name, like `.action()` and `.actionWithMetrics()`, or you could use a builder, like `builder.withMetrics().build()`, or you could have enums for flags, like `.action(EnumSet.of(WITH_METRICS, OTHER_FLAG))`

keith-turner:
> Boolean parameters can make code harder to read. Instead of boolean parameters, you could have a separate method name, like .action() and .actionWithMetrics(), or you could use a builder, like builder.withMetrics().build(), or you could have enums for flags, like .action(EnumSet.of(WITH_METRICS, OTHER_FLAG))

Another option would be to replace the last boolean argument in the method with a new enum type.  The single boolean here is not too bad, an enum would be more readable.  We have had methods in the past w/ multiple boolean arguments and that was awful to read.

dlmarion:
> Boolean parameters can make code harder to read. Instead of boolean parameters, you could have a separate method name, like .action() and .actionWithMetrics(), or you could use a builder, like builder.withMetrics().build(), or you could have enums for flags, like .action(EnumSet.of(WITH_METRICS, OTHER_FLAG))

I have no issue with modifying ThreadPools to use a builder pattern, but I think that should be done in a different issue. IDEs make it pretty easy to see the formal parameter name for the method and the formal parameter name is descriptive in this case.

> Another option would be to replace the last boolean argument in the method with a new enum type. The single boolean here is not too bad, an enum would be more readable. We have had methods in the past w/ multiple boolean arguments and that was awful to read.

I have no issue with modifying this PR to use an enum for the boolean parameter, but I'm not sure that it's necessary if there is going to be a follow-on issue created to modify the class to use a builder pattern.

keith-turner:
> IDEs make it pretty easy to see the formal parameter name for the method and the formal parameter name is descriptive in this case.

That is a good point.  That does dramatically increase readability. However when reviewing a PR on GH, you dont have that nice feature.  But thinking back to my review that was not a problem at all. The review was very easy for me to review because I knew what the single boolean did.  The more difficult part of the review was that I had to look at code outside the PR sometimes to see how code was called externally, which had nothing to do with the boolean.  More than one boolean would certainly have given me a headache during review though.  

> I'm not sure that it's necessary if there is going to be a follow-on issue created to modify the class to use a builder pattern.

That makes sense to me.

PR #2433, opened 2022-01-26
Make replication table creation happen only if needed

milleruntime:


milleruntime:
It looks like this breaks the GC so some additional changes would have to be made for this to work.

milleruntime:
This change is too risky with very little benefit.

PR #2445, opened 2022-02-01
Remove self-assignment statement.

jmark99:
Removed a self-assignment of the hostPortString variable to itself.

If the hostPortString is non-null, nothing happens and the variable is unchanged. If the hostPortString is null, a ullPointerException is thrown. There is no need to assign it to itself.

Change suggested by errorprone analysis.

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
+++ b/core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
@@ -19,6 +19,7 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 
 import java.io.Serializable;
+import java.util.Objects;
 
 /**
  * This class was copied from Guava release 23.0 to replace the older Guava 14 version that had been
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/accumulo/core/util/HostAndPort.java
```diff
@@ -143,7 +143,7 @@ public static HostAndPort fromParts(String host, int port) {
    *           if nothing meaningful could be parsed.
    */
   public static HostAndPort fromString(String hostPortString) {
-    hostPortString = java.util.Objects.requireNonNull(hostPortString);
+    java.util.Objects.requireNonNull(hostPortString);
```
EdColeman, line 146: Maybe add the optional message stating that hostPortString was null?  The way that the NPE shows up in the logs, adding the message seems to help figuring things out.

EdColeman (approved):

ctubbsii (changes requested):

DomGarguilo (approved):

EdColeman (approved):

PR #2446, opened 2022-02-01
Make ClientContext only verify instanceId once

milleruntime:


milleruntime:
I am working on similar changes but wanted to get this one in separately just to get eyes on it.

PR #2448, opened 2022-02-01
Remove seemingly useless method in AgeOffFilterTest

DomGarguilo:
From what I can tell, `createOutputData()` essentially creates a new (empty) Map, filters it (which would not change an empty map) and returns it. I'm not sure what the intended functionality of this method is, but to me it seems like just creating a new TreeMap in its place is equivalent. Hopefully I'm not overlooking something.

### DIFF
--- a/test/src/test/java/org/apache/accumulo/test/iterator/AgeOffFilterTest.java
+++ b/test/src/test/java/org/apache/accumulo/test/iterator/AgeOffFilterTest.java
@@ -19,7 +19,6 @@
 package org.apache.accumulo.test.iterator;
 
 import java.util.List;
-import java.util.Map.Entry;
 import java.util.TreeMap;
 
 import org.apache.accumulo.core.client.IteratorSetting;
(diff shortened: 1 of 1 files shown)


EdColeman:
It looks like the test should be creating data that has a TTL such that some data would not be subject to age off. When the age off iterator is run, some data should remain. If this functions as expected, the output would not be an empty map, but the surviving entries.  If the test is passing with an empty map, that would suggest that the test is not testing what was intended.

DomGarguilo:
> It looks like the test should be creating data that has a TTL such that some data would not be subject to age off. When the age off iterator is run, some data should remain. If this functions as expected, the output would not be an empty map, but the surviving entries. If the test is passing with an empty map, that would suggest that the test is not testing what was intended.

As far as I can tell, `createOutputData()` returns an empty TreeMap. The test does pass with my changes i.e., with an empty map. Maybe this PR should be closed in place of an issue if that is the case.

DomGarguilo:
> If the test is passing with an empty map, that would suggest that the test is not testing what was intended.

I created an issue to track this bug: #2449

PR #2451, opened 2022-02-01
Create InstanceId to replace use of String

milleruntime:
* Create InstanceId object to replace use of String across the code
* Create new API method on InstanceOperations called getInstanceId()
* Deprecate API method InstanceOperations getInstanceID()

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/data/InstanceId.java
+++ b/core/src/main/java/org/apache/accumulo/core/data/InstanceId.java
@@ -18,42 +18,54 @@
  */
 package org.apache.accumulo.core.data;
 
+import java.util.Objects;
+import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 
 /**
- * A strongly typed representation of a table ID. This class cannot be used to get a table ID from a
- * table name, but does provide the table ID string wrapped with a stronger type. The constructor
- * for this class will throw an error if the canonical parameter is null.
+ * A strongly typed representation of an Accumulo instance ID. The constructor for this class will
+ * throw an error if the canonical parameter is null.
  *
- * @since 2.0.0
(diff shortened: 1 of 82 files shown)


core/src/main/java/org/apache/accumulo/core/data/InstanceId.java
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Get a InstanceId object for the provided canonical string. This is guaranteed to be non-null
+   *
+   * @param canonical
+   *          Instance ID string
+   * @return InstanceId object
+   */
+  public static InstanceId of(final String canonical) {
+    try {
```
DomGarguilo, line 53: Should we check that `canonical` is non null here? Similar to how `uuid` is checked in the method below?

milleruntime: This is done in the constructor of the parent class `AbstractId`

milleruntime:
I actually think this may have already found a bug in `AuthenticationTokenIdentifier` that is being exposed by the now failing test `SaslDigestCallbackHandlerTest`.

ctubbsii (commented):
I haven't done a thorough review, but I skimmed it. I like the idea.

I was thinking that we might want to make `AbstractId.toString` final, so it always returns `.canonical()`, so that it's contractually consistent, just because there's lots of places where it's logged/concatenated/etc. with other strings, and it's assuming that it's the same as `.canonical()`, but that's not necessarily guaranteed.

milleruntime:
> I was thinking that we might want to make `AbstractId.toString` final, so it always returns `.canonical()`, so that it's contractually consistent, just because there's lots of places where it's logged/concatenated/etc. with other strings, and it's assuming that it's the same as `.canonical()`, but that's not necessarily guaranteed.

That is a good idea. It will help with the other *Id types as well.

ctubbsii (approved):

Issue #1249, opened 2019-07-05
Allow FATE operations to report goal and status information.


keith-turner:
It would be very useful if FATE operations could report goal and status information.  This information could be display via command line utilities and on the monitor page.  Alternatively the information could just be logged in a very consistent way.  Goal information would be what the FATE op is trying to do and should not change for lifetime of the FATE operation.  Status information changes as the FATE operations proceeds.  For example for bulk import, Goal and status information could look like the following.

 * Fate Transaction : 1f2e3d4c
 * Goal: Import files from `hdfs://nn1/userdir5` to table `mediumTable`
 * Status :
   * 07/04/19 20:33 Writing blip marker to metadata table
   * 07/04/19 20:34 Moving files from `hdfs://nn1/userdir5` to `hdfs://nn1/accumulo/table/5/b-678`
   * 07/04/19 20:37 Requesting tablet servers to import files

ctubbsii:
That'd be very neat.

keith-turner:
#1316 has overlapping goals.

jmark99:
@EdColeman are you currently working on this task? If not, do you mind if I give it a go?

EdColeman:
I'm working fate metrics, but have not started this.

milleruntime:
@jmark99 are you still looking at this?

jmark99:
@millerruntime yes, I’ve been away for two weeks on vacation. Just getting back today.

milleruntime:
Here are some ideas for the displaying FATE status in the monitor (and an issue specific to the monitor, that can be done separately): https://issues.apache.org/jira/browse/ACCUMULO-809

jmark99:
Removed myself from this ticket since I haven't had a chance to do much work on it lately and Keith may be addressing it with #1624.

PR #2475, opened 2022-02-09
Refactor fate

### LINKED ISSUE #2473: Refactor FATE
I was looking at the warnings we suppress in `ZooStore` and they all seem to stem from the serialization of the generics. I was wondering if it would be better to refactor out the generic types so we don't have to arbitrarily serialize objects and blindly cast them. These would be a bit of work but it might not be too bad. I think if we create an interface like `FateEnvironment` to pass to the operations we wouldn't have to pass the Manager around.

milleruntime:
* Refactor Fate to allow removing of generics and be able to improve serialization
* Move Fate classes from core to Manager
* Create new FateZooStore in core and make ZooStore extend it, consolidating
methods between the 2 classes. This allows sharing between Manager and Shell
* Extract inner classes to allow reuse and consolidation across packages. Pulled FateStatus out of
AdminUtil into its own class in core
* Drop Generics from many different Fate interfaces and classes
* Combine Fate admin utilities into one FateAdmin
* Drop master FateAdmin
* Drop AdminUtil in favor of FateCommandHelper and use in FateCommand
* Rename method from checkGlobalLock() to isManagerLockValid()
* Move TStatus to core and rename to FateTransactionStatus
* Closes #2473


server/manager/src/main/java/org/apache/accumulo/manager/tableOps/compact/CompactionDriver.java
```diff
@@ -120,28 +118,26 @@ public long isReady(long tid, Manager manager) throws Exception {
...
-    for (TServerInstance tsi : serversToFlush.keySet()) {
-      try {
-        final TServerConnection server = manager.getConnection(tsi);
-        if (server != null)
-          server.compact(manager.getManagerLock(), tableId.canonical(), startRow, endRow);
-      } catch (TException ex) {
-        LoggerFactory.getLogger(CompactionDriver.class).error(ex.toString());
-      }
-    }
+    // TODO figure this out
```
milleruntime, line 134: This is a problem. The `LiveTServerSet` is maintained in the Manager. I might be able to move the class to core.

milleruntime: @ctubbsii here is another place where it could be nice to have `context.getZooReaderWriter()` in ClientContext. If I move `LiveTServerSet` to core, it uses `ServerContext` and writes to ZK.

core/src/main/java/org/apache/accumulo/fate/Repo.java
```diff
@@ -34,16 +34,11 @@ default String getReturn() {
...
 
-  // return a string of the current status
-  default String getStatus() {
-    return "WARNING: Status has not been implemented yet.";
-  }
-
   // return the Goal of the current operation
   default Goal getGoal() {
-    return new Goal("WARNING: Status has not been implemented yet.");
+    return new Goal(0, 0, "WARNING: Operation has not implemented a goal.");
```
keith-turner, line 39: In #1249 I was thinking the goal information could be provided when a fate transaction is seeded and stored in ZK.  It would not be obtained from each repo, instead status information could be related with each repo in the stack in ZK.  So to get the information for a fate transaction, get the goal for the fate transaction and then travese the stack getting the status information for each repo.

keith-turner: For example maybe [this method](https://github.com/apache/accumulo/blob/7088cdec4fab5bb590645ab6dc10fb7b2021f50e/core/src/main/java/org/apache/accumulo/fate/Fate.java#L271) could be changed to :

```java
public void seedTransaction(long tid, Repo<T> repo, boolean autoCleanUp, String goalMessage)
```

milleruntime: Thanks for the ideas. I hadn't thought of that.

milleruntime: Also what do you think of trying to get the status reporting into 2.1 in a separate change? Seems like adding information to the serialized nodes shouldn't break whats already there.

keith-turner: I think it would be a useful feature to have in 2.1 if we can get it done.  Thinking about it a bit more I am puzzling over how the status info would stored in ZK.  Was thinking it would go w/ the stack of repos, however in the case of undoing repos the status info would be lost.   Under normal circumstances repos are pushed onto a per fate tx stack in ZK as they are executed, so this stack could also have status info.  However when there is an error fate will start popping up repos off the stack and executing undo ops on them.  Because of this thinking it may make more sense to just have a per fate transaction list of status strings/objects in ZK.

milleruntime:
@keith-turner I ran into a dependency snag. The FateCommand uses a bunch of Fate types and is in the Shell package. Moving the Fate stuff to manager, breaks this. I don't know of a good solution other than rewrite `FateCommand` and make new types that use API. Any ideas?

keith-turner:
> @keith-turner I ran into a dependency snag. The FateCommand uses a bunch of Fate types and is in the Shell package. Moving the Fate stuff to manager, breaks this. I don't know of a good solution other than rewrite FateCommand and make new types that use API. Any ideas?

@milleruntime I missed this comment.  I have not looked at this in a while, I can take a fresh look at it.  Is this PR in a state where you want a review or are you still changing some things?

milleruntime:
> > @keith-turner I ran into a dependency snag. The FateCommand uses a bunch of Fate types and is in the Shell package. Moving the Fate stuff to manager, breaks this. I don't know of a good solution other than rewrite FateCommand and make new types that use API. Any ideas?
> 
> @milleruntime I missed this comment. I have not looked at this in a while, I can take a fresh look at it. Is this PR in a state where you want a review or are you still changing some things?

Yeah it is in a good state, if you want to take a look at it. This PR isn't a priority though as it will probably have to go into 3.0. I pulled some of these changes out into https://github.com/apache/accumulo/pull/2541. Also, @Manno15 is working on changes here: https://github.com/apache/accumulo/pull/2215, which should be able to get into 2.1 but have some new API additions you might want to look at as well.

milleruntime:
> > > @keith-turner I ran into a dependency snag. The FateCommand uses a bunch of Fate types and is in the Shell package. Moving the Fate stuff to manager, breaks this. I don't know of a good solution other than rewrite FateCommand and make new types that use API. Any ideas?
> > 
> > 
> > @milleruntime I missed this comment. I have not looked at this in a while, I can take a fresh look at it. Is this PR in a state where you want a review or are you still changing some things?
> 
> Yeah it is in a good state, if you want to take a look at it. This PR isn't a priority though as it will probably have to go into 3.0. I pulled some of these changes out into #2541. Also, @Manno15 is working on changes here: #2215, which should be able to get into 2.1 but have some new API additions you might want to look at as well.

If you want to wait for me to merge the latest changes in main, I am working on that now.

keith-turner:
I plan to circle back to this next week and review it again.

milleruntime:
I am going to rework these changes for 3.0, incorporating the deprecating of FateCommand, but I may have to just create a new branch.

milleruntime:
Closing this PR. It is too broad and contains too many different changes.

PR #2477, opened 2022-02-09
Expresion of type int may overflow before assignment to long.

jmark99:
From errorpone bug pattern description:

Performing an arithmetic expression on arguments of type int and then
assigning the result to a long is error-prone. The result is widened to
a long as the final step, and the intermediate results may overflow.


ctubbsii (approved):
These numerical values are so small, they wouldn't hit the int overflow anyway, but if it gets rid of some warnings in errorprone, I'm fine with it.

PR #2486, opened 2022-02-10
Move zooReaderWriter into ClientContext

milleruntime:
* Move the zooReaderWriter object into ClientContext to help in places where we need to
do things to ZK but don't have ServerContext available

### DIFF
--- a/core/src/main/java/org/apache/accumulo/core/clientImpl/ClientContext.java
+++ b/core/src/main/java/org/apache/accumulo/core/clientImpl/ClientContext.java
@@ -77,6 +77,7 @@ import org.apache.accumulo.core.util.OpTimer;
 import org.apache.accumulo.fate.zookeeper.ServiceLock;
 import org.apache.accumulo.fate.zookeeper.ZooCache;
 import org.apache.accumulo.fate.zookeeper.ZooCacheFactory;
+import org.apache.accumulo.fate.zookeeper.ZooReaderWriter;
 import org.apache.accumulo.fate.zookeeper.ZooUtil;
 import org.apache.hadoop.conf.Configuration;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


ctubbsii (changes requested):
I don't think this is the right solution. The difference between ZooReader and ZooReaderWriter is that one is available on the client, and the other is only available on the server-side, because it has the instance secret used to grant access to the restricted nodes in ZK.

As such, it does not make sense to put the ZooReaderWriter in the client context, any client context that is not a server should not have access to the credentials necessary to do any of the write operations.

I am not sure what is motivating this change, so it is difficult to provide a suggested alternative.

milleruntime:
> I am not sure what is motivating this change, so it is difficult to provide a suggested alternative.

This would help a lot with my changes in #2475

milleruntime:
> > > I am not sure what is motivating this change, so it is difficult to provide a suggested alternative.
> > 
> > 
> > This would help a lot with my changes in #2475
> 
> Why would clients need write access to ZK, though, in your changes?

FATE does. And FATE is in core. ServerContext is not.

milleruntime:
I made the `RepoEnv` interface return `ClientContext` and then cast it to `ServerContext` like this:
<pre>
context = (ServerContext) environment.getContext();
</pre>
Which i guess is fine. There are just lots of places where we write to ZK.

milleruntime:
> It seems some of FATE probably shouldn't be in core, but should probably be in the manager jar instead.

I agree. I will try moving it to the Manager jar.

> Perhaps one alternative is to combine ZooReader and ZooReaderWriter into one utility, with an `isWritable()` flag that the methods that write could use to guard against. That would help with your current situation, but could make things worse elsewhere, because we wouldn't know until runtime when we have passed the wrong instance. But, if everything uses the context to get the instance, it may be okay, because it should always have the correct one.

That isn't a bad idea but I be worried that it would be too easy for developers to just start using the Writer everywhere or making it a runtime error would be harder to guard against.

